PR #49, opened 2022-01-20
chore(args): update arg parsing to clap v3

tranzystorekk:
## Description

Updates arg parsing to `clap` 3 with the `derive` feature (the new `structopt`).

## Types of changes

- [ ] Other 

## Checklist:

### DIFF
--- a/git-cliff/src/lib.rs
+++ b/git-cliff/src/lib.rs
@@ -107,7 +107,8 @@ pub fn run(mut args: Opt) -> Result<()> {
 	}
 	if args.sort == Sort::Oldest {
 		if let Some(ref sort_commits) = config.git.sort_commits {
-			args.sort = Sort::from_str(sort_commits, false).expect("");
+			args.sort = Sort::from_str(sort_commits, true)
+				.expect("Incorrect config value for 'sort_commits'");
 		}
 	}
 	if !args.topo_order {
--- a/git-cliff/src/lib.rs
+++ b/git-cliff/src/lib.rs
@@ -6,7 +6,7 @@ pub mod changelog;
 #[macro_use]
 extern crate log;
 
-use args::Opt;
+use args::{Opt, Strip, Sort};
 use changelog::Changelog;
 use git_cliff_core::commit::Commit;
 use git_cliff_core::config::Config;
@@ -76,18 +76,19 @@ pub fn run(mut args: Opt) -> Result<()> {
 	}
 
 	// Update the configuration based on command line arguments and vice versa.
-	match args.strip.as_deref() {
-		Some("header") => {
-			config.changelog.header = None;
+	if let Some(strip) = args.strip {
+		match strip {
+			Strip::Header => {
+				config.changelog.header = None;
+			}
+			Strip::Footer => {
+				config.changelog.footer = None;
+			}
+			Strip::All => {
+				config.changelog.header = None;
+				config.changelog.footer = None;
+			}
 		}
-		Some("footer") => {
-			config.changelog.footer = None;
-		}
-		Some("all") => {
-			config.changelog.header = None;
-			config.changelog.footer = None;
-		}
-		_ => {}
 	}
 	if args.prepend.is_some() {
 		config.changelog.footer = None;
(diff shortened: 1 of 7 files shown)


orhun (changes requested):
Hello, thanks for the PR!

A couple of things caught my eye during review. It would be nice if you can make those changes. 😊 

Another important thing is, do you know if there is any setting to make `clap` not squash every option under "OPTIONS"? I really liked the previous look of `--help` since everything was placed under their appropriate sections and honestly it doesn't look too good now:

```
ARGS:
	...
OPTIONS:
	-v, --verbose ...
	-c, --config ...
```

This is how it used to look:

```
FLAGS:
	-v,  --verbose ....
OPTIONS
	-c, --config ...
ARGS:
	...
```

P.S. CI is failing due to formatting. Can you run nightly `rustfmt` and commit the changes? 🙂

tranzystorek-io:
Hi, thanks for having me!

> Another important thing is, do you know if there is any setting to make `clap` not squash every option under "OPTIONS"?

Clap now has a `help_heading` method/attribute for arguments that should have something else than the default `OPTIONS`. It won't be added to the usage string though.

git-cliff/src/lib.rs
```diff
@@ -100,9 +101,13 @@ pub fn run(mut args: Opt) -> Result<()> {
...
 	}
-	if args.sort == "oldest" {
+	if args.sort == Sort::Oldest {
 		if let Some(ref sort_commits) = config.git.sort_commits {
-			args.sort = sort_commits.to_string();
+			args.sort = match sort_commits.as_str() {
+				"oldest" => Sort::Oldest,
+				"newest" => Sort::Newest,
+				_ => unreachable!(),
+			};
```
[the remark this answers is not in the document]
tranzystorek-io: `FromStr` seems more suitable, or we could use the `ArgEnum::from_str()` method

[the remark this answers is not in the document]
orhun: `ArgEnum::from_str()` 💯

git-cliff/src/lib.rs
```diff
@@ -100,9 +105,9 @@ pub fn run(mut args: Opt) -> Result<()> {
 	if args.body.is_some() {
 		config.changelog.body = args.body;
 	}
-	if args.sort == "oldest" {
+	if args.sort == Sort::Oldest {
 		if let Some(ref sort_commits) = config.git.sort_commits {
-			args.sort = sort_commits.to_string();
+			args.sort = Sort::from_str(sort_commits, false).expect("");
```
orhun, line 110: I think it is safe to ignore case here. (`from_str`)

Also, instead of using expect without a message, I'd suggest using `if-let` or implementing `Default` for `Sort` and unwrapping that way. Like so:

```rs

impl Default for Sort {
	fn default() -> Self {
		Self::Oldest
	}
}

args.sort = Sort::from_str(sort_commits, true).unwrap_or_default();
```

Or something simpler can be applied too. But not empty `expect` 🙂

Can you please make this change? 😊

orhun (changes requested):

orhun:
I started a discussion on clap about the unified help messages: https://github.com/clap-rs/clap/discussions/3368

Until that's resolved, I'll hold this PR a bit.

git-cliff/src/lib.rs
tranzystorek-io: I'm wondering, if this is parsed from a config file, then doing a `unwrap_or_default` silently ignores incorrect values, is that desirable?

orhun: Hmm, that's a good point. It'd be best if we implement `De/Serialize` on `Sort` and use it in `GitConfig` really 😄 Then in this case you can use `expect` with a message at least 🙈 I think it's not possible to hit this error but still, for the sake of _code quality_...

tranzystorek-io: In my experience, making a common enum for both config files and clap flags creates coupling issues between the cli and config crates (here `git_cliff` and `git_cliff_core`. Instead, I'd rather create a separate `Sort` enum for `git_cliff_core` and then `impl From<git_cliff_core::Sort> for Sort`

orhun: Instead let's keep it simple and give this expect a message :)

orhun (approved):
Awesome, thanks for 🐻 ing with me!

Issue #109, opened 2022-09-21
"Repository is not owned by current user" after lastest commit for #108


ChrisB9:
**Describe the bug**
ERROR git_cliff > Git error: `repository path '/builds/project/path is not owned by current user; class=Config (7); code=Owner (-36)`

The error is caused by the change of directory and ownership + enabling safe directory in the Dockerfile by default in https://github.com/orhun/git-cliff/commit/4fc2217868fceea81ab5e6aeeb63ca719a07fe91 

**To Reproduce**

Using the following gitlab.yml configuration:
```
changelog:
  stage: changelog
  image:
    name: orhunp/git-cliff:latest
    entrypoint: [""]
  variables:
    GIT_STRATEGY: clone
    GIT_DEPTH: 0
```

**Expected behavior**
I know I am taking the risky path here and following the `latest` tag.
This is more of an informative bug report, but I believe this should have continued to work, as we automatically add the directory to the safe property of git.
Am I missing something here?

**System (please complete the following information):**
 - Docker: orhunp/git-cliff:latest from yesterday

orhun:
Hello, thanks for reporting this! 🐻 

> I know I am taking the risky path here and following the `latest` tag. This is more of an informative bug report, but I believe this should have continued to work, as we automatically add the directory to the safe property of git. Am I missing something here?

The workaround that I used for fixing #108 was creating a new user in the Docker image (as `git-user`) and changing the ownership of mounted volume while the container is being run. For doing that, I was copying the `.git` directory to a new place and utilizing this directory for running `git-cliff`. Unfortunately, using `chown(1)` wasn't applicable since it requires root permissions and also changes the ownership of the mounted volume outside the container if not being mounted as read-only.

```sh
# contents of entrypoint.sh

#!/bin/sh
cp -r /app /git-home/app
cd /git-home/app
exec git-cliff "$@"
``` 

This entrypoint script was required since the mounted volume was not owned by the container user. However, at the end of the day, it made the container run as non-root at all times (which is generally a good thing). This is where this GitLab CI issue originates from.

When you specify an image and empty entrypoint, GitLab runner _drops_ you to the shell of the image. For example, if you run `ls` you will see the current files in the repository and they are all owned by `root`.

![ls-screenshot](https://user-images.githubusercontent.com/24392180/192106463-bb591e13-5b6c-4993-81e6-dfd5b52d96c1.png)

[...]

Issue #108, opened 2022-09-16
Docker : config value 'safe.directory' was not found


siwon:
**Describe the bug**
Hi, 
Since git upgrade, the docker image return an error : 
```
ERROR git_cliff > Git error: `config value 'safe.directory' was not found; class=Config (7); code=NotFound (-3)`
```

My cmmand line is :
`docker run -it -v $(pwd)/.git:/app/ orhunp/git-cliff`

**Expected behavior**
Generation of my changelog.

**Screenshots/Logs**
Full log : 

```
ERROR git_cliff > Git error: `config value 'safe.directory' was not found; class=Config (7); code=NotFound (-3)`
```

**System (please complete the following information):**
I use your Docker image.
My code is on a Windows FS mounted on Debian WSL (so file chmod is 777).

**Additional context**
I try to add this to my git config with no effect :

```ini
[safe]
        directory = *
```

Thanks for your help

orhun:
Thanks for reporting this! It is directly related to [CVE-2022-24765](https://github.blog/2022-04-12-git-security-vulnerability-announced/) and should be fixed in 4fc2217868fceea81ab5e6aeeb63ca719a07fe91 🐻 

It is a regression in https://github.com/rust-lang/git2-rs/pull/839

PR #142, opened 2023-02-04
refactor(docker): avoid copying volume inside container

alerque:
A different way to work around the problems noted in #108, c.f trailing comments in #109.


alerque:
Untested because I have to run out, but this is roughly the approach I took on some other projects adapted for the way this project is laid out.

orhun:
Looks like `git` is not installed in `debian:buster-slim` image.

alerque:
Interesting. We can probably just echo the config to the file in `/etc`. My project is also a Rust + libgit2 one and it *reads* that config, but my Docker container also happened to have `git` available so that was an easy way to get setup.

Dockerfile
```diff
@@ -15,12 +15,27 @@ RUN cargo build --release --locked --no-default-features
...
-cp -r /app /git-home/app
-cd /git-home/app
-exec git-cliff "$@"
+if [[ $(id -u) -ne $(stat -c '%u' .) ]]; then
+  eids=$(stat -c '--euid %u --egid %g' .)
+fi
+exec ${eids:+setpriv --clear-groups $eids} git-cliff $@
 EOF
+
 ENTRYPOINT ["sh", "entrypoint.sh"]
```
orhun, line 41: I'm getting the following error when I try to run the container:

```
sh: 0: Can't open entrypoint.sh
```

Dockerfile
```diff
@@ -15,12 +15,27 @@ RUN cargo build --release --locked --no-default-features
 RUN rm -f target/release/deps/git_cliff*
 
 FROM debian:buster-slim as runner
+
+# Everything inside this container will be explicitly mounted by the end user,
+# so we can sidestep some Git security restrictions. This app recommends
+# mounting data to /app, but this *can* be changed externally and *will* be
+# changed when run by GitHub Actions, so we need to cover our bases.
+RUN echo -e '[safe]\n\tdirectory = *' > /etc/gitconfig
```
orhun, line 23: The contents of this file is the following in the container I built locally:

```
root@040df823c1dc:/app# cat /etc/gitconfig
-e [safe]
	directory = *
```

Thus, `git-cliff` fails:

```
root@040df823c1dc:/app# git-cliff
 WARN  git_cliff > "cliff.toml" is not found, using the default configuration.
 ERROR git_cliff > Git error: `failed to parse config file: invalid configuration key (in /etc/gitconfig:1); class=Config (7)`
```

orhun:
@alerque Can you check out if this actually works?

orhun:
Kind ping @alerque

website/docs/docker.md
```diff
@@ -15,13 +15,13 @@ Docker builds are [automated](https://github.com/orhun/git-cliff/tree/main/.gith
 The easiest way of running **git-cliff** (in the git root directory with [configuration file](/docs/configuration) present) is to use the available tags from [Docker Hub](https://hub.docker.com/r/orhunp/git-cliff):
 
 ```bash
-docker run -t -v "$(pwd)/.git":/app/ "orhunp/git-cliff:${TAG:-latest}"
+docker run -t -v "$(pwd)":/app/ "orhunp/git-cliff:${TAG:-latest}"
```
orhun, line 18: Isn't this breaking for users who were mounting `.git` before?

Dockerfile
```diff
@@ -15,12 +15,26 @@ RUN cargo build --release --locked --no-default-features
...
+# environments insist on running the entrypoint as root inside the container
+# even when being run by a non priviledged user on their own files. Here we
+# check the ownership of the workdir (which may or may not be /app) and change
+# our effective user/group ID to match.
+RUN cat <<'EOF' > /usr/local/bin/entrypoint.sh
 #!/bin/sh
-cp -r /app /git-home/app
-cd /git-home/app
-exec git-cliff "$@"
+if [ "$(id -u)" -ne "$(stat -c '%u' .)" ]; then
```
orhun, line 35: Do you think there is any side effects of this for the `git-cliff-action` or similar projects who depend on the Docker image?

See: https://github.com/orhun/git-cliff-action/blob/main/entrypoint.sh

website/docs/docker.md
alerque: I don't think so. The Git tooling mostly detects bare vs. full checkouts anyway. It was broken before because the Docker version when run this way didn't have any access to the config files. It only had access to the raw Git history and hence ignored any customization and use all defaults.

Dockerfile
alerque: That action can probably be fixed so it doesn't have to do any munching of repository file ownership before execution. That was always an ugly work around in the first place.

alerque: BTW, if you are using a Docker based runner there isn't much reason to have a separate repo for a GH Action, it works quite well (in spite of the docs) to have it just use the upstream repository *as* the action since it can pull everything prebuilt from GHCR anyway. I'm doing this on quite a number of projects (e.g. `luacheck`, `sile`, etc.).

orhun (approved):

Issue #131, opened 2022-12-25
Adaption of `gitoxide`


simonsan:
**Is your feature request related to a problem? Please describe.**
**Describe the solution you'd like**
git-cliff uses `git2`, I ask myself if there have been any thoughts already about adapting [gitoxide](https://github.com/Byron/gitoxide). 

**Additional context**
I see `gix tree entries` is able to expose the `commit messages` and `hashes`, so I assume it might be far enough, to be considered for this use case as well?

Disclaimer: I haven't looked deep into the source code to check which other features are used to fully understand if (early) adaption of `gitoxide` is easily possible.

orhun:
I didn't look into the source as well but I believe some day in the future `gitoxide` adoption will happen since it's a Rust implementation for handling git. On top of that, having [WASM support](https://github.com/Byron/gitoxide/issues/463) for `git-cliff` would be a huge benefit.

PR #162, opened 2023-04-28
chore(core): make git2 optional

MarcoIeni:
## Description

Make git2 optional in git-cliff-core crate.

## Motivation and Context

In release-plz I use git-cliff-core.
However, I don't need git-cliff to read the git history, so I don't need it to include git2.
I depend on both cargo and git-cliff, and often they have conflicting git2 versions.
For example, right now I can't update to the latest git-cliff, because git2 of git-cliff isn't compatible with cargo.

However with this change, by using:
`git-cliff-core =  { path = "../git-cliff/git-cliff-core", default-features = false }`
in release-plz I'm able to update!

Some questions for you:
- are you ok with this additional complexity?
- do you want to find a better name for the feature?

Probably there are other dependencies that can be made optional, but for now git2 is the biggest pain-point for me.
If you are interested, I can find more dependencies.
## How Has This Been Tested?

release-plz is compiling.
Maybe we can add a CI check to test that git-cliff-core compiles even without default features enabled.

## Types of Changes

- [ ] Other 

## Checklist:

## alternative

Extract the code that doesn't depend on git2 in a third crate.
In particular, in my use case I already have the commits I want to write in the changelog, so it would be great if there was a crate that is only focused on generating the changelog, not in determining the commits to write into it.
It would be great if it could expose a subset of the configuration. For example, in release-plz not all git-cliff configuration option will have effect.

### DIFF
--- a/git-cliff-core/src/lib.rs
+++ b/git-cliff-core/src/lib.rs
@@ -1,6 +1,9 @@
 //! A highly customizable changelog generator
 //!
 //! ## Features
+//!
+//! The [cargo features](https://doc.rust-lang.org/cargo/reference/features.html)
+//! of the libraries are:
 //! - `repo`: Enable parsing commits from a git repository. Enabled by default.
 //!   You can turn this off if you already have the commits to put in the
 //!   changelog and you don't need `git-cliff` to parse them.
--- a/git-cliff-core/src/lib.rs
+++ b/git-cliff-core/src/lib.rs
@@ -1,4 +1,9 @@
 //! A highly customizable changelog generator
+//!
+//! ## Features
+//! - `repo`: Enable parsing commits from a git repository. Enabled by default.
+//!   You can turn this off if you already have the commits to put in the
+//!   changelog and you don't need `git-cliff` to parse them.
 #![warn(missing_docs, clippy::unwrap_used)]
 
 /// Changelog generator.
--- a/git-cliff-core/src/lib.rs
+++ b/git-cliff-core/src/lib.rs
@@ -3,7 +3,7 @@
 //! ## Features
 //!
 //! The [cargo features](https://doc.rust-lang.org/cargo/reference/features.html)
-//! of the libraries are:
+//! of the library are:
 //! - `repo`: Enable parsing commits from a git repository. Enabled by default.
 //!   You can turn this off if you already have the commits to put in the
 //!   changelog and you don't need `git-cliff` to parse them.
--- a/git-cliff-core/src/lib.rs
+++ b/git-cliff-core/src/lib.rs
@@ -15,6 +15,7 @@ pub mod embed;
 pub mod error;
 /// Common release type.
 pub mod release;
+#[cfg(feature = "repo")]
 /// Git repository.
 pub mod repo;
 /// Template engine.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,11 @@ jobs:
         with:
           command: check
           args: --locked --verbose
+      - name: Check without default features
+        uses: actions-rs/cargo@v1
+        with:
+          command: check
+          args: --locked --no-default-features --verbose
 
   test:
     name: Test suite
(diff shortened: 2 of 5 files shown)


alerque:
I would ask what you can do with `git-cliff` without any Git history, but I guess I don't care. I would have no objection to a way to avoid libgit2 issues which (having used it in several projects and packaged it for many more) I know to be an ongoing pain!

This doesn't have anything to do with making git2 optional, but just to throw out there as a data point: the `gitoxide` crate is much easier to work with and maintain in projects. Some future refactor from an adventurous contributor might consider that replacement operation.

orhun:
> In release-plz I use git-cliff-core. However, I don't need git-cliff to read the git history, so I don't need it to include git2. I depend on both cargo and git-cliff, and often they have conflicting git2 versions. For example, right now I can't update to the latest git-cliff, because git2 of git-cliff isn't compatible with cargo.
> 
> However with this change, by using: `git-cliff-core = { path = "../git-cliff/git-cliff-core", default-features = false }` in release-plz I'm able to update!

Very reasonable!
 
> * are you ok with this additional complexity?

Yeah.

> * do you want to find a better name for the feature?

Maybe we can rename it to `git2` instead of `repo`. As @alerque mentioned, in the future, `git-cliff` might have `gitoxide` implementation (#131) for git-repo handling as well. If we rename this `git2` then it would make sense to name that feature `gitoxide`.

> Probably there are other dependencies that can be made optional, but for now git2 is the biggest pain-point for me. If you are interested, I can find more dependencies.

Sounds good! It would be nice to use `git-cliff` as a minimal library.

> Extract the code that doesn't depend on git2 in a third crate. In particular, in my use case I already have the commits I want to write in the changelog, so it would be great if there was a crate that is only focused on generating the changelog, not in determining the commits to write into it. It would be great if it could expose a subset of the configuration. For example, in release-plz not all git-cliff configuration option will have effect.

I think using a feature flag is okay for now.

---

One thing, maybe add a note to the documentation about available features and their description? Also, is the `docs.rs` documentation being built with all the features enabled? If not, can we enable that? I'm not exactly sure how that works.

orhun:
Okay I saw that you made other dependencies optional as well. Maybe it makes sense to go with `repo` feature now since it doesn't involve `git2` only? 🤔

MarcoIeni:
docs.rs should show default features by default. Since `repo` is a default feature, we should see it in docs.rs :+1:
I confirmed this by running the docs locally.
Here is the result of `cargo doc --open --no-deps`:
image

> Okay I saw that you made other dependencies optional as well. Maybe it makes sense to go with repo feature now since it doesn't involve git2 only? 🤔

Exactly. Plus, I see using `git2` or `gitoxide` a low level detail that the user shouldn't care about. So I would name the feature something like `git` instead of `git2` or `gitoxide`.
But `repo` is fine as well, because it enables parsing of a repository.

orhun:
> docs.rs should show default features by default. Since `repo` is a default feature, we should see it in docs.rs +1 I confirmed this by running the docs locally.

Nice, thanks for trying it out!

> 
> Exactly. Plus, I see using `git2` or `gitoxide` a low level detail that the user shouldn't care about. So I would name the feature something like `git` instead of `git2` or `gitoxide`. But `repo` is fine as well, because the feature enables parsing of a repository.

Yup, let's go with `repo` then.

MarcoIeni:
Yay! Btw do you have an idea why tests are failing? 😅

orhun:
Yeah, the latest pushed commit isn't conventional :D you can just ignore it :3

MarcoIeni:
ok, great :)
Then this is ready to be reviewed 👍

git-cliff-core/src/lib.rs
```diff
@@ -1,4 +1,12 @@
 //! A highly customizable changelog generator
+//!
+//! ## Features
+//!
+//! The [cargo features](https://doc.rust-lang.org/cargo/reference/features.html)
+//! of the libraries are:
```
orhun, line 6: nit: `libraries` -> `library`

orhun (approved):
bors r+

Issue #165, opened 2023-05-01
Re-license under the MIT + Apache 2.0 license


orhun:
To all the contributors of `git-cliff`! ⛰

I hope you're doing well. I'm reaching out to discuss a potential change in licensing for the project.

Currently, the project is licensed under [**GPL3**](https://www.gnu.org/licenses/gpl-3.0.en.html), which has certain restrictions on distribution and usage.

I believe that a re-licensing to [**MIT**](https://opensource.org/license/mit/) + [**Apache 2.0**](https://www.apache.org/licenses/LICENSE-2.0) license will provide several benefits to both the project and its users:

- Greater flexibility in terms of usage, distribution, and modification of the codebase.
- Compatibility with a wider range of software licenses, making it easier for other developers to incorporate or build upon the project.
- Reduced legal barriers to adoption by companies and organizations that have policies against using or contributing to GPL3-licensed projects.
- Improved accessibility to potential contributors who may be hesitant to contribute to a project with a more restrictive license.
- The adoption of widely recognized and accepted licenses will provide greater legal certainty and protection to users and contributors alike.

This is also the license of choice for the [Rust project](https://www.rust-lang.org/policies/licenses) itself and you can read more about their reasoning [here](https://github.com/rust-lang/rust/blob/master/COPYRIGHT).

The re-licensing can only be done if _all_ of the below authors consent to re-licensing their work under the terms of the MIT and Apache 2.0 license or if their contributions are _removed_.

The authors of the project have been derived using `git log --all --format='%aN <%aE>' | sort -u`:

```
Aaron Friel <mayreply@aaronfriel.com>
alteregoart <41084168+alteregoart@users.noreply.github.com>
bors[bot] <26634292+bors[bot]@users.noreply.github.com>
Cassandra McCarthy <cassie@7596ff.com>
dependabot[bot] <49699333+dependabot[bot]@users.noreply.github.com>
Eliza Weisman <eliza@buoyant.io>
Florian Fromm <flrn.frmm@gmail.com>
Grachev Mikhail <work@mgrachev.com>
Herby Gillot <herby.gillot@gmail.com>
Jan Katins <jasc@gmx.net>
Jeroen Claassens <jeroen.claassens@live.nl>
Kaushal Modi <kaushal.modi@gmail.com>
Kenji Miyake <31987104+kenji-miyake@users.noreply.github.com>
Lygon Bowen-West <l.bowenwest@googlemail.com>
Mack Solomon <macksol@gmail.com>
[...]

PR #200, opened 2023-06-25
fix(changelog): fix previous release references

tvcsantos:
## Description

Currently the previous pointer is not available under the unreleased group when running `git cliff`. Also when running `git cliff --unreleased` the preivous release is not correct and is pointing to the release before last, instead of the last tagged release.

This PR deals with this by doing the following. For tackling the first problem we simply check if we produced more then one release at the end of the loop and add the missing pointers to the previous release.

For solving the other problem we just adapt the `checked_sub` argument to take the difference into account and fix it on the unreleased scenario, when it must be `1` instead of `2`.

## Motivation and Context

Currently the previous pointer is not available under the unreleased group when running `git cliff`. Also when running `git cliff --unreleased` the preivous release is not correct and is pointing to the release before last, instead of the last tagged release. This PR deals with these problems. 

This is particularly useful for example on a Keep A Changelog format where we need the correct previous pointers so that we can add the proper link references.

## How Has This Been Tested?

I've added an extra fixture with a keep a changelog format example to test the scenario.

## Types of Changes

- [ ] Other 

## Checklist:


orhun (approved):
Neat implementation. Thank you!

bors r+

orhun:
Oh btw, can you comment on this issue if you haven't already? #165

PR #288, opened 2023-09-19
refactor: edit LICENSE

jxd1337:
Thanks for another fantastic tool to use! I've noticed that this LICENSE wasn't also updated, so I updated it.


orhun:
We're updating the license soon (hopefully) so this won't be needed I guess. See #165

jxd1337:
Sure, I didn't notice that. Will close this one.

PR #316, opened 2023-10-15
docs(changelog): fix typos

### LINKED ISSUE #310: Fix typos in the changelog

There are some typos in the changelog and it would be nice to fix them.

Unfortunately we can't just push a commit to fix them since the changelog is re-generated with each release.

1. Spot the typo commits with `typos` or `codespell`.
2. Update the default changelog configuration for pre-processing the commit messages.

Or use post-processors to update the final changelog.

**Additional context**

There is already a PR for this at #304 but seems inactive.

woshilapin:
fixes #310

## Description

Fix the typos in the changelog and automate typos correction in the changelog generation.

## Motivation and Context

#310

## How Has This Been Tested?

On my machine at the moment. I expect the next release to reveal if it breaks or not. Note that `release.sh` will now depend on `typos` being installed on the machine. Not sure if I should document that somewhere.

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/cliff.toml
+++ b/cliff.toml
@@ -65,6 +65,8 @@ split_commits = false
 # regex for preprocessing the commit messages
 commit_preprocessors = [
   { pattern = '\((\w+\s)?#([0-9]+)\)', replace = "([#${2}](<REPO>/issues/${2}))" },
+  # Check spelling of the commit with https://github.com/crate-ci/typos
+  # If the spelling is incorrect, it will be automatically fixed.
   { pattern = '.*', replace_command = 'typos --write-changes -' },
 ]
 # regex for parsing and grouping commits
--- a/cliff.toml
+++ b/cliff.toml
@@ -65,6 +65,7 @@ split_commits = false
 # regex for preprocessing the commit messages
 commit_preprocessors = [
   { pattern = '\((\w+\s)?#([0-9]+)\)', replace = "([#${2}](<REPO>/issues/${2}))" },
+  { pattern = '.*', replace_command = 'typos --write-changes -' },
 ]
 # regex for parsing and grouping commits
 commit_parsers = [
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -16,6 +16,8 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
+      - name: Check typos
+        uses: crate-ci/typos@master
       - name: Generate a changelog
         uses: orhun/git-cliff-action@main
         id: git-cliff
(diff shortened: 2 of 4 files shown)


cliff.toml
```diff
@@ -65,6 +65,7 @@ split_commits = false
 # regex for preprocessing the commit messages
 commit_preprocessors = [
   { pattern = '\((\w+\s)?#([0-9]+)\)', replace = "([#${2}](<REPO>/issues/${2}))" },
+  { pattern = '.*', replace_command = 'typos --write-changes -' },
```
orhun, line 68: Wow, this is really clever.  However it means that it will run `typos` for each commit thus it takes a bit time for me to generate the changelog. But I guess it is a good tradeoff for the sake of having an example of this cool trick.

One thing that we're missing is a comment that is explaining what is done here and linking to the `typos` repository 🐻

orhun:
Can you maybe update the release script to check if `typos` is installed? Something like this:

```sh
if ! command -v typos &>/dev/null; then
  echo "typos is not installed. Run 'cargo install typos' to install it otherwise the typos won't be checked"
fi
```

Also, one other thing to note, the CD workflows (`cd.yml`) use the same `git-cliff` config so it should also install `typos` for generating the correct changelog.

cliff.toml
woshilapin: That said, I'm now wondering if it's such a clever trick. If you try this, you'll understand why

```
echo "overrided" | typos --write-changes -
```

when 2 different fixes are possible, `typos` just bail out. It's not even putting the message on `stderr` and the word on `stdout` (which might have help to, at least, keep the word as-is). What's your take about that?

woshilapin: At the very least, I added a comment for the time being in [1c98c8f](https://github.com/orhun/git-cliff/pull/316/commits/1c98c8f4c9b64d05d569f4ae2caebd6b411656ad).

woshilapin:
> Can you maybe update the release script to check if `typos` is installed? Something like this:
> 
> ```shell
> if ! command -v typos &>/dev/null; then
>   echo "typos is not installed. Run 'cargo install typos' to install it otherwise the typos won't be checked"
> fi
> ```
> 
> Also, one other thing to note, the CD workflows (`cd.yml`) use the same `git-cliff` config so it should also install `typos` for generating the correct changelog.

@orhun I added both in [f1983ec](https://github.com/orhun/git-cliff/pull/316/commits/f1983ece31b6e79ce1dbfae249c95702d8e27fef).

:warning: Note however that using the Github Actions will install `typos` **AND** check the repository for spelling. Before doing the next release, we should fix all spelling errors. An alternative would be to manually install `typos` (`curl` on the released binary), but that would mean giving the version number in the download URL: this means more maintenance burden since each new version of `typos` would need an update of `cd.yml` (and `typos` has really frequent releases, about one per week I'd say).

cliff.toml
orhun: I just pushed 205cdbb391b03244abaf2e5e0651976544c78ff4, it should be fine now.

woshilapin: Oh sweet idea. Thank you.

orhun:
Oops, I think we need to install `typos` via `cargo install` instead of using the `typos` action in `cd.yml`:

```
 sh: 1: typos: not found
 TRACE git_cliff_core::changelog > 51f5e5b - IO error: `command exited with ExitStatus(unix_wait_status(32512))` (chore(img): add git-cliff logo)
thread '<unnamed>' panicked at git-cliff-core/src/command.rs:44:18:
Failed to write to stdin: Os { code: 32, kind: BrokenPipe, message: "Broken pipe" }
```

Logs: https://github.com/orhun/git-cliff/actions/runs/6683630522/job/18160024949

PR #339, opened 2023-11-05
feat: add tera filter for ordered commit groups

robjtede:
## Description

Add a Tera filter that produces commit groups, with order determined by `commit_parsers` config.

At least, that is the end goal, the PR does not yet achieve this. It's definitely a mess but  I'd like some feedback on the overall goal before progressing further.

## Motivation and Context

Using `group_by(attribute="group")` may not produce desired behavior if order of changelog section matters, eg. "Features" before  "Bug Fixes".

As far as I can tell, the most sensible place to source the ordering details is the `commit_parsers` config (it's what I initially tried to force sections to be ordered differently). It's not wired up like that yet, but hopefully this half of the task shows that it _can_ be done.

## How Has This Been Tested?

A new integration test has been added showing that ordering of sections is definable and not accidentally determined by the hash in a regular map.

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -434,6 +434,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "either"
+version = "1.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a26ae43d7bcc3b814de94796a5e736d4029efb0ee900c12e2d54c993ad1a1e07"
+
+[[package]]
 name = "encoding_rs"
 version = "0.8.33"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -622,6 +628,7 @@ dependencies = [
  "git2",
  "glob",
  "indexmap 2.1.0",
+ "itertools",
  "lazy-regex",
  "log",
  "next_version",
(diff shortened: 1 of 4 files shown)


orhun:
This looks good, the question is how can we naturally support this in a non-breaking way. As far as I understood, the user needs to use `commit_groups(groups=commit_groups_filter)` rather than `group_by(attribute="group")`. Is that enough?

robjtede:
Yea idea of the non-breaking path is to require a different filter. The current `groupBy` behavior would remain the same.

orhun:
Cool, how far are we from implementing that? I'm not sure if I quite understood the contents of this PR - are you planning to make that happen?

robjtede:
It's def not done in the PR yet. You can see the test case for a lil peek at what the PR currently _does_.

The goal is:

```
- {% for group, commits in commits | group_by(attribute="group") %}
+ {% for group, commits in commits | commit_groups %}
```

orhun:
ping 🐻

orhun:
closing for now, I might come back to this in the future

Issue #37, opened 2021-11-28
Add an option for processing only the current tag


kenji-miyake:
**Is your feature request related to a problem? Please describe.**

Currently, `--latest` option is supported, but it's a bit difficult to use.

Suppose this situation:
1. First, `v0.10.0-test` is released
2. After that, `v0.11.0-test` is released
3. After that, a bug was found and `v0.10.1-test` is released

![image](https://user-images.githubusercontent.com/31987104/143727499-11298197-7bac-4d61-bfef-4e6817bcf8ce.png)

Then the current behaviors are:

```sh-session
$ git checkout v0.10.0-test
$ git cliff | head
# Changelog
All notable changes to this project will be documented in this file.

## [0.10.0-test] - 2021-11-28

### Features

- Add A

## [0.9.0-test] - 2021-11-25

$ git cliff --latest
# Changelog
All notable changes to this project will be documented in this file.

## [0.10.1-test] - 2021-11-28

### Bug Fixes

- Fix A

<!-- generated by git-cliff -->
```

```sh-session
$ git checkout v0.11.0-test
$ git cliff | head
# Changelog
All notable changes to this project will be documented in this file.

## [0.11.0-test] - 2021-11-28

### Features

- Add B

## [0.10.0-test] - 2021-11-28

$ git cliff --latest
# Changelog
All notable changes to this project will be documented in this file.

## [0.10.1-test] - 2021-11-28

### Bug Fixes

- Fix A

<!-- generated by git-cliff -->
```

```sh-session
$ git checkout v0.10.1-test
$ git cliff | head
# Changelog
All notable changes to this project will be documented in this file.

## [0.10.1-test] - 2021-11-28

### Bug Fixes

- Fix A

## [0.10.0-test] - 2021-11-28

$ git cliff --latest
# Changelog
All notable changes to this project will be documented in this file.

## [0.10.1-test] - 2021-11-28

### Bug Fixes

- Fix A

<!-- generated by git-cliff -->
```

I want an option that extracts commits that are in only one release based on the current branch, but I guess there is nothing like that.

I believe it's useful for generating the body for GitHub Releases like you show [here](https://github.com/orhun/git-cliff-action#advanced).
Currently, you use `--latest` for this, but since what `--latest` points to will change, it's difficult to use correctly.

**Describe the solution you'd like**

I guess adding `--current` option can solve this problem.
It behaves like this:

```sh-session
$ git checkout v0.10.0-test
$ git cliff --current
# Changelog
[...]

PR #364, opened 2023-12-03
fix(changelog): fix previous version links

tvcsantos:
## Description

In this PR the main goal was to simplify and fix the previous version links on all scenarios.

## Motivation and Context

There were some scenarios where the previous version link was incorrect. For example

```
sh test-fixtures-locally.sh test-keep-a-changelog-links-current-arg --current
```

was producing the wrong link for the current tag, considering it as the only release.

## How Has This Been Tested?

Created several extra tests for fixtures to address the several combinations and provide a strong test suite to avoid regressions further on the line.

## Types of Changes

- [ ] Other 

## Checklist:


orhun:
Hey, thanks for creating this! 🐻 

I didn't quite get which issue this PR is solving. Can you elaborate what's the problem before and what was your approach for solving it? Giving example input/output would be really helpful since it is kind of difficult to infer it from the code.

tvcsantos:
Also a bit more in detail on the approach to solve the previous links in a uniform approach. The idea was the following:

1. Find the first tag that is processed by in the releases loop.
2. Get the tag before this one, that will in the end always in any scenario be the previous of the first release.
3. If no tag is ever processed then we are seeing commits after the last tag, hence previous is set to the last tag.

orhun:
I think there is a general confusion about the usage of `--latest` vs `--current`. I tried to explain the use-case for `--current` here: https://git-cliff.org/blog/git-cliff-0.5.0#--current

Also see #37 for the implementation details.

So in your case, if you want the previous release to be set, you can simply use `--latest`:

```bash
$ git cliff --context --latest | jq .[].previous

{
  "version": "v0.1.0",
  "commits": [],
  "commit_id": "0b5b2e2cb8c5e921e1c34b0ba60e442255d2a659",
  "timestamp": 0,
  "previous": null
}
```

I hope this helps, let me know! I think I can update the docs about clarifying the usage of `--current` / happily accept a PR about that! 🐻

orhun:
ping 🐻

tvcsantos:
Hey hey sorry for the delayed response.

Okay so let me use this example

suppose that I have 3 tags

v1.0.0
v2.0.0
v3.0.0

And now at this moment I'm on tag v3.0.0. What is the previous version of v3.0.0? It's v2.0.0

From what I got If I use git-cliff --current previous should not be set, is that it?!

orhun:
> And now at this moment I'm on tag v3.0.0. What is the previous version of v3.0.0? It's v2.0.0
> From what I got If I use git-cliff --current previous should not be set, is that it?!

I think there is a misunderstanding. What I'm saying is simply there is no need to use `--current` if you are already on the latest tag (v3.0.0), you should use `--latest` instead which correctly sets the previous version.

However, if you check out `v2.0.0` for example and use `--current`, then previous version should be set of course. Is this what this PR is fixing?

git-cliff/src/lib.rs
```diff
@@ -221,22 +236,13 @@ fn process_repository<'a>(
...
-			tags.len().checked_sub(sub).and_then(|v| tags.get_index(v))
-		{
-			let previous_release = Release {
-				commit_id: Some(commit_id.to_string()),
-				version: Some(version.to_string()),
-				..Release::default()
-			};
-			releases[0].previous = Some(Box::new(previous_release));
-		}
+		releases[0].previous = Some(Box::new(previous_release));
```
orhun, line 245: Why are we not checking if `args.latest` or `args.unreleased` is used? Does this mean this will override the previous release even though it is not needed?

.github/fixtures/test-keep-a-changelog-links-latest-arg/cliff.toml
```diff
@@ -0,0 +1,79 @@
+# configuration file for git-cliff (0.1.0)
+
```
orhun, line 2: We can probably remove the header from the keep-a-changelog related fixtures - we don't use it for the other config files as well.

tvcsantos: I will do that 👍

git-cliff/src/lib.rs
tvcsantos: Maybe what I can do as an improvement, and that is probably what you mean, is to only set previous if not already set?

.github/fixtures/test-keep-a-changelog-links-latest-arg/cliff.toml
orhun: Great!

orhun:
> And yes the PR is fixing the last issue you mentioned and with that also makes the code uniform for all cases, without requiring special ifs and the cumbersome check for tag lens.

That's nice! I guess we can confirm that the behavior isn't changed from the fixture tests as well.

git-cliff/src/lib.rs
orhun: Yeah, we can do that but I guess this is also fine.

orhun:
There are some conflicts with `main`, can you fix those and address the review comments for proceeding with this? 🐻

.github/fixtures/test-keep-a-changelog-links-latest-arg/cliff.toml
tvcsantos: Done :)

git-cliff/src/lib.rs
tvcsantos: Updated the code with that check also, to avoid the unnecessary performance hit 👍

tvcsantos:
All updated and merged with the last changes from main. Ty 💚

tvcsantos:
All seems good except for the failing "Continuous Integration / Links (pull_request)" that I think is not related. I've checked the failure and the links seem correct on the CHANGELOG 🤔

git-cliff/src/lib.rs
```diff
@@ -235,15 +239,24 @@ fn process_repository<'a>(
...
-		if let Some((commit_id, version)) =
-			tags.len().checked_sub(sub).and_then(|v| tags.get_index(v))
-		{
+	if !releases.is_empty() &&
+		releases
+			.first()
+			.and_then(|r| r.previous.as_ref())
+			.and_then(|p| p.version.as_ref())
+			.is_none()
+	{
```
orhun, line 248: Can you add a one-line comment here about what is being checked here? It will be helpful when we come back and look at this code some time later 🐻

orhun (commented):
Last comment 🐻

git-cliff/src/lib.rs
tvcsantos: Done 👍

orhun (approved):
Thanks for your contribution! 💖

Issue #382, opened 2023-12-13
Add support to ignore commits


schoetbi:
I just reverted a (conventional) commit. The problem is, that in the release notes this feature is still included. I have not found a way to exclude (ignore) a commit.

Would it be a good idea to add parsers that ignore previously added commits? For example, I have the following history.

- sha=a001 message=Start
- sha=a002 message=feat: a new feature
- ...
- sha=a003 message=revert a002 feat: a new feature

If git cliff supported revert parsers, I could do something like this

```toml
commit_parsers = [
 ...
 { message = "^revert\s+(?<sha>.{40})", revert="true" }
]
```

Another option would be to add the git SHA explicitly

```toml
commit_parsers = [
 ...
 { sha = "7976e3a76d3a27e8024d95eeb51d75bbd7391e95", skip="true" }
]
```

Or is there another option that I did not find in the documentation?

alerque:
I ended up with a similar situation before, made ever worse by the fact that it was a breaking change. The generated release notes not only had a giant `BREAKING CHANGE` warning, but also a revert for it. These are definitely things that should cancel each-other out, whether automatically or manually.

orhun:
I think supporting `sha` in `commit_parsers` is a good start so I implemented that in #385 - feel free to have a look / test it out.

About detecting revert commits (i.e. cancelling out commits), there are some pending questions:
- How do we know a commit is a revert commit?
  - Does `git2` (i.e. git internals) provide any information about that?
  - Do we need to need to manually parse the commit body ("Reverts ... <sha1>") to get the SHA1 of the reverted commit?
  - Can this be offloaded to the conventional commit parser?

alerque:
No, yes, maybe so.

As far as I know off the top of my head revert commits to *not* have any special meta data that can reliably be used other than parsing the commit message. It is not even guaranteed that they follow the commit message pattern, although being built into Git as a default message means a large portion of them will follow it. Only some weirdos like yours truly sometimes edit the messages with more info.

```
Revert {{ original_message }}.
This reverts commit {{ original_sha }}.
```

The pattern you probably want to search for is not even the first line of the subject "Revert" but the "reverts commit <sha>" bit from the full body.

orhun:
> The pattern you probably want to search for is not even the first line of the subject "Revert" but the "reverts commit " bit from the full body.

Yeah I think that should be easily _parseable_.

[...]

Issue #117, opened 2022-10-04
Allow multiple parsers for skipped/non-skipped commits


orhun:
**Is your feature request related to a problem? Please describe.**

From PR #114:

> Note that during testing, one caveat was discovered: If a commit is matched by one parser, it can't be matched by any other. That means if a breaking commit was prevented from being skipped, it won't become part of any other group. Maybe it is worth to think about allowing multiple parsers for commits in an hierarchical order?

**Describe the solution you'd like**
Allow commits to be parsed by multiple `commit_parser`s (if skipped). 

**Describe alternatives you've considered**
None.

**Additional context**
I think this issue needs some brainstorming. Let me know if you want to tackle it!

sharmadhiraj86:
Hi I want to contribute this issue. 
Can you let me know how I can approach this issue.

tonny008:
I need this feature even it is not skipped. 

I thought below config will work. but it NOT. only the first match takes effect.

```
  [
{ message = "(scope)", scope = "Fancy Scope" },
{ message = "^feat", group = "<!-- 0 -->🚀 Features" }
]
```
to make it work. i need below parser. it is tedious with many scopes and groups.

```
[{ message = "^feat\(scope\)", scope = "Fancy Scope", group =  "<!-- 0 -->🚀 Features"}]
```

ognis1205:
Following.

PR #385, opened 2023-12-14
feat(parser): support using SHA1 of the commit

orhun:
## Description

See #382

## Motivation and Context

This PR makes it possible to use the SHA1 of a commit with `commit_parsers`.

For example:

```toml
# skip a specific commit
commit_parsers = [
    { sha = "f6f2472bdf0bbb5f9fcaf2d72c1fa9f98f772bb2", skip = true }
]
```

Another example:

```toml
# override the group of a specific commit
commit_parsers = [
    { sha = "f6f2472bdf0bbb5f9fcaf2d72c1fa9f98f772bb2", group = "Stuff" }
]
```

## How Has This Been Tested?

Unit/integration tests.
I didn't add fixture tests since it requires more work to find out the SHA1 of the created commit and substitue in `expected.md`.

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/git-cliff-core/src/changelog.rs
+++ b/git-cliff-core/src/changelog.rs
@@ -249,6 +249,40 @@ mod test {
 				}]),
 				commit_parsers:           Some(vec![
 					CommitParser {
+						sha:           Some(String::from("tea")),
+						message:       None,
+						body:          None,
+						group:         Some(String::from("I love tea")),
+						default_scope: None,
+						scope:         None,
+						skip:          None,
+						field:         None,
+						pattern:       None,
+					},
+					CommitParser {
+						sha:           Some(String::from("coffee")),
+						message:       None,
+						body:          None,
+						group:         None,
+						default_scope: None,
+						scope:         None,
+						skip:          Some(true),
+						field:         None,
+						pattern:       None,
+					},
+					CommitParser {
+						sha:           Some(String::from("coffee2")),
+						message:       None,
+						body:          None,
+						group:         None,
+						default_scope: None,
+						scope:         None,
+						skip:          Some(true),
+						field:         None,
+						pattern:       None,
+					},
+					CommitParser {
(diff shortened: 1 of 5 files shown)


schoetbi:
Awesome that you implemented this feature. It really helps me! What I noted when trying this out is that the order of the commit parsers matter. If I placed the sha skips at the very beginning it worked. When I moved them to the last position in the list it did not filter out these commits even though the git hashes matched. Is there a reason for this?

Also, the shas should match even if the casing of the letters is not the same as git delivers it.
These two lines should behave the same

```toml
commit_parsers = [
  { sha = "7ff168fbac5bf5dfc772e14e1ff23e62a338aab8", skip = true },
  { sha = "7FF168FBAC5BF5DFC772E14E1FF23E62A338AAB8", skip = true },
]
```

orhun:
Hey, thanks for testing it!

> When I moved them to the last position in the list, it did not filter out these commits, even though the git hashes matched. Is there a reason for this?

It is simply because the `commit_parsers` are processed in order and if a commit is matched by a parser it can't be parsed by another one. There is a related issue here: #117

Maybe I can note that the order matters in the documentation, or even better, support multiple parsers to be used.

> Also, the SHAs should match, even if the casing of the letters is not the same as git delivers it.

Good catch! Should be fixed in d698809a7a8ce1ce8d1341eacfd3135df26c8884

> Since this list might get lengthy over time, it might make sense to add a skiplist file to the configuration. 

Can you create an issue for discussing this feature? We can potentially have this configuration:

```toml
# cliff.toml
[git]
commit_parsers_file = "parsers.toml"
```

```toml
# parsers.toml
commit_parsers = [
    { sha = "f6f2472bdf0bbb5f9fcaf2d72c1fa9f98f772bb2", group = "Stuff" }
]
```

PR #403, opened 2023-12-22
fix(commit): trim the trailing newline from message

vbx:
When the commit message ends with a '\n', then it is not considered a valid conventional commit. That's why the message must be cleaned up.

## Description and Motivation and Context
Wrong next version when using the bumped-version parameter. The 'git commit' command creates a '\n' at the end of the line.
In this case, commit is considered as a patch. 
[https://github.com/MarcoIeni/release-plz/blob/4b0684ff8d9048032167a43fd70b70af4940239d/crates/next_version/src/next_version.rs#L27](https://github.com/MarcoIeni/release-plz/blob/4b0684ff8d9048032167a43fd70b70af4940239d/crates/next_version/src/next_version.rs#L27)
```
    /// - If some commits are present, but none of them match conventional commits specification,
    ///   the version is incremented as a Patch.
```

## How Has This Been Tested?

git cli command on linux/macos

## Types of Changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Other 

## Checklist:

### DIFF
--- a/git-cliff-core/src/release.rs
+++ b/git-cliff-core/src/release.rs
@@ -39,7 +39,7 @@ impl<'a> Release<'a> {
 			.next(
 				self.commits
 					.iter()
-					.map(|commit| commit.message.to_string())
+					.map(|commit| commit.message.trim_end().to_string())
 					.collect::<Vec<String>>(),
 			)
 			.to_string();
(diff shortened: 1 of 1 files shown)


orhun:
Hey, thanks for the PR! 🐻 

Do you think this might have side effects? It should be backwards compatible I reckon.

Also, looks like the formatting job has failed, can you run `cargo fmt`?

vbx:
no side effects, i think. 'cargo fmt' is ok. i rebased branch also. PR is up to date.

orhun (approved):
Thanks!

nachtien:
nice work!

Issue #514, opened 2024-02-25
bug: GitHub Integration does not work on Windows when running through NodeJS package


favna:
**Describe the issue**

When running through a NodeJS installed version of `git-cliff` (`npx git-cliff`) the GitHub integration is not activated/read/executed/similar despite all config being the same between the NodeJS based execution and the Cargo based execution.

This is an issue for the related project [favware/cliff-jumper](https://github.com/favware/cliff-jumper) because I cannot give users access to this awesome new feature until this is resolved.

**To reproduce**

1. Clone https://github.com/favware/cliff-jumper
1. Setup NodeJS and Yarn
1. Run `yarn install` in the repository
1. (Optional): If you want to see the git-cliff command executed add to `src/commands/update-changelog.ts` above line 25 the line
```ts
console.log(`npx git-cliff --tag ${options.tagTemplate} --prepend ./CHANGELOG.md -u -c ./cliff.toml ${githubConfig} ${monoRepoConfig}`);
```
5. Build the TypeScript code with `yarn build`
6. Run the code with `node . --skip-tag`

**Expected behavior**

1. GitHub integration works through the NodeJs based installation as well
4. The CHANGELOG is updated with a version 3.0.0 and the CHANGELOG.md has PR links for the relevant commits, expected markdown:

```md
# [@favware/cliff-jumper@3.0.0](https://github.com/favware/cliff-jumper/compare/@favware/cliff-jumper@2.2.3...@favware/cliff-jumper@3.0.0) - (2024-02-25)

## 🐛 Bug Fixes

- **deps:** Update dependency git-cliff to v2 ([508b7ad](https://github.com/favware/cliff-jumper/commit/508b7adbf04d79e18a51abcd07b6b69c59d142a7)) ([#157](https://github.com/favware/cliff-jumper/pull/157))
  - 💥 **BREAKING CHANGE:** Bumped git-cliff to v2, please see https://github.com/orhun/git-cliff/blob/main/CHANGELOG.md#200---2024-02-19 for their breaking changes
  - 💥 **Co-authored-by:** renovate[bot] <29139614+renovate[bot]@users.noreply.github.com>
- **deps:** Update dependency commander to v12 ([dd7684e](https://github.com/favware/cliff-jumper/commit/dd7684e27216d270e859f0e683efae2347602f52)) ([#151](https://github.com/favware/cliff-jumper/pull/151))
[...]

PR #528, opened 2024-03-03
chore(npm): publish rc version for prereleases

favna:
## Description

I wouldn't call myself a Bash legend but I think this does the trick. FWIW I tested it using a Bash session, but I can't say for 100% certain the `""` wrapping around the injected value from the GitHub workflow steps is correct, or if GitHub now adds double double quotes (as in `if [ ""true"" = "true" ]`). 

As for the change on line 252, that's mostly just cleaned and moved to `package.json.tmpl`. Note that the binaries cannot be published with `yarn npm publish` because that package.json doesn't reside in a yarn configured folder and we probably don't want that anyway.

I looked at the history for `npm/git-cliff/package.json` and found this commit: https://github.com/orhun/git-cliff/commit/42180437c709f10b9ed5efb8f5cc1595efb3aec7#diff-09031284f88e0ed62cdbc8693d473bf5b8b9c5674257763d8d8ca0cebfc87a44. I see that you manually set the version to an `rc` version and I assume this commit then triggered this workflow so that should make it all work.

## Motivation and Context

Requested in this comment: https://github.com/orhun/git-cliff/issues/514#issuecomment-1975264758

## How Has This Been Tested?

I tested the individual steps using a Bash terminal session locally:

image

N.A.

## Types of Changes

- [x] Other: Continuous delivery

## Checklist:

- [x] ~~I have updated the documentation accordingly.~~ N.A.
- [x] ~~I have formatted the code with [rustfmt](https://github.com/rust-lang/rustfmt).~~ N.A.
- [x] ~~I checked the lints with [clippy](https://github.com/rust-lang/rust-clippy).~~ N.A.
- [x] ~~I have added tests to cover my changes.~~ N.A.
- [x] ~~All new and existing tests passed.~~ N.A.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -300,12 +300,6 @@ jobs:
         with:
           node-version: 18
           registry-url: "https://registry.npmjs.org"
-      - name: Check if the release is a pre-release
-        working-directory: npm/git-cliff
-        shell: bash
-        id: check_pre_release
-        run: |
-          echo "IS_PRE_RELEASE=$([[ $(jq -r '.version' package.json | grep -E '\-rc.[0-9]+') ]] && echo "true" || echo "false")" >> $GITHUB_OUTPUT
       - name: Publish the package
         shell: bash
         working-directory: npm/git-cliff
@@ -313,7 +307,7 @@ jobs:
           yarn install
           yarn build
           cp ../../README.md .
-          if [ "${{ steps.check_pre_release.outputs.IS_PRE_RELEASE}}" = "true" ]; then
+          if [ ${{ contains(github.ref, '-') }} = "true" ]; then
             yarn npm publish --tag rc
           else
             yarn npm publish
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -249,7 +249,7 @@ jobs:
           cp "../target/${{ matrix.build.TARGET }}/release/${bin}" "${node_pkg}/bin"
           cp ../README.md "${node_pkg}"
           cd "${node_pkg}"
-          npm publish --access public
+          npm publish
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
       - name: Build Python wheels (linux)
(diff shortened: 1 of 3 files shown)


.github/workflows/cd.yml
```diff
@@ -300,14 +300,24 @@ jobs:
...
       - name: Publish the package
         shell: bash
+        working-directory: npm/git-cliff
         run: |
-          cd npm/git-cliff
           yarn install
           yarn build
           cp ../../README.md .
-          npm publish --access public
+          if [ "${{ steps.check_pre_release.outputs.IS_PRE_RELEASE}}" = "true" ]; then
```
orhun, line 316: We can probably just do:

```yml
${{ contains(github.ref, '-') }}
```

favna:
I'll be honest I didn't know you could use that syntax within a bash block lol 

image
image

```yml
      - name: Testing
        shell: bash
        run: |
          echo ${{ github.ref }}
          echo ${{ contains(github.ref, '-') }}
          if [ ${{ contains(github.ref, '-') }} = "true" ]; then
            echo "This is a prerelease"
          else
            echo "This is a release"
          fi
```

orhun (approved):
LGTM!

.github/workflows/cd.yml
```diff
@@ -302,12 +302,16 @@ jobs:
...
         shell: bash
+        working-directory: npm/git-cliff
         run: |
-          cd npm/git-cliff
           yarn install
           yarn build
           cp ../../README.md .
-          npm publish --access public
+          if [ ${{ contains(github.ref, '-') }} = "true" ]; then
+            yarn npm publish --tag rc
```
orhun, line 311: Ah, one question: is this going to tag the release as "rc"? If so, what happens if it tries to overwrite it?

favna: If by overwrite you mean a new RC version then the new version gets assigned the tag and it's unassigned from the previous one. In fact, under the hood the lack of --tag sets the tag to "latest" so the same mechanism applies for that as for rc.

orhun: got it, thanks for the explanation!

orhun:
Hey @favna, I just created a RC version, but the NPM step failed: <https://github.com/orhun/git-cliff/actions/runs/8412062961/job/23032878338>

Was wondering if you have any about this 🤔

favna:
Woops, my bad. Created #574 that will fix that.

PR #530, opened 2024-03-04
feat(release): allow preserving zero-based versions

### LINKED ISSUE #447: Make the bump version rules configurable

The new `bump` flag is great! However, since 4eef684c568ad16357e5d256180a51b1a46dd0cb the default behavior for zero-based releases is problematic, i.e. features always bump the minor version, while breaking changes bump the major version, including from 0 to 1. I think this is an undesired behavior for projects following the bump rules described in [next_version](https://docs.rs/next_version/latest/next_version/).

I think it should be possible to opt out from the default behavior. That is, the hardcoded values in

https://github.com/orhun/git-cliff/blob/4eef684c568ad16357e5d256180a51b1a46dd0cb/git-cliff-core/src/release.rs#L117-L118

should take values form the configuration instead.

I couldn't think of any alternative (other than forking the project).

**Additional context**

If you would be interested, I could write a PR for it.

jsurkont:
closes #447

## Description

This change adds a customization for version bumping. Specifically, it allows changing the default version bumping rules in order to preserve zero-based releases.
 

## Motivation and Context

The current bumping rules are hard-coded, so it's necessary to extend the input arguments to allow for customization.

This PR adds a new CLI flag `--keep-zerover`. I also considered extending the configuration. However, it wasn't clear to me which section would be appropriate. Perhaps `changelog`? I think that having it in the configuration would be quite convenient to users. It would also mean that the signature of `Changelog::bump_version` doesn't need to be modified. What do you think?

This PR closes #447.

## How Has This Been Tested?

I extended the unit tests for `calculate_next_version`. I also run `git cliff` with and without the new flag.

## Types of Changes

I wasn't sure if this should be considered a breaking change, but I've marked is a such to be on the safe side. My guess is that there wouldn't be a large number (if any) of users who use the current bump version feature with zero-based releases, as it would result in moving away from the zero-based system. 

- [ ] Bug fix (non-breaking change which fixesan issue)
- [ ] Other 

## Checklist:


orhun:
I'm thinking if we should offer this as a config option (also environment variable) rather than a command-line argument. I don't think this will be used very often especially when the user is only interested in changelog generation rather than bumping the version. What do you think?

jsurkont:
I agree. I guess that, for the majority of projects, adding this option to the configuration (or using an environmental variable) would be the most convenient way.

Do you have a preference under which config section this option should be added?

orhun:
I think the best option would be to add a new section as follows:

```toml
[bump]
# Configures automatic minor version increments for feature changes.
# When true, a feature will always trigger a minor version update.
# When false, a feature will trigger:
# - a patch version update if the major version is 0.
# - a minor version update otherwise.
features_always_bump_minor = true
# Configures 0 -> 1 major version increments for breaking changes.
# When true, a breaking change commit will always trigger a major version update
# (including the transition from version 0 to 1)
# When false, a breaking change commit will trigger:
# - a minor version update if the major version is 0.
# - a major version update otherwise.
breaking_always_bump_major = true
```

What do you think?

jsurkont:
Yes, I think this is the cleanest solution ✨ 

I've just updated my PR to use the new config section. I also changed the code so it doesn't change the pubic api of `git-cliff-core` at a (small) cost of function repetition. Unfortunately, the "Semver" check is still failing because of adding a new field to the `Config` struct.

website/docs/configuration/bump.md
orhun, line 1: Can we also mention the existence of this configuration option in `usage/bump-version.md`?

orhun (commented):
Looks good! Can you also [add a test fixture](https://github.com/orhun/git-cliff/tree/main/.github/fixtures)? There is also one example under `test-bump-version`.

website/docs/configuration/bump.md
jsurkont: Thanks for your feedback!

I added a new fixture and extended the `usage` docs so they now mention the new configuration options.

orhun: Awesome, thanks!

orhun (approved):
Awesome, thanks for your contribution!

PR #543, opened 2024-03-08
Update index.tsx

ganeshwhere:
I have made some changes to how a certain part of a web page looks and behaves. Specifically, I make sure that a section called "Feature" has the right style applied to it. I also make sure that any text shown in that section is centered on the page. To do this, I first wrapped the text inside the "Feature" section in a paragraph (like how you'd write a paragraph in a document). Then, I applied some special styles to this section to make sure the text is centered and looks the way we want it to. So now, when you look at that part of the web page, you'll see the text nicely centered and formatted, just like you intended.


orhun:
I guess this change is for:

image

vs.

image

I think I like the first version better and I didn't really understood how it is "just like I intended."

PR #545, opened 2024-03-08
fix(release): fix typo in the installation of typos-cli

radusuciu:
While creating releases locally I followed the directions from `release.sh` to run `cargo install typos` but I'm pretty sure that the correct command is `cargo install typos-cli`.


orhun (approved):
Good catch!

Issue #534, opened 2024-03-05
BadZipFile error while installing via pip


karl0ss:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description of the bug

Unable to install git-cliff via pip

### Steps To Reproduce

Brand new python venv 
```
❯ python -m venv venv
❯ venv/Scripts/activate
❯ pip install setuptools wheel
Requirement already satisfied: setuptools in c:\users\karl.hudgell\.pyenv\pyenv-win\versions\3.10.6\lib\site-packages (63.2.0)
Requirement already satisfied: wheel in c:\users\karl.hudgell\.pyenv\pyenv-win\versions\3.10.6\lib\site-packages (0.37.1)

[notice] A new release of pip is available: 23.3.2 -> 24.0
[notice] To update, run: python.exe -m pip install --upgrade pip
❯ python.exe -m pip install --upgrade pip

zsh: command not found: python.exe
❯ python -m pip install --upgrade pip

Requirement already satisfied: pip in c:\users\karl.hudgell\.pyenv\pyenv-win\versions\3.10.6\lib\site-packages (23.3.2)
Collecting pip
  Downloading pip-24.0-py3-none-any.whl.metadata (3.6 kB)
Downloading pip-24.0-py3-none-any.whl (2.1 MB)
   ---------------------------------------- 2.1/2.1 MB 33.9 MB/s eta 0:00:00
Installing collected packages: pip
  Attempting uninstall: pip
    Found existing installation: pip 23.3.2
    Uninstalling pip-23.3.2:
      Successfully uninstalled pip-23.3.2
Successfully installed pip-24.0
```

### Expected behavior

Git-Cliff should install

### Screenshots / Logs

```
❯ pip install git-cliff
Collecting git-cliff
  Downloading git_cliff-2.1.2-py3-none-win_amd64.whl.metadata (6.1 kB)
Downloading git_cliff-2.1.2-py3-none-win_amd64.whl (6.4 MB)
   ---------------------------------------- 6.4/6.4 MB 14.1 MB/s eta 0:00:00
Installing collected packages: git-cliff
ERROR: Exception:
Traceback (most recent call last):
  File "C:\Users\Karl.Hudgell\.pyenv\pyenv-win\versions\3.10.6\lib\site-packages\pip\_internal\cli\base_command.py", line 180, in exc_logging_wrapper
    status = run_func(*args)
[...]

PR #546, opened 2024-03-08
feat(ci): test that python wheels can be built and installed

radusuciu:
## Description

This adds a CI job that builds and installs python wheels for every platform that wheels are published to PyPI.

## Motivation and Context

This should warn against issues such as #534.

## How Has This Been Tested?

Please refer to GHA runs of this workflow in my fork: https://github.com/radusuciu/git-cliff/actions/workflows/ci.yml

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -256,9 +256,7 @@ jobs:
           # https://github.com/PyO3/maturin-action/issues/245
           manylinux: ${{ matrix.build.TARGET == 'aarch64-unknown-linux-gnu' && '2_28' || 'auto' }}
       - name: Build Python wheels (macos & windows)
-        if: |
-          matrix.build.PYPI_PUBLISH == true &&
-          (startsWith(matrix.build.OS, 'macos') || startsWith(matrix.build.OS, 'windows'))
+        if: (startsWith(matrix.build.OS, 'macos') || startsWith(matrix.build.OS, 'windows'))
         uses: PyO3/maturin-action@v1
         with:
           working-directory: pypi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -160,3 +160,121 @@ jobs:
             printf "Checking MSRV for $package..."
             cargo msrv --output-format json --path "$package" verify | tail -n 1 | jq --exit-status '.success'
           done
+
+  wheels:
+    name: Test that python wheels build and install
+    runs-on: ${{ matrix.build.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        build:
+          - {
+              NAME: linux-x64-glibc,
+              OS: ubuntu-22.04,
+              TOOLCHAIN: stable,
+              TARGET: x86_64-unknown-linux-gnu,
+            }
+          - {
+              NAME: linux-x64-musl,
+              OS: ubuntu-22.04,
+              TOOLCHAIN: stable,
+              TARGET: x86_64-unknown-linux-musl,
+            }
+          - {
+              NAME: linux-x86-musl,
+              OS: ubuntu-22.04,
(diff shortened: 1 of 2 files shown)


radusuciu:
It almost works:

https://github.com/radusuciu/git-cliff/actions/runs/8207985679/

The platforms that fail. fail because of the same reason:

```
Defaulting to user installation because normal site-packages is not writeable
Looking in links: pypi/wheels
Processing ./pypi/wheels/git_cliff-2.1.2.tar.gz
  Installing build dependencies: started
  Installing build dependencies: finished with status 'error'
  error: subprocess-exited-with-error
  
  × pip subprocess to install build dependencies did not run successfully.
  │ exit code: 1
  ╰─> [3 lines of output]
      Looking in links: pypi/wheels
      ERROR: Could not find a version that satisfies the requirement maturin<2,>=1.5 (from versions: none)
      ERROR: No matching distribution found for maturin<2,>=1.5
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error

× pip subprocess to install build dependencies did not run successfully.
│ exit code: 1
╰─> See above for output.

note: This error originates from a subprocess, and is likely not a problem with pip.
Error: Process completed with exit code 1.
```

Relevant line:

```
ERROR: Could not find a version that satisfies the requirement maturin<2,>=1.5 (from versions: none)
```

I think this may be because maturin isn't available for all of the same platforms that are listed in the matrix here. In any case, I think we don't actually need maturin to be installed.. I'll need to experiment with this more later, but I think this is a good base.

orhun:
Yup, the workflow looks good. Let's see if we need maturin installed. Also, do you think it is a good idea to test it for all the targets or testing it for each platform (Linux/MacOS/Windows) is enough?

radusuciu:
> Yup, the workflow looks good. Let's see if we need maturin installed. Also, do you think it is a good idea to test it for all the targets or testing it for each platform (Linux/MacOS/Windows) is enough?

I'm not sure, it does seem like testing each platform is overkill. Maybe we can check out what other rust projects that use maturin are doing?

orhun:
Yeah, I agree. Checking out other projects is definitely a good idea.

orhun:
I [looked around](https://github.com/search?q=maturin-action&type=code&p=2) in GitHub a bit and some projects build for each target and some projects just test the platform. I think we can have this check running for each platform as well.

davidhewitt:
Whenever you see the line

> Processing ./pypi/wheels/git_cliff-2.1.2.tar.gz

It means that the wheel wasn't installable on the platform you're testing on, probably because you built for a different platform than the host interpreter running on GitHub actions. As a result, CI is trying to extract the source tarball and build it. You can prevent this fallback with `--only-binary` flag to pip if I recall correctly. Might be worth adding that flag given you're trying to check the wheel.

To match the python to test on to the wheel, you might have to use something like https://github.com/uraimo/run-on-arch-action to run your tests.

radusuciu:
> Whenever you see the line
> 
> > Processing ./pypi/wheels/git_cliff-2.1.2.tar.gz
> 
> It means that the wheel wasn't installable on the platform you're testing on, probably because you built for a different platform than the host interpreter running on GitHub actions. As a result, CI is trying to extract the source tarball and build it. You can prevent this fallback with `--only-binary` flag to pip if I recall correctly. Might be worth adding that flag given you're trying to check the wheel.
> 
> To match the python to test on to the wheel, you might have to use something like https://github.com/uraimo/run-on-arch-action to run your tests.

Thanks @davidhewitt for that insight! I think we can fix that up.

orhun:
Any chance that we get this done before the next release? 😊 @radusuciu

orhun:
Closing due to inactivity. Feel free to come back to this later :)

PR #569, opened 2024-03-22
feat(changelog): support parsing commits by footer

jan-ferdinand:
## Description

Enable footer-based commit parsing. The idea is to add functionality like

```toml
commit_parsers = [
  { footer = "Hello: world", skip = true },
]
```

I would appreciate some guidance – is this useful, desired, worth it, …?

## Motivation and Context

In order to keep my changelogs concise, I would like to ignore certain commits. I would like to specify whether a commit should be ignored in the commit message. A somewhat elegant way to achieve this (I think) would be to include a footer like `changelog: ignore`.
(If you have a better way to approach this, I'm happy to learn of it.)

## How Has This Been Tested?

Append existing tests that create sample changelogs.

The test `changelog_generator_split_commits` is failing. It appears that a footer's value (but not the key) makes its way into the changelog. This might be because the footer's key is interpreted as a conventional commit type. I would appreciate some guidance here.

## Possible Limitations

The conventional commits' specification and [the reference implementation of its parser](https://github.com/conventional-commits/parser) are currently out of sync. This might mean that the commit parser `{footer = "Hello: world"}` does not work (even though it should), whereas the commit parser `{footer = "Hello:world"}` does work (even though omitting the space goes against the specification).

Some relevant issues concerning footer parsing are: 
- https://github.com/conventional-commits/parser/issues/47
- https://github.com/conventional-commits/parser/issues/41
- https://github.com/crate-ci/git-conventional/pull/54

## Types of Changes

- [ ] Other 

## Checklist:


orhun:
Thanks for the PR!

> I would appreciate some guidance – is this useful, desired, worth it, …?
> A somewhat elegant way to achieve this (I think) would be to include a footer like changelog: ignore.

I'm wondering if the same can be achieved with `body` which `git-cliff` already supports filtering.

I'm also wondering if this is useful for #382. But I'm guessing "This commit reverts" is not a part of the footer.

> The test changelog_generator_split_commits is failing. It appears that a footer's value (but not the key) makes its way into the changelog. This might be because the footer's key is interpreted as a conventional commit type. I would appreciate some guidance here.

That is the expected behavior so I recommend you to either add a new `CommitParser` to `changelog_generator_split_commits` to skip that commit or update the expected result accordingly to include that commit in the changelog.

Overall, I think this might be useful. Can you update the documentation about the existence of `footer` filter and also [add a test fixture](https://github.com/orhun/git-cliff/tree/main/.github/fixtures) to verify it's behavior?

orhun:
ping @jan-ferdinand 🙂

jan-ferdinand:
Other things have come up and are more pressing right now; I'll come back to this. Apologies if that creates inconveniencies.

orhun:
No worries!

jan-ferdinand:
Thanks for ping & patience. I have made the following changes, most of which were requested or guided towards:

- rebase onto current `main`
- fix test `changelog_generator_split_commits`
- add fixture
- add some documentation

I would like to have some additional guidance on the following points:

- Is the amount of documentation enough? It seems to be very little, but I couldn't readily identify more places where it makes sense to add additional documentation.
- The conventional commit reference parser being [out of sync with the specification](https://github.com/conventional-commits/parser/issues/47) is annoying for this PR. It requires users that want to use the suggested feature to write parsers of the form `{ footer = "^key:value }` (which goes against the spec), or, to be a bit more future proof `{ footer = ^key: ?value }` (which is confusing). Is this acceptable?

git-cliff-core/src/changelog.rs
```diff
@@ -907,6 +939,13 @@ mod test {
 		config.git.split_commits = Some(true);
 		config.git.filter_unconventional = Some(false);
 		config.git.protect_breaking_commits = Some(true);
+
+		if let Some(parsers) = config.git.commit_parsers.as_mut() {
+			for parser in parsers.iter_mut().filter(|p| p.footer.is_some()) {
+				parser.skip = Some(true)
+			}
+		}
+
```
jan-ferdinand, line 948: I'm not sure this is the most maintainable solution to the problem. Let me know if you have a better solution. :relieved:

orhun:
Thanks for picking up on this!

> Is the amount of documentation enough? It seems to be very little, but I couldn't readily identify more places where it makes sense to add additional documentation.

I think it is good enough! I will mention this feature in the [news](https://git-cliff.org/blog) so it should be visible enough.

>  Does it still make sense to go ahead with the suggested feature despite the possible confusion?

Yes, we can go ahead with it. Alternatively, I'm not sure if there are any actions needed but, you can report this to [git-conventional](https://github.com/crate-ci/git-conventional) (which `git-cliff` uses for parsing the commits) FWIW.

> I have left comments briefly discussing the out-of-syncedness of spec & parser in two places, one in the fixture and one in the spec. Does this make sense, or would you rather I remove or merge these? 

I would rather have them as comment like they are now. Good to have the references 👍🏼

git-cliff-core/src/changelog.rs
orhun: it's good enough!

orhun (commented):
LGTM, there is only one conflict that needs to be resolved then we are good to go!

jan-ferdinand:
> Thanks for picking up on this!

Took me long enough. :shrug: Thanks for your patience!

> I think it is good enough! I will mention this feature in the [news](https://git-cliff.org/blog) so it should be visible enough.

:ok_hand:

> Yes, we can go ahead with it. Alternatively, I'm not sure if there are any actions needed but, you can report this to [git-conventional](https://github.com/crate-ci/git-conventional) (which `git-cliff` uses for parsing the commits) FWIW.

I had [opened a PR](https://github.com/crate-ci/git-conventional/pull/54) in trying to resolve this issue. The maintainer chose to stay consistent with the reference implemenation rather than deviating from it, a decision I can understand.

> I would rather have them as comment like they are now. Good to have the references 👍🏼

:+1:

Thanks for your guidance. :blush: I'm going to rebase & force-push in hopse of removing the blocking conflict.

orhun (approved):

PR #574, opened 2024-03-24
fix(npm): fix authenticating for `yarn npm publish`

favna:
## Description

Follow up from #528 where I made a mistake as noted in https://github.com/orhun/git-cliff/pull/528#issuecomment-2016954138

## Motivation and Context

Fix a mistake

## How Has This Been Tested?

N.A.

N.A.

## Types of Changes

- [x] Other: Continuous delivery

## Checklist:

- [x] ~~I have updated the documentation accordingly.~~ N.A.
- [x] ~~I have formatted the code with [rustfmt](https://github.com/rust-lang/rustfmt).~~ N.A.
- [x] ~~I checked the lints with [clippy](https://github.com/rust-lang/rust-clippy).~~ N.A.
- [x] ~~I have added tests to cover my changes.~~ N.A.
- [x] ~~All new and existing tests passed.~~ N.A.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -304,6 +304,8 @@ jobs:
         shell: bash
         working-directory: npm/git-cliff
         run: |
+          yarn config set npmAuthToken ${NODE_AUTH_TOKEN}
+          yarn config set npmPublishRegistry "https://registry.npmjs.org"
           yarn install
           yarn build
           cp ../../README.md .


.github/workflows/cd.yml
```diff
@@ -304,6 +304,8 @@ jobs:
         shell: bash
         working-directory: npm/git-cliff
         run: |
+          yarn config set npmAuthToken ${NODE_AUTH_TOKEN}
```
orhun, line 307: I'm thinking if we should use this from the secrets just to be safe. 

```suggestion
          yarn config set npmAuthToken ${{ secrets.NPM_TOKEN }}
```

What do you think?

favna: Comes down to the same because that's where env gets it from too. This is how I always do it for my repos and I'm pretty sure it's also what you'll find in the GitHub workflow guides out there. I think it's whatever really. I can make the change if you really want but then I'll also drop the env binding a few lines below.

orhun: Ok sounds fair enough :)

orhun (approved):

PR #594, opened 2024-04-06
fix(npm): include the proper files in the npm tarball

### LINKED ISSUE #591: npm v2.2.0 is missing index.js for esm and cjs folders

### Description of the bug

The v2.2.0 version released last week is missing the `index.js` for the lib/esm and lib/cjs folders. This results in the programmatic API not working when importing.

https://www.npmjs.com/package/git-cliff?activeTab=code

### Steps To Reproduce

```ts
import { runGitCliff } from 'git-cliff';
```

### Expected behavior

The files should be included in the release

### Screenshots / Logs

_No response_

### Software information

- Operating system: N/A
- Rust version: N/A
- Project version: 2.2.0

### Additional context

_No response_

favna:
## Description

fixes #591, I think I broke this in #528 because `yarn npm publish` works differently with which files it includes in the tarball from `npm publish` (which just includes everything from the cwd)

I also added a `cp` of the `CHANGELOG.md` file, because Yarn likes to include that automatically, just like the README. They _used_ to also show it on the package pages before the redesign (ex: https://v3.yarnpkg.com/package/@sapphire/framework) so I can only assume it will be added back eventually.

## Motivation and Context

Emergency fix, as the current release on npm is broken.

## How Has This Been Tested?

I ran the `cd` commands locally:
```sh
yarn install
yarn build
cp ../../README.md .
cp ../../CHANGELOG.md .
yarn pack --dry-run
```
The only differing factor here being the last command because obviously I do not want to actually publish but this does show what gets put in the tarball:
```
yarn pack --dry-run    
➤ YN0000: CHANGELOG.md
➤ YN0000: README.md
➤ YN0000: lib/cjs/index.cjs
➤ YN0000: lib/cjs/index.cjs.map
➤ YN0000: lib/cjs/index.d.cts
➤ YN0000: lib/cli/cli.js
➤ YN0000: lib/esm/index.d.ts
➤ YN0000: lib/esm/index.js
➤ YN0000: lib/esm/index.js.map
➤ YN0000: package.json
➤ YN0000: Done in 0s 27ms
```

## Types of Changes

- [x] Other: releases

## Checklist:

- [X] All new and existing tests passed.

### DIFF
--- a/npm/git-cliff/package.json
+++ b/npm/git-cliff/package.json
@@ -22,6 +22,9 @@
       }
     }
   },
+  "files": [
+    "lib"
+  ],
   "scripts": {
     "typecheck": "tsc",
     "lint": "eslint .",
--- a/npm/git-cliff/yarn.lock
+++ b/npm/git-cliff/yarn.lock
@@ -1371,44 +1371,44 @@ __metadata:
   languageName: node
   linkType: hard
 
-"git-cliff-darwin-arm64@npm:2.1.0-rc.0":
-  version: 2.1.0-rc.0
-  resolution: "git-cliff-darwin-arm64@npm:2.1.0-rc.0"
+"git-cliff-darwin-arm64@npm:2.2.0":
+  version: 2.2.0
+  resolution: "git-cliff-darwin-arm64@npm:2.2.0"
   conditions: os=darwin & cpu=arm64
   languageName: node
   linkType: hard
 
-"git-cliff-darwin-x64@npm:2.1.0-rc.0":
-  version: 2.1.0-rc.0
-  resolution: "git-cliff-darwin-x64@npm:2.1.0-rc.0"
+"git-cliff-darwin-x64@npm:2.2.0":
+  version: 2.2.0
+  resolution: "git-cliff-darwin-x64@npm:2.2.0"
   conditions: os=darwin & cpu=x64
   languageName: node
   linkType: hard
 
-"git-cliff-linux-arm64@npm:2.1.0-rc.0":
-  version: 2.1.0-rc.0
-  resolution: "git-cliff-linux-arm64@npm:2.1.0-rc.0"
+"git-cliff-linux-arm64@npm:2.2.0":
+  version: 2.2.0
+  resolution: "git-cliff-linux-arm64@npm:2.2.0"
   conditions: os=linux & cpu=arm64
   languageName: node
   linkType: hard
 
-"git-cliff-linux-x64@npm:2.1.0-rc.0":
-  version: 2.1.0-rc.0
-  resolution: "git-cliff-linux-x64@npm:2.1.0-rc.0"
+"git-cliff-linux-x64@npm:2.2.0":
+  version: 2.2.0
+  resolution: "git-cliff-linux-x64@npm:2.2.0"
   conditions: os=linux & cpu=x64
   languageName: node
   linkType: hard
 
-"git-cliff-windows-arm64@npm:2.1.0-rc.0":
-  version: 2.1.0-rc.0
-  resolution: "git-cliff-windows-arm64@npm:2.1.0-rc.0"
+"git-cliff-windows-arm64@npm:2.2.0":
+  version: 2.2.0
+  resolution: "git-cliff-windows-arm64@npm:2.2.0"
   conditions: os=win32 & cpu=arm64
   languageName: node
   linkType: hard
 
-"git-cliff-windows-x64@npm:2.1.0-rc.0":
-  version: 2.1.0-rc.0
-  resolution: "git-cliff-windows-x64@npm:2.1.0-rc.0"
+"git-cliff-windows-x64@npm:2.2.0":
+  version: 2.2.0
+  resolution: "git-cliff-windows-x64@npm:2.2.0"
   conditions: os=win32 & cpu=x64
   languageName: node
   linkType: hard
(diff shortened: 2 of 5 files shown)


npm/git-cliff/package.json
favna, line 1: To clarify, what this essentially says is "include only the files in lib, and include everything in there". This ensures that the JavaScript files are also included. Yarn (and npm too for that matter) automatically include the `package.json` and `README.md` when found, so need to specify them.

npm/git-cliff/package.json
```diff
@@ -22,6 +22,9 @@
       }
     }
   },
+  "files": [
+    "lib"
```
aminya, line 26: Could you include the source Typescript files as well?

favna: There is absolutely no reason to do so. Types are provided through the output `.d.ts` files and node, bun and deno all will never even load the source typescript files.

favna: Also due to the export mapping (a few lines above this) the files in src won't even be accessible. Node will just say they're not exported. The only way to read then would with a manual readFile call.

aminya: Well, my build system Parcel is capable of using the source files directly. I also prefer untransformed source code that is on GitHub. Given what has been going on with the xz distribution, you should consider this.

favna: I don't see any relationship to what happened to xz. It's also not "going on", it's already resolved. Please show me the documentation where parcel says they use the source files. I've used parcel before and it doesn't. Please also explain what your exact use case is that you use parcel to not only consume the git-cliff npm package but also parcel bundle it. Parcel is for bundling webapps after all.

aminya: Here's the relevant Parcel documentation:
https://parceljs.org/features/dependency-resolution/#package-entries

I am not sure why this is controversial. 
The solution is simple, instead of only distributing transpiled code, the Typescript source can be included as well.

favna: Read the documentation carefully. The package.json is consulted.

- source -> this package doesn't define it, also this only affects symlinked packages which this wouldn't be
- exports -> this package defines this, directing to lib

The rest becomes irrelevant because an entrypoint has now been found, but to indulge the others:

- browser -> not defined
- module -> goes to lib
- main -> goes to lib

As for why it's controversial, it unnecessarily bloats the tarball causing the download size to be larger. 

As for JSR, this package does not (yet) get published to jsr so that's a non-issue and for later. Furthermore jsr is still in its infancy and I'd first like to see larger adoption. It's also a registry made by the Deno team whose Node compat layer is infamously horrible (especially when compared to Bun) so I have my opinionated reservations about whether it will even be relevant at all a year from now.

npm/git-cliff/yarn.lock
```diff
@@ -1422,12 +1422,12 @@ __metadata:
...
-    git-cliff-linux-arm64: "npm:2.1.0-rc.0"
-    git-cliff-linux-x64: "npm:2.1.0-rc.0"
-    git-cliff-windows-arm64: "npm:2.1.0-rc.0"
-    git-cliff-windows-x64: "npm:2.1.0-rc.0"
+    git-cliff-darwin-arm64: "npm:2.2.0"
+    git-cliff-darwin-x64: "npm:2.2.0"
+    git-cliff-linux-arm64: "npm:2.2.0"
+    git-cliff-linux-x64: "npm:2.2.0"
+    git-cliff-windows-arm64: "npm:2.2.0"
+    git-cliff-windows-x64: "npm:2.2.0"
```
orhun, line 1430: The versions in `package.json` are bumped automatically via `release.sh` script.

https://github.com/orhun/git-cliff/blob/98647800565f588f656234f5b717d3ce35db3344/release.sh#L17

I'm wondering if we should do the same for `yarn.lock`?

orhun (commented):
Hey guys! Thanks for your comments/work on this. I'm curious if there is any way to test the programmatic API via CI to ensure that we don't hit this issue again and everything works correctly.

npm/git-cliff/package.json
orhun: I'm not well versed in these topics so I cannot comment. But I would appreciate a ELI5 summary 😊

npm/git-cliff/yarn.lock
favna: That could be possible. This bump happened when I ran `yarn install`. Essentially to achieve the automatic bump that would have to be ran and then committed and pushed.

favna:
Hm we could maybe check the files returned by `ls`?

npm/git-cliff/package.json
favna: Sure! 

@aminya wants me to include the TypeScript files from `src` in the tarball. This is something that no library ever does. Search the favourite big-name packages written in TypeScript on https://yarnpkg.com as they have a file viewer for what's in the tarball, and you'll see that none of them include the source TypeScript files. This is because runtimes and bundlers will never access them. This is despite what @aminya says about Parcel, as described by my previous comment. Even if Parcel *would* access those files, it would be the only bundler to do so and it's far from the [most popular bundler (npmtrends)](https://npmtrends.com/esbuild-vs-parcel-vs-rollup-vs-snowpack-vs-vite-vs-webpack). Ultimately, including the source TypeScript files is unnecessary bloat and increasing filesize for no actual gain.

npm/git-cliff/yarn.lock
favna: Coming back to this, I'm going to have to retract my above statement. You will have to run `yarn install` in the `npm/git-cliff` folder yourself when making a commit such as [chore(release): prepare for v2.2.0
](https://github.com/orhun/git-cliff/commit/bd98caf95351d89a9f4f66e24b6d9217952621ab). That is the proper moment to update the lock file.

favna: Added in a21d221

npm/git-cliff/package.json
aminya: Optimizing the package size seems unnecessary, and the fact that the popular packages do not provide source files just reveals the wide attack surface in open source.
 
 The question remains though. How can I make sure the JavaScript files in the dist are exactly derived from the source TypeScript files in the repository? Do you have a byte-for-byte reproducible build and release process?

orhun: Thanks for the detailed summary and your thoughts! I agree that we should still have faith in the open source community. I don't have a strong preference between including/not including the TS files due to the fact that it's an whole another ecosystem that I'm not a part of. I can still very clearly see @aminya's concerns, but I'd say we can discuss this elsewhere for the sake of not blocking this PR.

I'm happy to include the TS files if there are strong arguments with practical and proven reproducibility along with a PR.

orhun (approved):
Thanks!

Issue #120, opened 2022-10-11
Allow running with `prepend` and `output`


bukowa:
**Is your feature request related to a problem? Please describe.**
In short I need 2 files, one with version `0.X.X.md` and one with `CHANGELOG.md`

**Describe the solution you'd like**
It should be possible to run with `--prepend` and `--output`.

**Describe alternatives you've considered**
Run the same command twice, once with `--prepend` one with `--output`.

orhun:
Hello, thanks for submitting this issue! 🐻 This feature is implemented in 7325be84045ad376e0989a111ed3c44a3e1400ea and will be available with the next release.

bukowa:
Thank you for implementing!

PR #653, opened 2024-05-24
fix(args): allow -o and -p together if they point to different files

antonengelhardt:
setting prepend and output together will ignore the prepend flag

## Description

- preprend and output flags are conflicting and should only be used together if they point to different paths

## Motivation and Context

https://github.com/orhun/git-cliff-action/issues/21#issuecomment-2130022787

## How Has This Been Tested?

I build it locally and tested it with one of my rust projects

Screenshot-Visual Studio Code-007271

## Types of Changes

- [ ] Other 

not sure

## Checklist:

### DIFF
--- a/git-cliff/src/lib.rs
+++ b/git-cliff/src/lib.rs
@@ -386,6 +386,11 @@ pub fn run(mut args: Opt) -> Result<()> {
 			)));
 		}
 	}
+	if args.output.is_some() && args.prepend.is_some() {
+		return Err(Error::ArgumentError(String::from(
+			"'-o' and '-p' cannot be used together",
+		)));
+	}
 	if args.body.is_some() {
 		config.changelog.body.clone_from(&args.body);
 	}
--- a/git-cliff/src/lib.rs
+++ b/git-cliff/src/lib.rs
@@ -386,9 +386,13 @@ pub fn run(mut args: Opt) -> Result<()> {
 			)));
 		}
 	}
-	if args.output.is_some() && args.prepend.is_some() {
+	if args.output.is_some() &&
+		args.prepend.is_some() &&
+		args.output.as_ref() == args.prepend.as_ref()
+	{
 		return Err(Error::ArgumentError(String::from(
-			"'-o' and '-p' cannot be used together",
+			"'-o' and '-p' can only be used together if they point to different \
+			 files",
 		)));
 	}
 	if args.body.is_some() {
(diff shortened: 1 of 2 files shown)


website/docs/usage/args.md
```diff
@@ -36,7 +36,7 @@ git-cliff [FLAGS] [OPTIONS] [--] [RANGE]
     --tag-pattern <PATTERN>      Sets the regex for matching git tags [env: GIT_CLIFF_TAG_PATTERN=]
     --with-commit <MSG>...       Sets custom commit messages to include in the changelog [env: GIT_CLIFF_WITH_COMMIT=]
     --skip-commit <SHA1>...      Sets commits that will be skipped in the changelog [env: GIT_CLIFF_SKIP_COMMIT=]
--p, --prepend <PATH>             Prepends entries to the given changelog file [env: GIT_CLIFF_PREPEND=]
+-p, --prepend <PATH>             Prepends entries to the given changelog file, incompatible with -o [env: GIT_CLIFF_PREPEND=]
```
orhun, line 39: I don't think it is required to mention it here. 

```suggestion
-p, --prepend <PATH>             Prepends entries to the given changelog file [env: GIT_CLIFF_PREPEND=]
```

orhun: We can probably add a comment to the `prepend` example in `examples.md` instead.

antonengelhardt: @orhun I added a line to the examples and removed it from the website docs :)

antonengelhardt:
@orhun i ticked that this is a breaking change. Not sure though. WDYT?

orhun:
Ah, I found an issue: <https://github.com/orhun/git-cliff/issues/120>

It turns out someone actually requested this feature in the past, so I'm a bit unsure about merging this :/

I think we can bring back the note in the `git-cliff-action` repo instead. What do you think?

antonengelhardt:
Yeah, I get your point.

In my opinion, output and prepend should work together, but only if the file paths are not equal. 

If the file paths are the same, it does not make sense to use both flags and an error should be shown.

Do you agree?

orhun:
Actually yes, that sounds more reasonable!

antonengelhardt:
Any idea for the error message?

orhun:
Something along the lines of "Output files cannot be the same" maybe

orhun (approved):

orhun:
Awesome, thanks!

Issue #659, opened 2024-05-26
Do not fail on unexisting GitHub/GitLab template variables


orhun:
### Is there an existing issue or pull request for this?

- [X] I have searched the existing issues and pull requests

### Feature description

When the `github` or `gitlab` feature isn't enabled and any remote-related template variable such as `github.contributors` is being used, `git-cliff` fails with the following error:

```sh
$ cargo run --no-default-features -- -c examples/github.toml

 ERROR git_cliff > Template render error:
Variable `github.contributors` not found in context while rendering 'template'
```

### Desired solution

We should warn the user with a message like:

> WARN: You seem to be using a GitHub-related variable but GitHub feature is not enabled.

### Alternatives considered

None.

### Additional context

I think we can add the `RemoteReleaseMetadata` to the `Release` struct even though the GitHub feature isn't enabled and warn the user instead (via checking the used variables via `contains_variable` method).

PR #672, opened 2024-06-03
chore(example): allow using github template without github variables

R11baka:
## Description

Fixed default template related in bug https://github.com/orhun/git-cliff/issues/659
## Types of Changes

## Checklist:

How to check ? run in shell
`cargo run --no-default-features -- -c examples/github.toml`

### DIFF
--- a/examples/github.toml
+++ b/examples/github.toml
@@ -25,6 +25,7 @@ body = """
       [#{{ commit.github.pr_number }}]({{ self::remote_url() }}/pull/{{ commit.github.pr_number }}) \
     {%- endif %}
 {%- endfor -%}
+
 {%- if github -%}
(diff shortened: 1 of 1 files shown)


orhun:
I will accept this as a short term solution, but it does not really fix the actual underlying issue.

https://github.com/orhun/git-cliff/blob/2c2a300616fa151b91858a2a7d88bdc9b9dae497/git-cliff-core/src/release.rs#L33-L41

We should instead have these structs available at all times (even though the feature isn't enabled) so that they are usable in the template although they are empty.

orhun (approved):

R11baka:
@orhun thanks for advice. Will try to implement your  solution

Issue #697, opened 2024-06-14
Make changelog.header into a template


Cyclonit:
### Is there an existing issue or pull request for this?

- [X] I have searched the existing issues and pull requests

### Feature description

Currently only `changelog.body` and `changelog.footer` are processed as Tera templates. Making `changelog.header` a template improves parity between the three parts of the changelog.

### Desired solution

- The processing of `changelog.header` will be aligned with that of `changelog.body` and `changelog.footer`.
- The template in `changelog.header` gets the same context as that in `changelog.footer`.

### Alternatives considered

none

orhun:
+1 to this, will check the PR soon.

Cyclonit:
The feature was implemented in #698.

PR #698, opened 2024-06-14
feat(changelog): make `changelog.header` a template

Cyclonit:
Made changelog.header into a template like changelog.body and changelog.footer. Added test test-header-template.

## Description

Implement #697.

## Motivation and Context

Create parity between `changelog.header`, `changelog.body` and `changelog.footer`.

## How Has This Been Tested?

- Added new test fixture "test-header-template".

## Types of Changes

- [X] New feature (non-breaking change which adds functionality)
- [X] Documentation (no code change)
- [ ] Other 

## Checklist:

- [X] My code follows the code style of this project.
- [X] I have formatted the code with [rustfmt](https://github.com/rust-lang/rustfmt).

### DIFF
--- a/git-cliff-core/src/changelog.rs
+++ b/git-cliff-core/src/changelog.rs
@@ -29,6 +29,7 @@ use std::time::{
 pub struct Changelog<'a> {
 	/// Releases that the changelog will contain.
 	pub releases:       Vec<Release<'a>>,
+	header_template:    Option<Template>,
 	body_template:      Template,
 	footer_template:    Option<Template>,
 	config:             &'a Config,
@@ -41,6 +42,10 @@ impl<'a> Changelog<'a> {
 		let trim = config.changelog.trim.unwrap_or(true);
 		let mut changelog = Self {
 			releases,
+			header_template: match &config.changelog.header {
+				Some(header) => Some(Template::new(header.to_string(), trim)?),
+				None => None,
+			},
 			body_template: Template::new(
 				config
 					.changelog
@@ -420,16 +425,14 @@ impl<'a> Changelog<'a> {
 			serde_json::to_value(self.config.remote.clone())?,
 		);
 		#[cfg(feature = "github")]
-		let (github_commits, github_pull_requests) = if self.config.remote.github.is_set()
-		{
+		let (github_commits, github_pull_requests) = if self.config.remote.github.is_set() {
 			self.get_github_metadata()
 				.expect("Could not get github metadata")
 		} else {
 			(vec![], vec![])
 		};
 		#[cfg(feature = "gitlab")]
-		let (gitlab_commits, gitlab_merge_request) = if self.config.remote.gitlab.is_set()
-		{
+		let (gitlab_commits, gitlab_merge_request) = if self.config.remote.gitlab.is_set() {
 			self.get_gitlab_metadata()
 				.expect("Could not get gitlab metadata")
 		} else {
@@ -502,15 +505,30 @@ impl<'a> Changelog<'a> {
 			.postprocessors
 			.clone()
 			.unwrap_or_default();
-		if let Some(header) = &self.config.changelog.header {
-			let write_result = write!(out, "{header}");
+
+		let mut releases = self.releases.clone();
+
+		// render the header template
+		if let Some(header_template) = &self.header_template {
+			let write_result = writeln!(
+				out,
+				"{}",
+				header_template.render(
+					&Releases {
+						releases: &releases,
+					},
+					Some(&self.additional_context),
+					&postprocessors,
+				)?
+			);
 			if let Err(e) = write_result {
 				if e.kind() != std::io::ErrorKind::BrokenPipe {
 					return Err(e.into());
 				}
 			}
 		}
-		let mut releases = self.releases.clone();
+
+		// render the body template
 		for release in releases.iter_mut() {
 			let write_result = write!(
 				out,
@@ -527,6 +545,8 @@ impl<'a> Changelog<'a> {
 				}
 			}
 		}
+
+		// render the footer template
 		if let Some(footer_template) = &self.footer_template {
 			let write_result = writeln!(
 				out,
@@ -545,6 +565,7 @@ impl<'a> Changelog<'a> {
 				}
 			}
 		}
+
 		Ok(())
 	}
 
@@ -927,6 +948,7 @@ mod test {
 		assert_eq!(
 			String::from(
 				r#"# Changelog
+
 			## Release [v1.1.0] - 1970-01-01
 
 
@@ -1032,6 +1054,7 @@ chore(deps): fix broken deps
 		assert_eq!(
 			String::from(
 				r#"# Changelog
+
 			## Unreleased
 
 			### Bug Fixes
@@ -1135,6 +1158,7 @@ chore(deps): fix broken deps
 		changelog.generate(&mut out)?;
 		expect_test::expect![[r#"
     # Changelog
+
     ## Unreleased
 
     ### Bug Fixes
--- a/git-cliff-core/src/changelog.rs
+++ b/git-cliff-core/src/changelog.rs
@@ -425,14 +425,16 @@ impl<'a> Changelog<'a> {
 			serde_json::to_value(self.config.remote.clone())?,
 		);
 		#[cfg(feature = "github")]
-		let (github_commits, github_pull_requests) = if self.config.remote.github.is_set() {
+		let (github_commits, github_pull_requests) = if self.config.remote.github.is_set()
+		{
 			self.get_github_metadata()
 				.expect("Could not get github metadata")
 		} else {
 			(vec![], vec![])
 		};
 		#[cfg(feature = "gitlab")]
-		let (gitlab_commits, gitlab_merge_request) = if self.config.remote.gitlab.is_set() {
+		let (gitlab_commits, gitlab_merge_request) = if self.config.remote.gitlab.is_set()
+		{
 			self.get_gitlab_metadata()
 				.expect("Could not get gitlab metadata")
 		} else {
@@ -506,16 +508,13 @@ impl<'a> Changelog<'a> {
 			.clone()
 			.unwrap_or_default();
 
-		let mut releases = self.releases.clone();
-
-		// render the header template
 		if let Some(header_template) = &self.header_template {
 			let write_result = writeln!(
 				out,
 				"{}",
 				header_template.render(
 					&Releases {
-						releases: &releases,
+						releases: &self.releases,
 					},
 					Some(&self.additional_context),
 					&postprocessors,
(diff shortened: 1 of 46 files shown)


Cyclonit:
Hi @orhun,
functionally this is a very small change. I was able to copy the logic from the body and footer template processing 1-to-1. However, I do fear that we will have to flag this as a breaking change regardless. All of the tests and examples contain explicit `\n` in the header. These are now obsolete. Without removing them, we would end up with an extra empty line between the header and the first release.

Cyclonit:
I am  not quite sure what to do about the allegedly wrong formatting in `changelog.rs`. When I run `cargo fmt` locally, it formats the code the way it is in this PR. I don't know why the pipeline demands it to be formatted differently. See https://github.com/orhun/git-cliff/actions/runs/9518131251/job/26238303168?pr=698

Similarly I am having trouble figuring out why the test suite fails: https://github.com/orhun/git-cliff/actions/runs/9518131251/job/26238304948?pr=698
Actual and expected look identical to me.

Cyclonit:
And this is another problem: https://github.com/orhun/git-cliff/actions/runs/9518517603/job/26239610418?pr=698#step:3:842

I have no idea why the template behaves differently in test-header-template compared to test-footer-template. They are identical, but in test-header-template there is indentation, in test-footer-template there is not. I am done for today, maybe someone else could take a look into this.

website/docs/templating/context.md
orhun, line 1: Maybe it makes sense to make these documentation changes in another PR? Not sure if everything here is related to the scope of this PR.

git-cliff-core/src/changelog.rs
```diff
@@ -434,13 +433,44 @@ impl<'a> Changelog<'a> {
...
+
+			// always insert an empty line after the header
+			if let Err(e) = out.write(b"\n") {
+				if e.kind() != std::io::ErrorKind::BrokenPipe {
+					return Err(e.into());
+				}
+			}
+		}
+
+		// render the body_template for each release
```
orhun, line 465: unnecessary comment

git-cliff-core/src/changelog.rs
```diff
@@ -434,13 +433,44 @@ impl<'a> Changelog<'a> {
...
+					&postprocessors,
+				)?
+			);
+			if let Err(e) = write_result {
+				if e.kind() != std::io::ErrorKind::BrokenPipe {
+					return Err(e.into());
+				}
+			}
+
+			// always insert an empty line after the header
```
orhun, line 457: unnecessary comment

git-cliff-core/src/changelog.rs
```diff
@@ -409,14 +414,8 @@ impl<'a> Changelog<'a> {
...
-		if let Some(header) = &self.config.changelog.header {
-			let write_result = write!(out, "{header}");
-			if let Err(e) = write_result {
-				if e.kind() != std::io::ErrorKind::BrokenPipe {
-					return Err(e.into());
-				}
-			}
-		}
+
+		// fetch release metadata from remote repositories
```
orhun, line 418: unnecessary comment

git-cliff-core/src/changelog.rs
```diff
@@ -434,13 +433,44 @@ impl<'a> Changelog<'a> {
 				bitbucket_commits.clone(),
 				bitbucket_pull_request.clone(),
 			)?;
+		}
+
+		// render the header_template
```
orhun, line 438: unnecessary comment

git-cliff-core/src/changelog.rs
```diff
@@ -434,13 +433,44 @@ impl<'a> Changelog<'a> {
...
+					return Err(e.into());
+				}
+			}
+
+			// always insert an empty line after the header
+			if let Err(e) = out.write(b"\n") {
+				if e.kind() != std::io::ErrorKind::BrokenPipe {
+					return Err(e.into());
+				}
+			}
```
orhun, line 462: Do we want to do this? I think it can be left to user to insert a new line.

git-cliff-core/src/changelog.rs
```diff
@@ -449,6 +479,8 @@ impl<'a> Changelog<'a> {
 				}
 			}
 		}
+
+		// render the footer_template
```
orhun, line 483: unnecessary comment

git-cliff-core/src/changelog.rs
```diff
@@ -1044,6 +1079,7 @@ chore(deps): fix broken deps
 		changelog.generate(&mut out)?;
 		expect_test::expect![[r#"
     # Changelog
+
```
orhun, line 1082: What are the reason for these newline changes? Can we preserve the default behavior?

orhun:
Can you resolve the conflicts? 😊

git-cliff-core/src/changelog.rs
Cyclonit: Without this there can never be an empty line after header unless  `changelog.trim` is `false`, which would also impact the body and footer. Given that the body should always start with a heading, I think the forced new line after the header is a reasonable choice. But I am not entirely happy with it.

git-cliff-core/src/changelog.rs
Cyclonit: This is caused by the difficult handling of newlines at the end of the header. I'll look into it.

website/docs/templating/context.md
Cyclonit: Sure. I noticed that the current context documentation doesn't line up with the reality of the context in body and footer. Thus I decided to update it straight away, because I didn't want to add half-baked documentation for the header template.

git-cliff-core/src/changelog.rs
Cyclonit: I disagree. `changelog.rs` is a huge file. The method generate especially is a long and very repetitive method, making it very difficult tor read unless you happen to have spent a lot of time in there. These simple comments help those who aren't familiar with the codebase to orient themselves.

Do you have a general policy as to what you would consider a necessary comment?

git-cliff-core/src/changelog.rs
Cyclonit: I could not find a straight forward way of preserving the new-line handling. It boils down to whether we use `write` or `writeln` when rendering the header template in [changelog.rs#513](https://github.com/Cyclonit/git-cliff/blob/feature/make_changelog_header_a_template/git-cliff-core/src/changelog.rs#L513).

- When using `write` (no additional `\n` after the header), the tests in `changelog.rs` are successful. They do not expect a new line between the header and body.

- When using `writeln` (forced additional `\n` after the header`), the test fixtures are successful. All fixtures that contain a header expect a new line between the header and body.

But in either case, the respective other tests fail. Personally I prefer the option of using `writeln`. In almost all use cases, the body begins with a header. Having a new line in front of it results in cleaner markdown. This would require changing the tests in `changelog.rs` but leaves the fixtures unchanged. It is also unlikely to have an impact on current users.

git-cliff-core/src/changelog.rs
Cyclonit: After some more testing I concluded that there is no general solution that preserves new-line handling in all tests. We either enforce a new-line after the header or not and either the rust tests in `changelog.rs` or the test fixtures break. I would opt for the former because it has less impact on users. Thus I removed this way of forcing a new-line.

website/docs/templating/context.md
Cyclonit: I removed the documentation overhaul and just changed the wording for `changelog.header` in the documentation.

git-cliff-core/src/changelog.rs
orhun: There isn't a general policy but I usually try to avoid inline comments unless the code is in a unreadable / hard-to-understand state.

> These simple comments help those who aren't familiar with the codebase to orient themselves.

Yeah, you are right. But I think we should split the long functions into multiple parts instead and document each one. We can also add unit tests this way.

git-cliff-core/src/changelog.rs
orhun: Yeah, I'm on the same page. We can simply go with `println!` and update the tests in `changelog.rs`. Like you said, it is not likely to have a big impact.

orhun (commented):
LGTM! Can you update the tests and documentation?

Cyclonit:
I've fixed the tests and added the minimalist documentation update. How do we deal with the formatting? I cannot get my local rustfmt to format the code the way the pipeline wants it to be.

website/docs/configuration/changelog.md
```diff
@@ -24,7 +24,7 @@ postprocessors = [{ pattern = "foo", replace = "bar"}]
 
 ### header
 
-Header text that will be added to the beginning of the changelog.
+Footer template that will be rendered and added to the beginning of the changelog.
```
orhun, line 27:
```suggestion
Header template that will be rendered and added to the beginning of the changelog.
```

orhun:
> I've fixed the tests and added the minimalist documentation update. How do we deal with the formatting? I cannot get my local rustfmt to format the code the way the pipeline wants it to be.

Thanks! The formatting is already fixed in `main`, so if you merge those changes it will be fine I think.

I can also handle it before merging the PR, no biggie.

website/docs/configuration/changelog.md
orhun: Also, do you think this documentation change will be enough? I'm thinking if we should add another sentence saying that this works similar to body and footer as well. Or maybe a little example?

Cyclonit:
> > I've fixed the tests and added the minimalist documentation update. How do we deal with the formatting? I cannot get my local rustfmt to format the code the way the pipeline wants it to be.
> 
> Thanks! The formatting is already fixed in `main`, so if you merge those changes it will be fine I think.
> 
> I can also handle it before merging the PR, no biggie.

I've tried improving the documentation by specifying the context of each template. Is this okay?

orhun (approved):
LGTM, thank you!

One thing that crossed my mind is that `--prepend` won't work as expected now when there are template variables in the `header`. It used to work like:

1- Read the contents from existing changelog file.
2- Replace the `header` in the file with the new changelog entries.

Now that the `header` is not a raw string & is not rendered, we won't be able to perform the replace properly.

Are you interested in fixing this as well? I think we only need to render `header` before replacing it in the file. I would really appreciate if you can test this case and create an issue!

Cyclonit:
> LGTM, thank you!
> 
> One thing that crossed my mind is that `--prepend` won't work as expected now when there are template variables in the `header`. It used to work like:
> 
> 1- Read the contents from existing changelog file. 2- Replace the `header` in the file with the new changelog entries.
> 
> Now that the `header` is not a raw string & is not rendered, we won't be able to perform the replace properly.
> 
> Are you interested in fixing this as well? I think we only need to render `header` before replacing it in the file. I would really appreciate if you can test this case and create an issue!

I was not aware of this feature. I created a bug issue to deal with that: #712

PR #709, opened 2024-06-15
refactor(lib): clean up some code

DaniPopes:
## Description

Some clean up opportunities I found while reading the code.

No functional changes intended.

Please let me know if breaking changes are a concern for you, I can revert the change to the public API in the core crate.

## Motivation and Context

Refactor

## How Has This Been Tested?

`cargo test` + `cargo clippy`

## Types of Changes

- [ ] Other 

## Checklist:


git-cliff/src/lib.rs
```diff
@@ -532,6 +524,16 @@ pub fn run(mut args: Opt) -> Result<()> {
...
+	let mut out: Box<dyn io::Write> = if let Some(path) = &args.output {
+		if path == Path::new("-") {
+			Box::new(io::stdout())
+		} else {
+			Box::new(io::BufWriter::new(File::create(path)?))
+		}
+	} else {
+		Box::new(io::stdout())
+	};
+	let out = &mut *out;
```
orhun, line 536: Something is broken after this change:

```sh
# outputs to stdout instead of prepending to the file
git cliff -l -p CHANGELOG.md
```

git-cliff/src/lib.rs
```diff
@@ -544,40 +546,22 @@ pub fn run(mut args: Opt) -> Result<()> {
...
-			changelog.write_context(&mut io::stdout())
-		};
+		changelog.write_context(out)?;
+		return Ok(());
 	}
-	if let Some(ref path) = args.prepend {
-		changelog.prepend(fs::read_to_string(path)?, &mut File::create(path)?)?;
+
+	if let Some(path) = &args.prepend {
+		changelog.prepend(fs::read_to_string(path)?, out)?;
```
orhun, line 560: I think here we shouldn't change the output:

```suggestion
		changelog.prepend(fs::read_to_string(path)?, &mut File::create(path)?)?;
```

git-cliff/src/lib.rs
```diff
@@ -544,40 +546,22 @@ pub fn run(mut args: Opt) -> Result<()> {
...
-		};
-		if args.context {
-			changelog.write_context(&mut output)
-		} else {
-			changelog.generate(&mut output)
-		}
-	} else if args.prepend.is_none() {
-		changelog.generate(&mut io::stdout())
-	} else {
-		Ok(())
```
orhun, line 581: I feel like this part is extremely fragile, I hope we are not breaking people's workflows by this change 🤞🏼

git-cliff/src/lib.rs
orhun, line 536: Can we not do this and use `&mut out` where necessary?

orhun (commented):
Thanks for this! It mostly looks good, the behavior looks like the same. I have some concerns about `lib.rs` but will double check to make sure.

DaniPopes:
@orhun should be good now :pray:

git-cliff/src/lib.rs
```diff
@@ -559,40 +557,23 @@ pub fn run(mut args: Opt) -> Result<()> {
...
-		};
+		changelog.write_context(&mut out)?;
+		return Ok(());
 	}
-	if let Some(ref path) = args.prepend {
-		changelog.prepend(fs::read_to_string(path)?, &mut File::create(path)?)?;
+
+	if let Some(path) = &args.prepend {
+		let mut out = io::BufWriter::new(File::create(path)?);
+		changelog.prepend(fs::read_to_string(path)?, &mut out)?;
```
orhun, line 572: Okay something is still wrong here. `cargo run -- -u -p CHANGELOG.md` overwrites the entire changelog content with the unreleased section instead of prepending to the top.

Maybe revert the changes after `// Print the result.` and compare the result to see if it is caused by this change or not. Like I said, this part is extremely fragile (not tested very well) so it might be better to not touch it 😅  I'm happy to merge this PR with the other changes. (and I don't want to block #711 any further)

DaniPopes: Sorry about that, fixed in b6d2a219c64cad45f3d659826c961fa16b957b0e. I am pretty confident the code is equivalent now, comparing side by side with <https://github.com/orhun/git-cliff/blob/e7807e13c4b38aaa4a735ff05b69fdd6b57a7a85/git-cliff/src/lib.rs#L573-L599>
- the `output` variable is extracted
- the second `if args.context` is removed as it's unreachable
- after these changes the last if block becomes `if args.output.is_some() { generate } else if args.prepend.is_none() { generate }`, which is collapsed into one

The only behavior change I can see here is there is an extra new line in `bumped_version` with `output` specified (now using `writeln!` for both cases instead of `write_all`), I can revert this if you'd like https://github.com/orhun/git-cliff/blob/e7807e13c4b38aaa4a735ff05b69fdd6b57a7a85/git-cliff/src/lib.rs#L563-L571

orhun: Looks good & works as expected now, thanks!

orhun:
Awesome work, thanks a lot!

Issue #687, opened 2024-06-10
Gitlab integration with self hosted instance does not work


nqdrizzt:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description of the bug

While using a self hosted gitlab instance, the gitlab context of each commit is always empty like in the example bekow:

{
	"id": "49bf75934a826c6d182e958405d680938bb474c3",
	"message": "change all input variables to lower case",
	"body": "Merge branch 'fix_lower_case_inputs' into 'main'",
	"footers": [
		{
			"token": "Closes",
			"separator": " #",
			"value": "2",
			"breaking": false
		},
		{
			"token": "BREAKING CHANGE",
			"separator": ":",
			"value": "change all input variables to lower case\n\nSee merge request me-and-only-me/git-cliff-changelog!2",
			"breaking": true
		}
	],
	"group": "🐛 Bug Fixes",
	"breaking_description": "change all input variables to lower case\n\nSee merge request me-and-only-me/git-cliff-changelog!2",
	"breaking": true,
	"scope": "inputs",
	"links": [
		{
			"text": "#2",
			"href": "/-/issues/2"
		},
		{
			"text": "!2",
			"href": "/-/merge_requests/2"
		}
	],
	"author": {
		"name": "John Snow",
		"email": "john.snow@got.com",
		"timestamp": 1710149612
	},
	"committer": {
		"name": "John Snow",
		"email": "john.snow@got.com",
		"timestamp": 1710149612
	},
	"conventional": true,
	"merge_commit": true,
	"github": {
		"username": null,
		"pr_title": null,
		"pr_number": null,
		"pr_labels": [],
		"is_first_time": false
	},
	"gitlab": {
		"username": null,
		"pr_title": null,
		"pr_number": null,
		"pr_labels": [],
		"is_first_time": false
	},
...
...
},

git-cliff verbose output: 
TRACE git_cliff_core::repo > Upstream URL: https://my.gitlabinstance.com/me-and-only-me/git-cliff-changelog.git
 DEBUG git_cliff            > No GitHub remote is set, using remote: me-and-only-me/git-cliff-changelog
 TRACE git_cliff            > Arguments: Opt {
    help: None,
    version: None,
    verbose: 4,
    init: None,
    config: "/config/git-cliff.toml",
    workdir: None,
    repository: None,
    include_path: None,
    exclude_path: None,
    tag_pattern: None,
    with_commit: None,
    skip_commit: None,
    prepend: None,
[...]

PR #742, opened 2024-07-03
fix(gitlab): URL-encode the owner in remote requests for GitLab

tonybutt:
## Description

Gitlab requires that the project path be fully encoded. See the docs for details. https://docs.gitlab.com/ee/api/rest/#namespaced-path-encoding Pulls in the url encoding crate here. Might not be needed when you have the choice to just update the documentation. I don't think the end user should be required to url encode their own strings. It might be nicer to just add the subgroup paradigm to the gitlab config in general for the future. This will fix the problem for now.

## Motivation and Context
I want to use git cliff on my self hosted gitlab.
https://github.com/orhun/git-cliff/issues/687

## How Has This Been Tested?
I wrote a test. Doesn't cover the case on IF the end user url encodes their own string. This will break unexpectedly, but I rushed this in.

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -812,6 +812,7 @@ dependencies = [
  "tokio",
  "toml",
  "url",
+ "urlencoding",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


tonybutt:
I use nix so getting onto nightly rust fmt/clippy was non trivial for me tonight. 

Feel free to do the clean up. There were lots of failing tests so I only ensured mine passed locally. I imagine I am missing some setup for those tests specifically, but maybe I broke them however my change seems unrelated. More importantly running git cliff with my changes produced the desired outcome in the issue.

tonybutt:
There are some additional problems related here. Encoding subgroups is one and I'd like to keep this PR pure to that particular issue. I am going to open another PR for the deserialize problem.

orhun (approved):
Great, thanks!

Issue #757, opened 2024-07-12
Allow the whole commit object to be used as a field inside of a commit parser


DerTiedemann:
### Is there an existing issue or pull request for this?

- [X] I have searched the existing issues and pull requests

### Feature description

We are grouping our commit by labels on PRs and there is currently no easy way to group them and rename the groups based on regex.

### Desired solution

I would love to just use any field in the commit object to group objects inside of the commit parsers, most prominently `github.pr_labels`. To be fair the easiest solution would to just add this as an option, but I think that would just kick the can along until someone decides they want a different label.

### Alternatives considered

There is a workaround inside of the template definition itself, described here: https://github.com/orhun/git-cliff/issues/461#issuecomment-1913595602

### Additional context

Original Implementation PR: https://github.com/orhun/git-cliff/pull/312

DerTiedemann:
One idea to realisze this is to use the [TeraContext](https://github.com/Keats/tera/blob/master/src/context.rs)which allows accessing values dynamically after the object has been serialized. As seen [here](https://github.com/Keats/tera/blob/ae13d7ce39d732aae3f68435ed52c60732fe0ee0/src/context.rs#L116C1-L119C6). Though this would add a serialization cost for each commit, not sure if that would be an issue or not for such a niche feature. Also problematic would be the fact that the TeraContext does return a `serde_json::Value` which would be turned into a string and therefore would lose some accuracy regarding arrays for example. The workaround given above has to deal with that for example.

PR #758, opened 2024-07-12
Allow whole commit context to be used inside of CommitParser config

DerTiedemann:
- fix: reorder execution to have context poplulated before parsing/pruning
- feat: allow field to read from context object

Description WIP see https://github.com/orhun/git-cliff/issues/757


git-cliff-core/src/changelog.rs
```diff
@@ -62,9 +62,9 @@ impl<'a> Changelog<'a> {
 			config,
 			additional_context: HashMap::new(),
 		};
+		changelog.add_remote_data()?;
```
orhun, line 65: This was added in #703 - I'm not sure if changing the order would affect things but I'm guessing it should be fine.

git-cliff-core/src/commit.rs
```diff
@@ -287,25 +294,17 @@ impl Commit<'_> {
...
-					match field_name.as_str() {
-						"id" => Some(self.id.clone()),
-						"message" => Some(self.message.clone()),
-						"body" => body,
-						"author.name" => self.author.name.clone(),
-						"author.email" => self.author.email.clone(),
-						"committer.name" => self.committer.name.clone(),
-						"committer.email" => self.committer.email.clone(),
-						_ => None,
+				match tera::dotted_pointer(&lookup_context, field_name) {
```
orhun, line 297: Hmm, this function is not documented anywhere. How did you find it? Do we know if it is stable?

git-cliff-core/src/commit.rs
```diff
@@ -728,4 +727,55 @@ mod test {
 
 		Ok(())
 	}
+
+	#[test]
+	fn field_name_regex() -> Result<()> {
```
orhun, line 732: Thanks for adding the unit test! Can you also [add a fixture](https://github.com/orhun/git-cliff/tree/main/.github/fixtures) for testing the behavior?

orhun (commented):
This is definitely good to have!

Can you also update the [documentation](https://git-cliff.org/docs/configuration/git#commit_parsers) (i.e. add some examples)? Adding your use case to [Tips and Tricks](https://git-cliff.org/docs/tips-and-tricks) would be also pretty nice.

git-cliff-core/src/commit.rs
DerTiedemann: Found the functionality in the tera docs [here](https://keats.github.io/tera/docs/#dot-notation) which allows stuff like this to work: `{% if commit.github.username %} by @{{ commit.github.username }}{%- endif -%}`. Then just looked where user context is provided and picked around a little. Found in code [here](https://github.com/Keats/tera/blob/ae13d7ce39d732aae3f68435ed52c60732fe0ee0/src/context.rs#L244C1-L269C10).

Its true that this API is not documented, but its used by tera itself to make user contexts possible/easy to use. So IMHO if this is changed something similar has to be in tera itself to replace it. If this is of concern I can yeet the relevant code into a module and call it a day. The functionality is tested and benched.

git-cliff-core/src/changelog.rs
DerTiedemann: I looked at [add_remote_data](https://github.com/orhun/git-cliff/blob/32da6d390394c0907317b0da0b8f4d2c72ec2fa7/git-cliff-core/src/changelog.rs#L420-L481) and found no relevant field read access that would have been populated by `process_commits` or `process_releases`.

git-cliff-core/src/commit.rs
orhun: Got it, it makes sense to use it then. Good discovery 👍🏼

orhun (approved):
Perfect, thanks for the PR 🔥

PR #768, opened 2024-07-25
feat(profiler): support performance profiling via pprof

aminya:
## Description

This adds support for performance profiling, allowing for finding bottlenecks and performance optimizations. 

It can be enabled via `profiler` feature and the `bench` build profile.

```shell
cargo build --profile=bench --features=profiler
```
## Motivation and Context

This allows for debugging performance issues such as #767 

## How Has This Been Tested?

I tested on the repo mentioned in #767 

## Screenshots / Logs (if applicable)

This screenshot shows that Git Cliff doesn't have any caching, and it goes over the whole history every time no matter what in the subsequent calls.

git-cliff 3758528088515003613 flamegraph

## Types of Changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation (no code change)
- [ ] Refactor (refactoring production code)
- [ ] Other 

## Checklist:

- [x] My code follows the code style of this project.
- [ ] I have updated the documentation accordingly.
- [x] I have formatted the code with [rustfmt](https://github.com/rust-lang/rustfmt).
- [x] I checked the lints with [clippy](https://github.com/rust-lang/rust-clippy).
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #772, opened 2024-07-27
perf(changelog): cache commit retain checks (258 times faster generation)

### LINKED ISSUE #767: Slow for large repositories with binary files in Git LFS and long history

### Description of the bug

Git-cliff is quite slow when it is used for large repositories with binary files in Git LFS when the history is long. Is there a way to cache the results so that it is faster in the subsequent runs?

### Steps To Reproduce

Run git-cliff in a large repository with LFS files and with a deep history. In my case, it takes 80 seconds to generate the changelog.

I use a code like this to generate changelogs for each package in the mono-repository

```js
import glob from "fast-glob"
import { runGitCliff } from 'git-cliff';

async function main() {
  const packages = await glob("packages/*", { onlyDirectories: true })
  await Promise.all(packages.map(generatePackageChangelog))
}
await main()

function generatePackageChangelog(name: string) {
  return runGitCliff(
    [
      "--repository",
      "../../",
      "--config",
      "../../cliff.toml",
      "--exclude-path",
      `packages/${name}/CHANGELOG.md`,
      "--include-path",
      `packages/${name}/**/*`,
      "--output",
      "./CHANGELOG.md",
    ],
    {
      cwd: `packages/${name}`,
    }
  )
}

```

### Expected behavior

Some caching mechanisms to improve the performance. Git cliff shouldn't check the older commits every time to generate the changelog.

### Screenshots / Logs

_No response_

### Software information

- Operating system: Linux 
- Rust version: N/A
- Project version:     "git-cliff": "2.4.0"

### Additional context

_No response_

aminya:
## Description

This caches the commit retain checks whenever include/exclude paths are specified.

This speeds up changelog generation by `258` times in a big repository I have:
```
Now: 0.080 s
Before:  20.633 s
```

Fixes #767 

It also improves handling including/excluding patterns, considering the first commit, and various other improvements.

## How Has This Been Tested?

I tested this on the repository mentioned in #767

Here are the profiling flamegraphs created via #768. The bottleneck is removed after this fix is applied as seen in the above image (4366 samples to only 16 samples). After the fix, other parts of the code finally show up in the profiler.

After - Fully Cached:

git-cliff - after cached - flamegraph

After - Cached + new commit:

git-cliff - after new commit - flamegraph

Before:

git-cliff - Before - flamegraph

## Types of Changes

- [x] Other 

## Checklist:

### DIFF
--- a/git-cliff-core/Cargo.toml
+++ b/git-cliff-core/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "git-cliff-core"
-version = "2.4.0" # managed by release.sh
+version = "2.4.0"                                               # managed by release.sh
 description = "Core library of git-cliff"
 authors = ["git-cliff contributors <git-cliff@protonmail.com>"]
 license = "MIT OR Apache-2.0"
@@ -75,6 +75,7 @@ futures = { version = "0.3.30", optional = true }
 url = "2.5.2"
 dyn-clone = "1.0.17"
 urlencoding = "2.1.3"
+cached = { version = "0.53.1", features = ["disk_store"], default-features = true}
 
 [dependencies.git2]
 version = "0.19.0"
--- a/git-cliff-core/Cargo.toml
+++ b/git-cliff-core/Cargo.toml
@@ -52,6 +52,7 @@ lazy_static.workspace = true
 thiserror = "1.0.61"
 serde = { version = "1.0.203", features = ["derive"] }
 serde_json = "1.0.118"
+bincode = "2.0.0-rc.3"
 serde_regex = "1.1.0"
 tera = "1.20.0"
 indexmap = { version = "2.2.6", optional = true }
--- a/git-cliff-core/Cargo.toml
+++ b/git-cliff-core/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "git-cliff-core"
-version = "2.4.0"                                               # managed by release.sh
+version = "2.4.0" # managed by release.sh
 description = "Core library of git-cliff"
 authors = ["git-cliff contributors <git-cliff@protonmail.com>"]
 license = "MIT OR Apache-2.0"
--- a/git-cliff-core/Cargo.toml
+++ b/git-cliff-core/Cargo.toml
@@ -75,7 +75,7 @@ futures = { version = "0.3.30", optional = true }
 url = "2.5.2"
 dyn-clone = "1.0.17"
 urlencoding = "2.1.3"
-cached = { version = "0.53.1", features = ["disk_store"], default-features = true}
+cacache = { version = "13.0.0", features = ["mmap"], default-features = false }
 
 [dependencies.git2]
 version = "0.19.0"
--- a/git-cliff-core/Cargo.toml
+++ b/git-cliff-core/Cargo.toml
@@ -99,6 +99,7 @@ features = ["debug-embed", "compression"]
 [dev-dependencies]
 pretty_assertions = "1.4.0"
 expect-test = "1.5.0"
+temp-dir = "0.1.13"
 
 [package.metadata.docs.rs]
 all-features = true
--- a/git-cliff-core/src/repo.rs
+++ b/git-cliff-core/src/repo.rs
@@ -87,10 +87,10 @@ impl Repository {
 		if include_path.is_some() || exclude_path.is_some() {
 			// Normalize the glob patterns
 			let include_patterns = include_path.map(|patterns| {
-				patterns.into_iter().map(Self::normalize_glob).collect()
+				patterns.into_iter().map(Self::normalize_pattern).collect()
 			});
 			let exclude_patterns = exclude_path.map(|patterns| {
-				patterns.into_iter().map(Self::normalize_glob).collect()
+				patterns.into_iter().map(Self::normalize_pattern).collect()
 			});
 
 			commits.retain(|commit| {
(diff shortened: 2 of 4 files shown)


alerque:
Only a measly 87× faster? Pshaw nothing to see here moving along now…

aminya:
> Only a measly 87× faster? Pshaw nothing to see here moving along now…

Challenge accepted. Now 258 times faster. 😄

aminya:
I changed the cache implementation to be more useful by caching the changed files of each commit.

This slows down the retain_commit check by ~2 times but allows reusing the cache in case the include/exclude patterns change. This is beneficial for running in monorepos where git-cliff runs multiple times to generate changelog for different subsets of the repo. This is a good trade-off in my opinion.

git-cliff-core/src/repo.rs
```diff
@@ -76,36 +86,104 @@ impl Repository {
...
+					return files;
+				}
+			}
+		}
+
+		// If the cache is not found, calculate the result and set it to the cache.
+		let result = self.commit_changed_files_no_cache(commit);
+
+		// Set the result to the cache.
+		match bincode::encode_to_vec(&result, bincode::config::standard()) {
```
orhun, line 144: Just curious why we need `bincode` here

orhun (commented):
🔥

git-cliff-core/src/repo.rs
```diff
@@ -36,15 +36,25 @@ pub struct Repository {
...
 
 impl Repository {
 	/// Initializes (opens) the repository.
 	pub fn init(path: PathBuf) -> Result<Self> {
 		if path.exists() {
+			let inner = GitRepository::open(&path)?;
+
+			// hash the path to create a unique cache id
+			let changed_files_cache_path =
+				inner.path().join("git-cliff").join("changed_files_cache");
```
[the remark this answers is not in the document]
aminya: It is saved under `.git/git-cliff/changed_files_cache`. That's the best place for it in my opinion as it is local to the repository. We can also use this path in future if will want to store more things.

git-cliff-core/src/repo.rs
aminya: Bincode helps with serializing the paths so that they can be stored in the cache. I tried serde_json first, but the whole generation ended up being 2.5 times slower than using bincode. So, I chose bincode to avoid adding too much caching overhead.

git-cliff-core/Cargo.toml
```diff
@@ -52,6 +52,7 @@ lazy_static.workspace = true
 thiserror = "1.0.61"
 serde = { version = "1.0.203", features = ["derive"] }
 serde_json = "1.0.118"
+bincode = "2.0.0-rc.3"
```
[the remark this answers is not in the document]
aminya: Since the Rc version has been in continuous development, and since it has been in the RC state for quite some time, it works stable. `v1` looks stale (released 3 years ago).
 
There are some breaking changes between the two, and I chose the Rc to avoid needing to upgrade this later once the final tag comes.
https://github.com/bincode-org/bincode/releases

aminya:
I fixed some bugs in the last three commits to improve the include/exclude behaviour:
- The include/exclude patterns are considered together.
- The glob patterns are normalized so that `./` doesn't make a difference in the result
- Added special case for the first commit where it has no parent

git-cliff-core/Cargo.toml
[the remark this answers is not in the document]
orhun: Makes sense, thanks for the explanation!

git-cliff-core/src/repo.rs
```diff
@@ -75,37 +85,152 @@ impl Repository {
...
+					!exclude_pattern
+						.iter()
+						.any(|pattern| pattern.matches_path(path))
+				});
+			}
+			(None, None) => true,
+		}
+	}
+
+	fn commit_changed_files(&self, commit: &Commit) -> Vec<PathBuf> {
```
orhun, line 161: Can you add comments explaining what this function and `_no_cache` variant is doing? It's hard to figure it out while reading it through.

git-cliff-core/src/repo.rs
```diff
@@ -75,37 +85,152 @@ impl Repository {
...
+	fn normalize_glob(pattern: Pattern) -> Pattern {
+		match pattern.as_str().strip_prefix("./") {
+			Some(stripped) => Pattern::new(stripped)
+				.expect("Removing the leading ./ will not fail"),
+			None => pattern,
+		}
+	}
+
+	/// Calculates whether the commit should be retained or not.
+	fn should_retain_commit(
```
orhun, line 117: Can you add unit tests for this?

git-cliff-core/src/repo.rs
```diff
@@ -75,37 +85,152 @@ impl Repository {
...
+					&include_patterns,
+					&exclude_patterns,
+				)
 			});
 		}
 		Ok(commits)
 	}
 
+	/// Normalizes the glob pattern by removing the leading `./`.
+	fn normalize_glob(pattern: Pattern) -> Pattern {
```
orhun, line 108: I feel like this can be moved into `commits` function (as a inner function maybe) since it is not used anywhere + not being tested.

git-cliff-core/src/repo.rs
aminya: Added a comprehensive test suit in [`33f56cf` (#772)](https://github.com/orhun/git-cliff/pull/772/commits/33f56cf2d916106168afc5b21ec8685c573304bf)

git-cliff-core/src/repo.rs
aminya: Added docs in [`7c63698` (#772)](https://github.com/orhun/git-cliff/pull/772/commits/7c6369876cb1ea5ad9ba85d17ba1c98677b020ca)

git-cliff-core/src/repo.rs
aminya: It's been used in the tests now as well, so it's better to keep it separate.

orhun (approved):
LGTM, can you resolve conflicts caused by #709 ? 😇

orhun (approved):
Awesome, thanks!

PR #784, opened 2024-08-04
feat(cli): generate changelog from JSON context

### LINKED ISSUE #657: Support generating changelog from a JSON context

### Feature description

`git-cliff` currently supports printing out the context via `-x` argument:

```sh
$ git cliff -x | jq . | head -n 5

[
  {
    "version": null,
    "commits": [
      {
```

It would be nice to generate a changelog from this JSON context as well.

### Desired solution

A new command-line argument might be introduced or we can take the context from `stdin`:

```sh
$ git-cliff -x > context.json

$ git-cliff --from-context context.json -o CHANGELOG.md

$ cat context.json | git-cliff
```

### Alternatives considered

None.

### Additional context

https://git-cliff.org/docs/templating/context

closes #773

janbuchar:
## Description

This implements generating a changelog from a pre-generated JSON context, which allows for more extensibility.

## Motivation and Context

- closes #657

## How Has This Been Tested?

As of now, only manual tests have been performed. I'm not sure what the best strategy for unit-testing this would be.

## Types of Changes

- [ ] Other 

## Checklist:


janbuchar:
This would be more useful if we added an optional free-form dictionary (e.g., `additional_data`) to both `Release` and `Commit`. In the current state, we may only change the existing context fields.

orhun (commented):
Damn, this is nice! Thanks for going ahead and implementing it 🔥 

> This would be more useful if we added an optional free-form dictionary (e.g., additional_data) to both Release and Commit. In the current state, we may only change the existing context fields.

I definitely agree, would love to have that. We can use `serde_json::Value` for that I reckon.

A couple of things:
- We should update the documentation about this. We can create a new page under `/docs/usage`. (I'm okay with adding it if you are not interested :))
- Can you add a [fixture test](https://github.com/orhun/git-cliff/tree/main/.github/fixtures) to verify this behavior?

janbuchar:
> Damn, this is nice! Thanks for going ahead and implementing it 🔥
> 
> > This would be more useful if we added an optional free-form dictionary (e.g., additional_data) to both Release and Commit. In the current state, we may only change the existing context fields.
> 
> I definitely agree, would love to have that. We can use `serde_json::Value` for that I reckon.

Great! Is `extra` an acceptable name? The originally proposed `additional_data` is also an OK option, but a bit long IMO.

> * We should update the documentation about this. We can create a new page under `/docs/usage`. (I'm okay with adding it if you are not interested :))

If you could just push a file with some kind of an outline here, I'll be glad to fill it in.

> * Can you add a [fixture test](https://github.com/orhun/git-cliff/tree/main/.github/fixtures) to verify this behavior?

Absolutely, once we've settled on the extra attribute name.

git-cliff-core/src/changelog.rs
```diff
@@ -39,8 +42,16 @@ pub struct Changelog<'a> {
...
 	/// Constructs a new instance.
 	pub fn new(releases: Vec<Release<'a>>, config: &'a Config) -> Result<Self> {
+		let mut changelog = Changelog::build(releases, config)?;
+		changelog.process_commits();
+		changelog.process_releases();
+		changelog.add_remote_data()?;
+		Ok(changelog)
+	}
+
+	fn build(releases: Vec<Release<'a>>, config: &'a Config) -> Result<Self> {
```
janbuchar, line 52: Maybe this should be `new` and `new` should be `from_git_releases` or something.

orhun: I'm okay with both as long as we don't change the public API a lot :)

orhun:
Yes, I think `extra` (or `extra_data` or `custom_data`) is a good name!

git-cliff-core/src/changelog.rs
janbuchar: Let's keep it this way then :slightly_smiling_face:

janbuchar:
@orhun I added the `extra` attribute and a fixture test. However, it seems like there is a bug - the `commit.extra` field gets lost somewhere on its way into the template. I found that it is still present in `changelog.rs#542` (I did a `dbg!(&release)` there), but it is not present inside the `Template.render` method. How can that happen?

orhun:
Ah, there is a custom `Serialize` implementation for `Commit` where you should add the new field :)

https://github.com/orhun/git-cliff/blob/a71649b6cd5a72c2a6b20c08dc1b8dfa37c19a6b/git-cliff-core/src/commit.rs#L397

git-cliff-core/src/commit.rs
```diff
@@ -30,14 +30,15 @@ use serde::{
...
+use serde_json::value::Value;
 
 /// Regular expression for matching SHA1 and a following commit message
 /// separated by a whitespace.
 static SHA1_REGEX: Lazy<Regex> = lazy_regex!(r#"^\b([a-f0-9]{40})\b (.*)$"#);
 
 /// Object representing a link
 #[derive(Debug, Clone, Eq, PartialEq, Deserialize, Serialize)]
-#[serde(rename_all = "camelCase")]
+#[serde(rename_all(serialize = "camelCase"))]
```
orhun, line 41: Why is this required btw?

janbuchar: IIRC, it expected camelCase fields in the input JSON before I added this. I'll check if it's really necessary.

janbuchar: Yes, looks like it's not necessary for `Link` , but for `Release` and `Commit`, it is.

.github/fixtures/test-from-context/expected.md
```diff
@@ -0,0 +1,31 @@
...
+
+### Bug Fixes
+
+- Fix feature 2 (that was a tough one)
+
+### Features
+
+- Add feature 2
+
+## [0.1.0] - 2022-04-06
```
janbuchar, line 21: I keep getting 2022-04-05 in the tests - probably because of some timezone issues. Locally, I get the same behavior in `test-commit-preprocessors` as well. What should I do?

janbuchar: Ah, it's probably because the `context.json` was created in a different timezone. I'll update this.

janbuchar:
@orhun I added a simple docs page and everything except Docker automated builds passed - those are still running.

orhun (approved):
Hot stuff 🔥 

I pushed some commits to improve this even further:
- now supports `GIT_CLIFF_CONTEXT` environment variable
- remote data is made available via calling `add_remote_data` on context
- support stdin out of the box via checking `is_terminal()` (e.g. `cat context.json | git-cliff`)
- updated docs

I think this opens so many doors as an extension point, so thanks a lot again! <3

orhun:
Oops, `is_terminal()` check seems to have broken the fixture tests, I will look into it later.

ref: 135938f698470974860c43f1b25414ebc46a9d73

janbuchar:
Thanks for the review and for the improvements, I can't wait to use it!

PR #807, opened 2024-08-22
feat(args): allow GitLab groups with `--gitlab-repo`

weichweich:
## Description

The cli argument `--gitlab-repo` is split at `/` and the last two parts are used as `owner` and `repo`. This PR changes this behavior, splits at the right most `/` and takes the first part as `owner` and the second part as `repo`.

| |raw string|owner|repo|
|--|--|--|--|
|before| `gitlab/group/a/repo_name` | `a` | `repo_name` |
|after this PR| `gitlab/group/a/repo_name` | `gitlab/group/a` | `repo_name` |

## Motivation and Context

GitLab allows projects to be organized in groups. Groups can also be part of groups. Because of that projects are not identified by `owner` and `repo`, but by `group/path` and `repo`.

The same behavior can be achieved by using the following inside the config:

```
[remote.gitlab]
owner = "gitlab/group/a"
repo = "repo_name"
```

But since I would like to use the same config in multiple projects, I rather provide the `remote` information via an environment variable.

## How Has This Been Tested?

I tested this with GitLab projects that are organized in groups.

There was a testcase that ignored the `https://github.com` domain if it was included. Since I never saw this inside the documentation I simply removed it, but I'm happy to adjust this PR in a way that this test would still pass, if needed.

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/git-cliff/src/args.rs
+++ b/git-cliff/src/args.rs
@@ -346,9 +346,12 @@ impl TypedValueParser for RemoteValueParser {
 	) -> Result<Self::Value, clap::Error> {
 		let inner = clap::builder::StringValueParser::new();
 		let value = inner.parse_ref(cmd, arg, value)?;
-		let parts = value.split('/').rev().collect::<Vec<&str>>();
-		if let (Some(owner), Some(repo)) = (parts.get(1), parts.first()) {
-			Ok(RemoteValue(Remote::new(*owner, *repo)))
+		let parts = value.rsplit_once('/');
+		if let Some((owner, repo)) = parts {
+			Ok(RemoteValue(Remote::new(
+				owner.to_string(),
+				repo.to_string(),
+			)))
 		} else {
 			let mut err = clap::Error::new(ErrorKind::ValueValidation).with_cmd(cmd);
 			if let Some(arg) = arg {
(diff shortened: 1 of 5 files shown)


orhun:
Thanks for the PR! We should definitely include this fix in the upcoming release.

> There was a testcase that ignored the https://github.com/ domain if it was included. Since I never saw this inside the documentation I simply removed it, but I'm happy to adjust this PR in a way that this test would still pass, if needed.

Yeah, the current state of this PR is breaking since it splits on the last occurrence. We need to somehow support both cases.

git-cliff/src/args.rs
```diff
@@ -346,9 +346,12 @@ impl TypedValueParser for RemoteValueParser {
 	) -> Result<Self::Value, clap::Error> {
 		let inner = clap::builder::StringValueParser::new();
 		let value = inner.parse_ref(cmd, arg, value)?;
-		let parts = value.split('/').rev().collect::<Vec<&str>>();
-		if let (Some(owner), Some(repo)) = (parts.get(1), parts.first()) {
-			Ok(RemoteValue(Remote::new(*owner, *repo)))
+		let parts = value.rsplit_once('/');
```
orhun, line 349: One idea is to parse it as URL then split it (just to be safe):

```suggestion
		let mut value = inner.parse_ref(cmd, arg, value)?;
		if let Ok(url) = Url::parse(&value) {
			value = url.path().trim_start_matches('/').to_string();
		}
		let parts = value.rsplit_once('/');
```

(you need to add `url` crate to the workspace dependencies)

orhun (approved):
I'm planning a release soon so I will merge this as-is, thank you!

weichweich:
Sorry, did read it just now. Thanks a lot for adjusting the code and merging it so quickly. And of course for building such an amazing tool! :heart:

PR #808, opened 2024-08-22
feat(changelog): support generating changelog for different branches

### LINKED ISSUE #750: --unreleased does not behave as expected with tags on side branches

### Description of the bug

If I have a project where tags are also created on side branches and not only on the master branch, git cliff seems to get confused, if the `--unreleased` flag is used.
I would expect to see only commit messages parsed that do not belong to a commit, but this does not seem to be the case.
If on a side branch a new tag was created to fix a older version also already tagged commits form the master branch start to appear in the change log. It seams the last commit in time on any branch is used to determine the unreleased commits.

### Steps To Reproduce

1. Create tag `2.0.0` on branch `master`
2. Create tag `1.0.1` on branch e.g. `release/1`
3. Create tag `2.1.0` on head of  `master`
4. Add new commit on `master`
5. Create a change log on `master` with `git cliff -u`

### Expected behavior

It is expected that only the newly created commit on `master` appears in the change log, created by git cliff.

But what I see instead, that also the commits of the tag 2.0.0 are included.

### Screenshots / Logs

image001

### Software information

- Operating system: WSL - Ubuntu 22.04
- Rust version: 
- Project version: 2.4.0

### Additional context

As soon as a new tag is created on the `master` branch, the behavior is again as expected.

braineo:
## Description

close #750 

## Motivation and Context

When a project needs to maintain diverged branches and create changelog for each branch this change is necessary. 

for example a project's v3 is a complete rewrite and the changes onward have nothing to do which v2 branch. While v2 is in maintenance mode that the fixes are not relevant to v3 anymore.

when creating changelog for v3, it is necessary not to include the changes in the v2 which are not even included in the current branch

## Types of Changes

- [ ] Other 

## Checklist:


braineo:
although the result of `test-topo-order` is not matching the `expected.md` but i feel like the new behavior should probably be the expected result.

the test commit first tags v0.1.0 and v0.2.0, then it checks out v0.1 and creates a new v0.1.1 tag.

in my opinion the expected full changelog should contain v0.1.0 and v0.1.1 instead of v0.2.0 only see #750 @chrissi-p

git-cliff-core/src/repo.rs
```diff
@@ -310,6 +310,17 @@ impl Repository {
 		None
 	}
 
+	fn should_include_tag(
```
orhun, line 313: Can you explain this function a bit?

Are you simply checking if a commit belongs to a branch?

A descriptive comment would be really helpful.

orhun (commented):
Thanks for the PR! The issue is more clear to me now and I'm considering to switch to this new behavior. However, I need some explanations about how this is supposed to work :)

orhun:
Also, any idea why the topo order fixture is failing? I'm trying to understand the change of behavior and its relation to the topo order 🤔

braineo:
> Also, any idea why the topo order fixture is failing? I'm trying to understand the change of behavior and its relation to the topo order 🤔

Because the tags chosen for the changelog are different now.

git-cliff-core/src/repo.rs
braineo: sure updated

braineo:
@orhun i've updated the description and the test cases. can you take a look at it plz?

orhun:
Well, I'm not fully convinced to go ahead with this. It breaks the topo order argument in a way that I don't understand.

> Because the tags chosen for the changelog are different now.

Maybe you can expand this a bit?

orhun:
Maybe there is a way of supporting both cases if the commits are on other branches?

I'm not really sure if this fix is worth breaking the topo order argument.

braineo:
> Well, I'm not fully convinced to go ahead with this. It breaks the topo order argument in a way that I don't understand.
> 
> 
> 
> > Because the tags chosen for the changelog are different now.
> 
> 
> 
> Maybe you can expand this a bit?

those features are independent from each other. My change only affects "what" commits are chosen, while leaving "how" they are sorted untouched. 

See the chart below if it helps:	

When one project has both v1 and v2 branches. It is typical that v1 branch change log does not include the changes from v2 since the code never exists in v1. And vice versa. 

image

braineo:
hi @orhun 

is the explanation good enough or you are still looking into more details?

orhun:
> those features are independent from each other. My change only affects "what" commits are chosen, while leaving "how" they are sorted untouched.

Okay I see now. And thanks for the diagram 🙂 

How about we do something like this to support both cases?

```rust
self.should_include_tag(&head_commit, &commit)? || topo_order
```

braineo:
> > those features are independent from each other. My change only affects "what" commits are chosen, while leaving "how" they are sorted untouched.
> 
> Okay I see now. And thanks for the diagram 🙂
> 
> How about we do something like this to support both cases?
> 
> ```rust
> self.should_include_tag(&head_commit, &commit)? || topo_order
> ```

@orhun maybe i don't fully understand the purpose of topo_order, it seems when not specified, it uses whatever order `git2` returns for the tags. 

it would be useful for any chosen set of the tags. with the suggested change, in order to use topo_order one must include all the tags in the changelog.

maybe you mean some configuration like `filter_tags=false`?

orhun:
> it seems when not specified, it uses whatever order git2 returns for the tags.

Yeah, the tags are sorted by date as default.

> with the suggested change, in order to use topo_order one must include all the tags in the changelog.

That's correct, we simply disable this new `should_include_tag` functionality with the code that I suggested above, since they work as expected together.

> maybe you mean some configuration like filter_tags=false?

Do you mean adding a new configuration option?

braineo:
> > with the suggested change, in order to use topo_order one must include all the tags in the changelog.
> 
> That's correct, we simply disable this new `should_include_tag` functionality with the code that I suggested above, since they work as expected together.
> 
> > maybe you mean some configuration like filter_tags=false?
> 
> Do you mean adding a new configuration option

yes. i would suggest adding another option to toggle the tag filtering explicitly. because in the case when the tags are filtered, it is also desired to use the topo order but not the chronicle order. i still think filtering the tags and sorting the tags are independent from each other. with the suggested change, we will find two cases not available in the function matrix.

|                 | filter tags | not filter tags |
|-----------------|-------------|-----------------|
| topo order      | ❌          | ⭕              |
| chronicle order | ⭕          | ❌              |

braineo:
@orhun by separating the sort and filter option. it should be easier to add new sorting method like sort by semver version number etc. wdyt?

orhun:
> yes. i would suggest adding another option to toggle the tag filtering explicitly

yeah, I think that's the most sensible thing to do for supporting both use cases at this point.

> by separating the sort and filter option. it should be easier to add new sorting method like sort by semver version number etc. wdyt?

Do you mean something like:

```toml
sort_tags = "topo"
sort_tags = "date" # or chronically
sort_tags = "semver"
```

braineo:
@orhun ok i added a new flag to opt-in.

> Do you mean something like:
> 
> ```toml
> sort_tags = "topo"
> sort_tags = "date" # or chronically
> sort_tags = "semver"
> ```

yes, if in the future there are more options added to sorting, the change should not be breaking the tag filtering.

braineo:
hi @orhun 

what do you think of the latest changes?

orhun (approved):
LGTM, thanks for the contribution & bearing with me :)

Applied some nitpicks and approved!

PR #822, opened 2024-08-27
feat(core): add `remote` to commit and deprecate fields

### LINKED ISSUE #803: git-cliff-core: allow adding arbitrary values to commits

### Feature description

First of all thank you for the amazing crate and the hard work you put into it 🙏

I would like to attach a `remote` context to the commit, so that people can write `commit.remote.username` in release-plz changelog (which is similar to `commit.github.username` git-cliff equivalent.

Even though I can't have `commit.remote` I tried enabling the `"github"` feature and my Cargo.lock exploded because of many dependencies to retrieve remote info I'm not using (because I'm retrieving remote info by myself). So I would like to have a "lighter" way to add arbitrary fields to the `commit` object.

So a very similar mechanism to https://github.com/orhun/git-cliff/pull/613 but limited to the `commit` object.

### Desired solution

The `Commit` struct could have a Map of <String, Json> (this is psudo-code). In my example, "String" is `remote` and Json is `{ author }`.

Library users should be able to add any fields they like to it.

### Alternatives considered

The alternative is rely entirely on git-cliff for commits remote info retrieval.

My issue is that I generate multiple changelogs that share some commits, so we would hit GitHub API multiple times for the same commits for no reason.
We could solve this by implementing https://github.com/orhun/git-cliff/issues/468 but I'm not sure if it's easier than this issue.

Also, with this approach, the following problems remain:
- I would like to have `commit.remote` instead of `commit.github`, `commit.gitea`, etc. (unless you change my mind that `commit.github` is a better idea 👀)
- I prefer to avoid adding all those dependencies because release-plz already takes a while to compile 😵

### Additional context

PR where I experimented: https://github.com/MarcoIeni/release-plz/pull/1630

Also, I'm curious, why you have `commit.github`, `commit.gitea`, etc instead of a unified solution (e.g. `commit.remote`)?

MarcoIeni:
## Description

Close #803

Add the `remote` field so that libraries using git-cliff-core can fill it. git-cliff can later use it after a refactoring.

## Motivation and Context

https://github.com/orhun/git-cliff/issues/803#issuecomment-2310832669

## How Has This Been Tested?
It's a trivial change, so I haven't tested it.

## Types of Changes

- [ ] Other 

## Checklist:


git-cliff-core/src/commit.rs
```diff
@@ -125,18 +125,20 @@ pub struct Commit<'a> {
 	pub merge_commit:  bool,
 	/// Arbitrary data to be used with the `--from-context` CLI option.
 	pub extra:         Option<Value>,
+	/// Remote metadata of the commit.
+	pub remote:        Option<crate::remote_contributor::RemoteContributor>,
 	/// GitHub metadata of the commit.
 	#[cfg(feature = "github")]
-	pub github:        crate::remote::RemoteContributor,
+	pub github:        crate::remote_contributor::RemoteContributor,
```
MarcoIeni, line 132: I moved the `RemoteContributor` from the `remote` module because the `remote` module is under the `remote` feature flag.

MarcoIeni:
Oh, I didn't get you wanted to move from `commit.github` to `commit.remote` in git-cliff as well. Cool 😁

Then, I'll work on this 👍

MarcoIeni:
> Update the feature gates for remote so that the types can be exposed without requiring the integration to be enabled.

I extracted the RemoteContributor into `git-cliff-core/src/remote_contributor.rs` already. Do you think something else is necessary?

orhun:
Thanks for the update on the PR! A couple of points:

1. We should either tune down the log level or check the config once and print warnings to avoid flooding the logs:

image

2. We need to check for the presence of `commit.remote` variable alongside of e.g. `github::TEMPLATE_VARIABLES` constants.

3. I don't like the place where `remote_contributor.rs` is sitting right now. Maybe we can rename it to `contributor.rs` or move it somewhere else. (I'm open to other ideas).
 
4. Would be nice to update the existing examples/configs 🙂

MarcoIeni:
I addressed all points except 2. Could you handle it since it might be a bit involved? 👀

git-cliff-core/src/commit.rs
```diff
@@ -459,17 +466,35 @@ impl Serialize for Commit<'_> {
...
-		commit.serialize_field("bitbucket", &self.bitbucket)?;
+		serialize_remote::<S>(&mut commit, "bitbucket", &self.bitbucket)?;
+		if let Some(remote) = &self.remote {
+			commit.serialize_field("remote", remote)?;
+		}
 		commit.end()
 	}
 }
 
+fn serialize_remote<S>(
```
orhun, line 483: Why do we need this method? Also, can you document it? 😊

website/docs/integration/github.md
orhun, line 1: Thanks for updating the docs! Can you also update the other usages of `commit.github` etc. in the same document as well?

git-cliff-core/src/changelog.rs
```diff
@@ -612,6 +605,33 @@ impl<'a> Changelog<'a> {
...
+		"gitea",
+		"gitlab",
+		"bitbucket",
+	]);
+	body_template
+}
+
+fn warn_on_deprecated_commit_vars(body_template: &str, vars: &[&str]) {
+	for var in vars {
+		if body_template.contains(&format!("commit.{var}")) {
```
orhun, line 626: Unfortunately performing a string check here is not going to be reliable because it is going to print out a warning even though I just use a "commit.github" as a string in the template.

Instead, we should use the `contains_variable` function which does a proper check in the template variables :)

orhun (commented):
Looks good! Just needs a bit of tweaks ⚙️

git-cliff-core/src/commit.rs
MarcoIeni: removed in https://github.com/orhun/git-cliff/pull/822/commits/04e62efc17772d4e20809e98b332f4ff678fb740 as it was no longer needed 👍
Initially I didn't know you wanted git cliff commit to move to `commit.remote` so I thought I had to fill `commit.remote` with `commit.github` info. That's why I needed the method 👍

git-cliff-core/src/changelog.rs
MarcoIeni: done in [5a67a20](https://github.com/orhun/git-cliff/pull/822/commits/5a67a2066761ea40edb05bc7634a27a4e5841fef)

website/docs/integration/github.md
MarcoIeni: done in [9ee9571](https://github.com/orhun/git-cliff/pull/822/commits/9ee9571d99c48ed1e55a2ba97a405b94159cea84)

orhun:
Hmm, that's not good. Why are we not serializing it? I didn't get why the fixtures are failing because using `commit.remote.*` locally works for me.

MarcoIeni:
Imo the issue is that the git-cliff cli never sets the `remote` field of the struct here:
https://github.com/MarcoIeni/git-cliff/blob/aee8952fbdc191aca5663205bf3cd3e76f5f066d/git-cliff-core/src/release.rs#L64

MarcoIeni:
Yeah 😎

orhun (approved):
Perfect, thanks for your contribution @MarcoIeni ❤️ 

Can you also shoot me an issue about removing the `github`/`gitlab` etc. fields for the next release? Currently there is quite a bit of duplication in the context which leads to more memory usage and it would be better to get rid of those things after the next release :)

MarcoIeni:
Thanks for reviewing and merging. Looking forward for the release!

I created the issue as you asked https://github.com/orhun/git-cliff/issues/856 👍

Issue #618, opened 2024-04-21
Use stable rust for cargo-fmt


MarcoIeni:
### Is there an existing issue or pull request for this?

- [X] I have searched the existing issues and pull requests

### Feature description

Not really a feature.
working with cargo nightly for cargo fmt is painful :(
Please, consider switching to stable to make the experience of contributors a bit better.

### Desired solution

- Remove rust-toolchain.toml file
- Run cargo fmt

Downside: all opened PRs will have merge conflicts probably, so it's better to do this when there aren't big PRs opened.

### Alternatives considered

The alternative is keeping the current configuration I guess.

### Additional context

feedback coming from https://github.com/orhun/git-cliff/pull/613#issuecomment-2059908791

alerque:
> Downside: all opened PRs will have merge conflicts probably because stable uses spaces instead of tabs.

Merge conflicts for this (and other similar cases where deterministic tooling is used to restyle code) by applying the new formatting rules at the end of the PR/branch before attempting to merge. Optionally this can be later squashed in with the merge commit itself so that the merge diff only shows what the PR brought to the table *after* all the formatting matched on both sides. If a rebase is preferred this can be done by applying formatting on each commit in the branch **in reverse order** (branch HEAD first, then working backwards).

joshka:
I'm rather partial to many unstable formatting options, particularly comment width related ones and import ordering / grouping, and would suggest keeping the nightly formatting around.

The unstable options in use in this repo are:
```toml
binop_separator = "Back" # https://github.com/rust-lang/rustfmt/issues/3368
format_code_in_doc_comments = true # https://github.com/rust-lang/rustfmt/issues/3348
format_strings = true # https://github.com/rust-lang/rustfmt/issues/3353
imports_layout = "Vertical" # https://github.com/rust-lang/rustfmt/issues/3361
normalize_doc_attributes = true # https://github.com/rust-lang/rustfmt/issues/3351
overflow_delimited_expr = true # https://github.com/rust-lang/rustfmt/issues/4991
struct_field_align_threshold = 20 # https://github.com/rust-lang/rustfmt/issues/3371
wrap_comments = true # https://github.com/rust-lang/rustfmt/issues/3347
```

I agree that nightly should not be in the toolchain config however.

An alternative to using nightly to format is to pass the [unstable options directly to rustfmt](https://github.com/rust-lang/rustfmt/issues/3387#issuecomment-2125874325). E.g. this would look like:

```shell
[...]

PR #834, opened 2024-08-31
chore: use stable rustfmt

marcoieni:
## Description
close https://github.com/orhun/git-cliff/issues/618

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,15 +1,5 @@
-edition = "2018"
 max_width = 85
 newline_style = "Unix"
 hard_tabs = true
-tab_spaces = 4
 use_field_init_shorthand = true
 reorder_imports = true
-binop_separator = "Back"
-format_code_in_doc_comments = true
-format_strings = true
-imports_layout = "Vertical"
-normalize_doc_attributes = true
-overflow_delimited_expr = true
-struct_field_align_threshold = 20
-wrap_comments = true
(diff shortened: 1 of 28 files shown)


orhun:
We should probably cut a new point release just for these changes 💀 

Do you know if there is a way of keeping the git blame while changing the formatting? I'm guessing it shouldn't be possible but it would be super nice if we can keep the historic footprint on the files somehow.

MarcoIeni:
Yes, it's possible:
- https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view
- https://www.stefanjudis.com/today-i-learned/how-to-exclude-commits-from-git-blame/

Should I add the `.git-blame-ignore-revs` file?

orhun:
Wow perfect, please do it 👍🏼

MarcoIeni:
Actually, I think we need to do this in another PR, after merging this so that we know the commit to ignore 👍

orhun:
Can you push this again? There seems to be some conflicts.

> Actually, I think we need to do this in another PR, after merging this so that we know the commit to ignore 👍

I won't do a squash merge for this PR so it should be safe to use the commit hash :)

rustfmt.toml
MarcoIeni, line 1: I removed all unstable options. Plus `edition` is read from Cargo.toml

MarcoIeni:
Read for another look :)

orhun:
I think I will get to this when I'm doing a major / `v3.0.0` release. The development is ongoing and I can't find a good time to push a release for this.

Issue #879, opened 2024-09-22
ARM64 docker images take 6h+ to build


M0NsTeRRR:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description of the bug

https://github.com/orhun/git-cliff/actions/runs/10978847295/job/30482488830

The github action failed to build and push 2.6.0 tag to ghcr.io

The Docker job times out after 360 minutes, as described in the [GitHub documentation](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes) 

### Steps To Reproduce

.

### Expected behavior

.

### Software information

Github action

orhun:
Thanks for reporting!

Yup, I tried running the workflow 2 more times, they have all failed: https://github.com/orhun/git-cliff/actions/runs/10978847295

I think it is timing out due to QEMU/AMD builds :/ 

According to [this docs](https://docs.github.com/en/actions/administering-github-actions/usage-limits-billing-and-administration#usage-limits), it is not possible to increase this limit.

> Job execution time - Each job in a workflow can run for up to 6 hours of execution time. If a job reaches this limit, the job is terminated and fails to complete.

Not sure what can I do expect disabling them temporarily, which I did in cde2a8e3222f5e8f8bdd9ae841fd0e5c42f68846

M0NsTeRRR:
@orhun Looks like an issue with github action, the docker is running now ? (could you rerun the 2.6.0 docker action)

orhun:
Yup, I need to create a new release soon. I don't think it is possible to trigger the pipeline again with the new changes in the workflow file :)

orhun:
In the meantime you can probably use one of the `latest` versions.

M0NsTeRRR:
Yup, I close this one as the issue seems resolved :)

LtdSauce:
Hi!

Sorry to re-heat this issue. I just stumbled upon this and thought that this must be doable in a way. (Building the arm64 image again I mean)

Hopefully I am not pointing to anything that is already known to you, but the docker docs mention how to achieve this again: https://docs.docker.com/build/ci/github-actions/multi-platform/ .

Basically building the images separately and then merge them together. AFAIK building the arm image could even be done on a native arm runner. I would submit a PR... But unfortunately I have no time for this at the moment:/ 
So hopefully I could at least help with the above link :)

orhun:
[...]

PR #888, opened 2024-09-26
feat(docker): use matrix for building multi-platform Docker images

orhun:
## Description

See <https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners>

## Motivation and Context

Docker builds are hella slow. This might be a way to speed them up.

See #879

## How Has This Been Tested?

Not tested yet.

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -13,12 +13,25 @@ on:
     - cron: "0 0 * * 0"
 
 jobs:
-  docker:
-    name: Docker Build and Push
+  build:
+    name: Build
     runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - linux/amd64
+          - linux/arm/v6
+          - linux/arm/v7
+          - linux/arm64
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+
+      - name: Prepare
+        run: |
+          platform=${{ matrix.platform }}
+          echo "PLATFORM_PAIR=${platform//\//-}" >> $GITHUB_ENV
 
       - name: Docker meta
         id: meta
@@ -34,6 +47,9 @@ jobs:
             type=sha
             type=raw,value=latest
             type=semver,pattern={{version}}
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
 
       - name: Set up Docker Buildx
         id: buildx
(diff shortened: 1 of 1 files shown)


LtdSauce:
Hey!

I already mentioned this in the issue that causes this PR, but i guess for the sake of completeness i will leave this here as well.
Afaik this can be done with using a native arm64 build node. Sadly those are only planned to be available for OSS Project by the end of the year. See [here](https://github.com/orgs/community/discussions/19197#discussioncomment-10761673).

LtdSauce (changes requested):

.github/workflows/docker.yml
```diff
@@ -62,25 +69,98 @@ jobs:
...
+      - name: Download digests
+        uses: actions/download-artifact@v4
+        with:
+          path: /tmp/digests
+          pattern: digests-*
+          merge-multiple: true
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
```
LtdSauce, line 118: I guess these can just be removed as MacOs-14 already uses arm64.

```suggestion
```

orhun:
Now we hit another thing: https://github.com/orhun/git-cliff/actions/runs/11075086138/job/30775267343

.github/workflows/docker.yml
```diff
@@ -62,25 +66,95 @@ jobs:
...
+    needs:
+      - build
+    steps:
+      - name: Download digests
+        uses: actions/download-artifact@v4
+        with:
+          path: /tmp/digests
+          pattern: digests-*
+          merge-multiple: true
+
```
LtdSauce, line 112:
```suggestion
          merge-multiple: true

      - name: Set up Docker Buildx
          uses: docker/setup-buildx-action@v3
```

Wow sorry... i just noticed i suggested to remove the wrong lines. The docs mentioned in the description have this step as well, so i guess it is needed for the Docker meta step as well.

This will not solve the issue with the missing docker socket on MacOs though...

LtdSauce (changes requested):

.github/workflows/docker.yml
M0NsTeRRR: I think it's pretty clear ...
https://github.com/docker/setup-buildx-action/issues/176

M0NsTeRRR:
> Now we hit another thing: https://github.com/orhun/git-cliff/actions/runs/11075086138/job/30775267343

https://github.com/docker/setup-buildx-action/issues/176

:/

M0NsTeRRR:
it looks like we can use colima https://github.com/actions/runner-images/issues/6216

LtdSauce:
> it looks like we can use colima [actions/runner-images#6216](https://github.com/actions/runner-images/issues/6216)

The last two comments mention it is (or at least was) not available on the macos-13 and -14 runners. So i suspect this has to be manually installd. [This](https://github.com/actions/runner-images/tree/main/images/macos) looks like it is not available on arm based macos nodes.

.github/workflows/docker.yml
LtdSauce: But still doesn't matter as macos on the M1 chip cannot do docker builds anyway.

orhun:
I really appreciate you investing time into this and making everything clear! @LtdSauce - sadly it is an unfortunate situtation.

> Another solution would be to somehow split the work that is done in the Dockerfile (basically the compilation) into multiple build stages, and only build one stage in one build step and then use the result in the next build step... but this feels needlessly complex.

Can't say that I fully understood it, but yeah, sounds complex for sure.

> Or what might work, though it is a little bit prone to dependency issues: build outside of docker and only copy the binaries into the docker image during the build.

It might work but sounds too hacky :/

M0NsTeRRR:
Splitting dockerfile will not help as the issue in the build is only triggered in `cargo build` (compilation)

LtdSauce:
Hey,

i have not enough knowledge about how to split up compilations in cargo. But if i get it right, than it might be possible to do the following as i think i found 3 compilation targets in the project. If i am wrong and there is no way to build things seperately, then please just ignore my comment. Otherwise maybe the following might work:

## First Adjust the Dockerfile:

Instead of doing a full build in https://github.com/orhun/git-cliff/blob/bfe9beb093df42badf762945cf04548fbe75949a/Dockerfile#L14 you could split that up into several build commands and put each in its own docker build stage.
E.g. something like this:
```
FROM builder AS lib-builder
RUN cargo build --release --locked --no-default-features --features github --features gitlab --features bitbucke --lib
FROM lib-builder AS bin-builder
RUN cargo build --release --locked --no-default-features --features github --features gitlab --features bitbucke --bin
```
And then in the runner stage use `--from=bin-builder` instead of `builder`.

Is it possible to build just the dependencies alone? If so that could be another stage before doing the lib-builder stage. Also as there are 2 binaries in the workspace, those could be build in their own stage as well.

This change is not visible to the user of the Dockerfile if one just runs `docker build`.

## Second split the workflow into multiple jobs:

After this something like https://github.com/djbender/github-actions-docker-caching-example/blob/main/.github/workflows/main.yml could be applied to cache each of the above layers in one build job. So basically have one build job for each layer and let them run in sequence. This is necessary as a build job needs to terminate after 6h... splitting the stages over the steps in one job would be easier but is not possible here.

I have no time to try this out this week though :/ But feel free to adapt my idea if you think it is worth it. (I'll try it out as soon as i have some spare time to fiddle around with this idea.)

And this can only work as long as the individual cargo build steps do not take more then 6h each.

orhun:
I think you are onto something there... it might work. We should also check out [`cargo-chef`](https://hub.docker.com/r/lukemathwalker/cargo-chef) to see if someone else did something similar.

I'm a bit burnt out by these Docker issues recently, feel free to try it out whenever you have time! I might come back to it next week :)

LtdSauce:
FTR: I left a comment in the issue that caused this PR as i am now quite certain, that the actual issue lies in the usage of `cargo-chef`. If the  proposed solution there is not sufficient to bring the arm64 builds back, then it could be combined with what i proposed here. But instead of doing multiple `cargo build` jobs, it should be enough to put the `cargo chef cook` in one job and the `cargo build` in another one... then both can take up to 6h to complete.

PR #889, opened 2024-09-27
docs(website): add conversion to pdf to tips-and-tricks

Krensi:
## Description

Using `pandoc` as separate command, the markdown output can be converted to pdf

## Motivation and Context

If you need to share the changelog as pdf this could be handy

### DIFF
--- a/website/docs/tips-and-tricks.md
+++ b/website/docs/tips-and-tricks.md
@@ -90,3 +90,15 @@ commit_parsers = [
 ```jinja2
 {{ get_env(name="CI_PROJECT_URL") }}/-/tags/{{ version }}
 ```
+
+## Convert the markdown output to pdf
+
+```bash
+pandoc --from=gfm --to=pdf -o CHANGELOG.pdf CHANGELOG.md
+```
+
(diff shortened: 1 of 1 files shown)


website/docs/tips-and-tricks.md
```diff
@@ -90,3 +90,15 @@ commit_parsers = [
 ```jinja2
 {{ get_env(name="CI_PROJECT_URL") }}/-/tags/{{ version }}
 ```
+
+## Convert the markdown output to pdf
```
orhun, line 94:
```suggestion
## Convert markdown output to PDF
```

website/docs/tips-and-tricks.md
```diff
@@ -90,3 +90,15 @@ commit_parsers = [
...
+
+```bash
+pandoc --from=gfm --to=pdf -o CHANGELOG.pdf CHANGELOG.md
+```
+
+To support unicode characters use `xelatex` as pdf engine and a font family which includes the needed unicode characters
```
orhun, line 100:
```suggestion
To support unicode characters, use `xelatex` as PDF engine and a font family which includes the needed unicode characters:
```

orhun (approved):
Awesome, thanks!

PR #901, opened 2024-10-02
fix(changelog): include the root commit when `--latest` is used with one tag

rarescosma:
## Description, Motivation and Context

This is a proof of concept implementation for the fix of including the `root` commit of a repository when `git-cliff` is called with the `--latest` flag and only ONE tag exists in the repository.

Ranges passed to to `Repository::commits` are left-exclusive. This works well when we have two or more tags:

```
- sha0 (old_tag)
- sha1
- sha2
- sha3 (new_tag)
```

We'll get the range argument `sha0..sha3`, which will make the `::commits` call capture only the `sha1`, `sha2` and `sha3` commits, which is both expected and great.

However, when dealing with a _first_ release, and a _single_ tag, we'd still like the root commit of the repository to be included in the report. Think of it as having a tag _before_ the root commit itself. 

In this case, with the current implementation of `::commits` we get the range `root_sha..single_tag_sha` which, being left-exclusive, will exclude the root commit from the report.

My implementation addresses this by adding a flag to `::commits` to tell it to include the root commit explicitly. When this boolean flag is present, we chop the range and include everything up to the `..single_tag_sha`.

## How Has This Been Tested?

Added both a unit test and adapted the `test-latest-with-one-tag` fixture for the integration tests.

## Types of Changes

- [ ] Other 

## Checklist:

- [ ] I have updated the documentation accordingly.  _NOTE: unsure where to update this, but would gladly do._


.github/fixtures/test-latest-with-one-tag/commit.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -e
 
-GIT_COMMITTER_DATE="2021-01-23 01:23:45" git commit --allow-empty -m "Initial commit"
+GIT_COMMITTER_DATE="2021-01-23 01:23:45" git commit --allow-empty -m "feat: initial commit"
```
orhun, line 4: I just tried out this fixture without changing the rest of the code and I'm getting the same result. Can you check if the behavior change is correct here?

orhun: Oops nvm, I was not using `--latest` :)

git-cliff-core/src/repo.rs
```diff
@@ -72,13 +72,21 @@ impl Repository {
...
 		if let Some(range) = range {
-			revwalk.push_range(range)?;
+			// Push only the last range commit if we need to start at root.
+			if include_root {
+				if let Some(dots_index) = range.find("..") {
+					revwalk.push(Oid::from_str(&range[dots_index + 2..])?)?;
+				}
+			} else {
+				revwalk.push_range(range)?;
+			}
```
orhun, line 89: Instead of adding a new flag to the function, we can fallback to using the given `range` string as-is if it does not contain two dot range notation.

e.g.

```rust
if range.contains("..") {
	revwalk.push_range(range)?;
} else {
	revwalk.push(Oid::from_str(&range)?)?;
}
```

Also see my other comment :)

git-cliff/src/lib.rs
```diff
@@ -160,18 +160,20 @@ fn process_repository<'a>(
...
 		}
 	} else if args.latest || args.current {
 		if tags.len() < 2 {
-			let commits = repository.commits(None, None, None)?;
+			let commits = repository.commits(None, false, None, None)?;
 			if let (Some(tag1), Some(tag2)) = (
 				commits.last().map(|c| c.id().to_string()),
 				tags.get_index(0).map(|(k, _)| k),
 			) {
 				commit_range = Some(format!("{tag1}..{tag2}"));
```
orhun, line 175: No need for an additional flag here, just doing this would be enough:

```rust
if tags.len() == 1 {
	commit_range = Some(format!("{tag2}"));
} else {
	commit_range = Some(format!("{tag1}..{tag2}"));
}
```

orhun (changes requested):
Looks good! The implementation could be simplified a bit :)

git-cliff-core/src/repo.rs
rarescosma: Nice, we have an algebraic type (Option) to decide between range and "everything" plus a special value to say "between root and this sha".

I wanted to make CommitRange an enum but then said "surely a boolean would be enough". 😂

Wdyt, is the blast radius of enum too much? If yes, then I'll gladly apply the proposed changes instead.

rarescosma: Went with the simple solution for now.

orhun: Let's keep it simple :)

git-cliff-core/src/repo.rs
```diff
@@ -78,7 +78,12 @@ impl Repository {
 		let mut revwalk = self.inner.revwalk()?;
 		revwalk.set_sorting(Sort::TOPOLOGICAL)?;
 		if let Some(range) = range {
-			revwalk.push_range(range)?;
+			if range.contains("..") {
+				revwalk.push_range(range)?;
+			} else {
+				// when we get a single sha as our "range" => start at root.
```
orhun, line 84: nit:

```suggestion
				// When a single SHA is provided as the "range", start from the root.
```

orhun (approved):

orhun:
btw can you revert 90e0202bcc69984a047904ccb41e28724c003da2 and fa85e9569d19179f1f990a750fe129bef5795dff

those are not in the scope of this PR

rarescosma:
Hi, I think the `main` linting configuration drifted a bit, so I included an extra commit to make the "pedantic" clippy lints pass. Also, not sure what's going on with the "Check NodeJS tarball" check. Everything else seems green. 🥳

rarescosma:
> btw can you revert [fa85e95](https://github.com/orhun/git-cliff/commit/fa85e9569d19179f1f990a750fe129bef5795dff)
> 
> those are not in the scope of this PR

Sure, np

orhun (approved):
Thanks!

PR #920, opened 2024-10-13
fix(changelog): fix missing commit fields in context (#837)

### LINKED ISSUE #837: `--from-context` doesn't include footers

### Description of the bug

```
 ERROR git_cliff > Template render error:
Variable `commit.footers` not found in context while rendering 'template'
```

### Steps To Reproduce

You can test with my repository's template at https://github.com/muja/unrar.rs, for example where I use footers.

`git cliff --context` gives `footers`, however

`git cliff --context | git cliff --context --from-context -` has no footers.

### Expected behavior

`footers` are read from context

### Screenshots / Logs

_No response_

### Software information

- Project version: 2.5.0

### Additional context

Possibly other values are also missing? I haven't fully compared the contexts

dqkqd:
## Description

Add new fields `raw_message` in `Commit` to avoid lossy message conversion.

Fix #837 

## Motivation and Context
I implemented according to https://github.com/orhun/git-cliff/issues/837#issuecomment-2404818161

## How Has This Been Tested?

Add 2 test fixtures
- Create changelog with all fields from conventional commits
- Run `git cliff --context | git cliff --from-context` should yield identical result with `git cliff`

## Types of Changes

- [ ] Other 

## Checklist:


git-cliff-core/src/commit.rs
```diff
@@ -423,8 +433,9 @@ impl Serialize for Commit<'_> {
 			}
 		}
 
-		let mut commit = serializer.serialize_struct("Commit", 9)?;
+		let mut commit = serializer.serialize_struct("Commit", 20)?;
```
dqkqd, line 436: I counted there are 20 fields, so I put these as 20. Not sure if it is correct.
I found in `cargo` they use separated struct for serializing. Maybe we should do that as well?
https://github.com/rust-lang/cargo/blob/82c489f1c612096c2dffc48a4091d21af4b8e046/src/cargo/core/manifest.rs#L324

orhun: Yup, 20 should be correct.

>  they use separated struct for serializing.

I'm not sure. If you think it will simplify the code, yes. But the current serialization is already super straightforward :)

What would be the benefit of using a separate struct?

git-cliff-core/src/commit.rs
```diff
@@ -476,6 +487,24 @@ impl Serialize for Commit<'_> {
 	}
 }
 
+/// Deserialize commits into conventional commits if they are convertible
```
orhun, line 490: Nit:
```suggestion
/// Deserialize commits into conventional commits if they are convertible.
```

.github/workflows/test-fixtures.yml
```diff
@@ -93,6 +94,8 @@ jobs:
             command: --bump --unreleased --with-tag-message "Some text"
           - fixtures-name: test-from-context
             command: --from-context context.json
+          - fixtures-name: test-from-context-does-not-discard-fields
+            command: --context | git cliff --from-context -
```
orhun, line 98: I like this, but I think here we are not comparing the final output against the `git cliff --context`, instead we are asserting what we get from `--from-context`.

It would be better if we add a `context.json` file to the fixture and use that instead of running `--context`. What do you think?

orhun (commented):
Neat!

Left some comments :) Also, can you update the documentation (in `/website`) about the existence of the new `raw_message` field?

.github/workflows/test-fixtures.yml
dqkqd: > It would be better if we add a context.json file to the fixture and use that instead of running --context. What do you think?

Yes that's great, it would make the test case deterministic instead of depending on output from `--context`. I will fix that.

git-cliff-core/src/commit.rs
dqkqd: > What would be the benefit of using a separate struct?

Actually, I'm pretty new to serde so I'm not quite sure.
But, I think we don't need to manually change the number of fields when using separated struct.
We can also use proc macro directly, which is more friendly.

But those fields aren't change quite often, so I think just leave them there is also good.

git-cliff-core/src/commit.rs
dqkqd: > What would be the benefit of using a separate struct?

Actually, I'm pretty new to serde so I'm not quite sure.
But, I think we don't need to manually change the number of fields when using separated struct.
We can also use proc macro directly, which is more friendly.

But those fields aren't change quite often, so I think just leave them there is also good.

dqkqd:
@orhun

I fixed according to your comments.
I also added the new field to [context.md](https://github.com/orhun/git-cliff/blob/main/website/docs/templating/context.md), but I'm not sure is there any other places I should look into.

.github/fixtures/test-from-context-does-not-discard-fields/commit.sh
orhun, line 1: These test commits should no longer be needed. Can you test with an empty `commit.sh` and commit it if it works?

P.S. we should do the same for `test-from-context` fixture too.

check.md
orhun, line 1: Accidentally committed? 🙂

orhun (commented):
One last thing :)

.github/fixtures/test-from-context-does-not-discard-fields/commit.sh
dqkqd: Sorry, it didn't work. I got an error, we must at least have one commit. I think this is a bug, but maybe we should fix this in another PR?

```bash
ERROR git_cliff > Git error: `reference 'refs/heads/main' not found; class=Reference (4); code=UnbornBranch (-9)`
```

However, even if we don't use `commit.sh`, I think we should just keep it there, as a reference. I find it more difficult to read `context.json` compared to `commit.sh`

check.md
dqkqd: My bad :(, removed it

.github/fixtures/test-from-context-does-not-discard-fields/commit.sh
orhun: Ah, okay. Yup, we need to fix that bug in another PR.

> I find it more difficult to read context.json compared to commit.sh

Fair enough!

orhun (approved):
This is great, thanks for your contribution!

janbuchar:
Hi @orhun, I just encountered this issue. Could you please make a release that includes this fix?

orhun:
Yup, I will try to do a release next week.

janbuchar:
> Yup, I will try to do a release next week.

Sorry to bother you, but any chance you could release it?

orhun:
released now :)

PR #925, opened 2024-10-19
fix(remote): preserve first time contributors

### LINKED ISSUE #891: Fix the first contributor filter

### Description of the bug

In the following scenario:

1. User foo contributes for the first time.
2. New release is created and changelog is generated.
3. User foo contributes after the release.
4. New release is created and changelog is generated BUT user foo is removed from the first contributors section from the previous release.

Might be related to #825 

### Steps To Reproduce

See above.

### Expected behavior

This happens due to **not checking** the dates of the new commits while updating the release metadata.

https://github.com/orhun/git-cliff/blob/8d10edb7450aaf189fbce5f78a72274739f73ba9/git-cliff-core/src/remote/mod.rs#L337-L347

Here, we check all the commits, but you see, in the new release user foo already contributed, so his contribution is not marked as first time anymore.

To fix it, we should simply:

```patch
diff --git a/git-cliff-core/src/remote/mod.rs b/git-cliff-core/src/remote/mod.rs
index 77d14e7..32e61a9 100644
--- a/git-cliff-core/src/remote/mod.rs
+++ b/git-cliff-core/src/remote/mod.rs
@@ -83,6 +83,8 @@ pub trait RemoteCommit: DynClone {
 	fn id(&self) -> String;
 	/// Commit author.
 	fn username(&self) -> Option<String>;
+	/// Timestamp.
+	fn timestamp(&self) -> Option<i64>;
 }
 
 dyn_clone::clone_trait_object!(RemoteCommit);
@@ -340,6 +342,9 @@ macro_rules! update_release_metadata {
 					.map(|mut v| {
 						v.is_first_time = !commits
 							.iter()
+							.filter(|commit| {
+								commit.timestamp() < Some(self.timestamp)
+							})
 							.map(|v| v.username())
 							.any(|login| login == v.username);
 						v
```

Apply the patch above and you will see that we are missing the implementation for GitHub, GitLab, etc. We need to check each remote and add the appropriate field for serializing the timestamp from their API response.

### Screenshots / Logs

nA

### Software information

nA

### Additional context

_No response_

mcwarman:
## Description

Preserve first time contributors in older change logs

Fixes #891 

## Motivation and Context

When a new release is created and all change logs regenerated the first time contributor section is removed.

## How Has This Been Tested?

Tested using existing fixtures.

## Types of Changes

- [ ] Other 

## Checklist:


mcwarman:
> My code follows the code style of this project.

I'm pretty very new to rust and I'm not entirely sure what I've done around date parsing is correct.

mcwarman:
I'll look at test coverage as well.

orhun:
Thanks for the PR! I will be reviewing this soon. No need to worry about test coverage or the code style for now.

Can you also update `CHANGELOG.md` file at the root of the repository by running `git-cliff --latest` with these changes? I want to see if the previously deleted first-time contributors will come back :)

mcwarman:
Updated the `CHANGELOG.md`, I had to run `cargo run -- -o CHANGELOG.md` and remove `[unreleased]` section manually.

 `--latest` only return latest tag.

git-cliff-core/Cargo.toml
```diff
@@ -77,6 +77,7 @@ url.workspace = true
 dyn-clone = "1.0.17"
 urlencoding = "2.1.3"
 cacache = { version = "13.0.0", features = ["mmap"], default-features = false }
+chrono = "0.4.38"
```
orhun, line 80: Can we use the [`time`](https://docs.rs/time/) crate directly instead of chrono? If I understood correctly, it is only used for converting a Rfc3339 string to a timestamp. `time` also [supports that](https://time-rs.github.io/api/time/format_description/well_known/struct.Rfc3339.html) and it should be more lightweight.

See: https://stackoverflow.com/questions/74935683/convert-utc-rfc3339-timestamp-to-local-time-with-the-time-crate

Thoughts?

orhun (commented):
Nice! The behavior seems to be fixed 👌🏼 

I have one little comment.

orhun (approved):
Thanks for the PR! ❤️

PR #930, opened 2024-10-23
feat: jujustu support

### LINKED ISSUE #875: Add support for Jujutsu (when using git store)

### Feature description

Jujutsu is a version control system very similar to Git. It currently uses Git as it's local store. It can work with Git in one of two ways. One is 'co-located', where the `.git` repo is in it's normal location. The other stores the `.git/` files in `.jj/repo/store/git/`. This second is the default when the user clones a Git repo without explicitly requesting that is be co-located.

I would like to be able to use `git-cliff` with a repo that I have checked out locally using the default non-co-located configuration.

### Desired solution

Detecting the alternate location of the `.git/` files should be straight forward. When the `.git/` directory is not found, simply retry using the `.jj/repo/store/git/` directory instead.

Jujutsu is branchless, so there may not be any local branches. There will still be references to the remote branches, e.g. `origin/main`, etc. e.g. `.jj/repo/store/git/refs/remotes/origin/main`.

So, when using the Jujutsu location for the git repo, using the remote branches rather than local ones may be required. (It is possible to have local branches, but this is up to the users discretion/workflow).

### Alternatives considered

Only clone repos with Jujutsu using the co-located option. This is an easy solution with no real downside other than the need to remember to pass the required parameter when cloning.

### Additional context

I recognise that `git-cliff` is aimed at Git compatibility (it's right there in the name!). Jujutsu uses the same on-disk store, so support shouldn't be _too_ difficult to implement. (Then again there's the other 80% of the problem I can't see).

kemitix:
## Description

Adds support for opening the Git repository used when a repository is cloned using Jujutsu and the `--colocate` option was *NOT* used.

Note: this requires that the user updates the `HEAD` of the git repository to point to their mainline branch, as the Git repository the Jujutsu creates isn't 100% configured.

If the .git directory can't be opened (usually because the `.git` directory is missing) it attempts to find the git repository in the location used by Jujutsu (i.e. `.jj/repo/store/git`). If the Jujutsu git directory doesn't exist, then the origin error is propogated.

## Motivation and Context

Jujutsu is a Version Control System that uses Git as it's local store, and can work as an effective replacement for Git on a Developer's workstation.

When a Git repository is cloned by Jujutsu, the default location for the Git respository files is `.jj/repo/store/git`, rather than the usual `.git`. `git-cliff` expectes the `.git` location. This PR teaches `git-cliff` to also check the `.jj/repo/store/git` location.

Closes #875

### Test with Jujutsu repository:

- Cloned a Git repository using Jujutsu: `jj git clone git@git.kemitix.net:kemitix/git-next.git && cd git-next`
- Update the HEAD to point to the remote `main` branch: `echo 'ref: refs/remotes/origin/main`
- Run `git-cliff`: `../git-cliff/target/debug/git-cliff`

### Regression test with a Git repository:

- Cloned a Git repository using Jujutsu: `git clone git@git.kemitix.net:kemitix/git-next.git git-next-git && cd git-next-git`
- Run `git-cliff`: `../git-cliff/target/debug/git-cliff`

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/git-cliff-core/src/repo.rs
+++ b/git-cliff-core/src/repo.rs
@@ -48,7 +48,15 @@ impl Repository {
 	/// Initializes (opens) the repository.
 	pub fn init(path: PathBuf) -> Result<Self> {
 		if path.exists() {
-			let inner = GitRepository::open(&path)?;
+			let inner = GitRepository::open(&path).or_else(|err| {
+				let jujutsu_path =
+					path.join(".jj").join("repo").join("store").join("git");
+				if jujutsu_path.exists() {
+					GitRepository::open_bare(&jujutsu_path)
+				} else {
+					Err(err)
+				}
+			})?;
 			let changed_files_cache_path = inner
 				.path()
 				.join(env!("CARGO_PKG_NAME"))
@@ -503,9 +511,12 @@ fn ssh_path_segments(url: &str) -> Result<Remote> {
 mod test {
 	use super::*;
 	use crate::commit::Commit as AppCommit;
-	use std::env;
 	use std::process::Command;
 	use std::str;
+	use std::{
+		env,
+		fs,
+	};
 	use temp_dir::TempDir;
 
 	fn get_last_commit_hash() -> Result<String> {
@@ -733,6 +744,62 @@ mod test {
 		);
 
 		(repo, temp_dir)
+	}
+
+	#[test]
+	fn open_jujutsu_repo() {
+		let (repo, _temp_dir) = create_temp_repo();
+		// working copy is the directory that contains the .git directory:
+		let working_copy = repo.path;
+
+		// Make the Git repository bare and set HEAD
+		std::process::Command::new("git")
+			.args(["config", "core.bare", "true"])
+			.current_dir(&working_copy)
+			.status()
+			.expect("failed to make git repo non-bare");
+		// Move the Git repo into jj
+		let store = working_copy.join(".jj").join("repo").join("store");
+		fs::create_dir_all(&store).expect("failed to create dir");
+		fs::rename(working_copy.join(".git"), store.join("git"))
+			.expect("failed to move git repo");
+
+		// Open repo from working copy, that contains the .jj directory
+		let repo = Repository::init(working_copy).expect("failed to init repo");
+
+		// macOS canonical path for temp directories is in /private
+		// libgit2 forces the path to be canonical regardless of what we pass in
+		if repo.inner.path().starts_with("/private") {
+			assert_eq!(
+				repo.inner.path().strip_prefix("/private"),
+				store.join("git").strip_prefix("/"),
+				"open git repo in .jj/repo/store/"
+			);
+		} else {
+			assert_eq!(
+				repo.inner.path(),
+				store.join("git"),
+				"open git repo in .jj/repo/store/"
+			);
+		}
+	}
+
+	#[test]
+	fn propagate_error_if_no_repo_found() {
+		let temp_dir =
+			TempDir::with_prefix("git-cliff-").expect("failed to create temp dir");
+
+		let path = temp_dir.path().to_path_buf();
+
+		let result = Repository::init(path.clone());
+
+		assert!(result.is_err());
+		if let Err(error) = result {
+			assert!(format!("{error:?}").contains(
+				format!("could not find repository at '{}'", path.display())
+					.as_str()
+			))
+		}
 	}
 
 	fn create_commit_with_files<'a>(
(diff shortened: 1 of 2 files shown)


website/docs/usage/jujutsu.md
```diff
@@ -0,0 +1,39 @@
...
+```bash
+echo "gitdir: .jj/repo/store/git" > .git
+```
+
+### Remote branches
+
+Update the `HEAD` of the Git repository to point to your remote mainline branch.
+
+```bash
+echo "ref: refs/remotes/origin/main"
```
orhun, line 35: Isn't this missing a redirection to the `.jj/repo/store/git/HEAD` file?

orhun (commented):
This is a good starting point, thanks :)

website/docs/usage/jujutsu.md
```diff
@@ -0,0 +1,39 @@
...
+
+You can do one of two things to make your repository compatible with `git-cliff`:
+
+### Fake Colocation
+
+Create a file in the root of your repository that tells Git, and `git-cliff` where the Git repository is:
+
+```bash
+echo "gitdir: .jj/repo/store/git" > .git
+```
```
orhun, line 28: I tried this for a cloned repository but it didn't work for me:

```
$ jj git clone https://github.com/orhun/menyoki
$ echo "gitdir: .jj/repo/store/git" > .git

# and then run git-cliff built from this branch
#     Running `/drv/cargo/target/debug/git-cliff -r /tmp/tmp.sP4i0aaZ62/menyoki -c config/cliff.toml`
 ERROR git_cliff > Git error: `reference 'refs/heads/master' not found; class=Reference (4); code=UnbornBranch (-9)`
```

git-cliff-core/src/repo.rs
```diff
@@ -735,6 +746,62 @@ mod test {
 		(repo, temp_dir)
 	}
 
+	#[test]
+	fn open_jujutsu_repo() {
+		let (repo, _temp_dir) = create_temp_repo();
+		// working copy is the directory that contains the .git directory:
+		let working_copy = repo.path;
+
+		// # Make the Git repository bare and set HEAD
```
orhun, line 755:
```suggestion
		// Make the Git repository bare and set HEAD
```

git-cliff-core/src/repo.rs
```diff
@@ -735,6 +746,62 @@ mod test {
...
+		std::process::Command::new("git")
+			.args(["config", "core.bare", "true"])
+			.current_dir(&working_copy)
+			.status()
+			.expect("failed to make git repo non-bare");
+		// # Move the Git repo into jj
```
orhun, line 761:
```suggestion
		// Move the Git repo into jj
```

git-cliff-core/src/repo.rs
```diff
@@ -735,6 +746,62 @@ mod test {
...
+		let store = working_copy.join(".jj").join("repo").join("store");
+		fs::create_dir_all(&store).expect("failed to create dir");
+		fs::rename(working_copy.join(".git"), store.join("git"))
+			.expect("failed to move git repo");
+
+		// Open repo from working copy, that contains the .jj directory
+		let repo = Repository::init(working_copy).expect("failed to init repo");
+
+		// macos canonical path for temp directories is in /private
```
orhun, line 770:
```suggestion
		// macOS canonical path for temp directories is in /private
```

git-cliff-core/src/repo.rs
```diff
@@ -735,6 +746,62 @@ mod test {
...
+			assert_eq!(
+				repo.inner.path(),
+				store.join("git"),
+				"open git repo in .jj/repo/store/"
+			);
+		}
+	}
+
+	#[test]
+	fn propagate_original_error_when_no_jujutsu_repo() {
```
orhun, line 788:
```suggestion
	fn propagate_error_if_no_repo_found() {
```

website/docs/usage/jujutsu.md
kemitix: Okay. @orhun: how about this:

```bash
jj git clone https://github.com/orhun/menyoki
cd menyoki
echo "gitdir: .jj/repo/store/git" > .git
echo "ref: refs/remotes/origin/master" > .jj/repo/store/git/HEAD
```

2024-10-29T07-48-04Z_code

orhun: Yup, that works. In that case we shouldn't present it as "you need to do one of these two options" in the documentation then.

kemitix: lol yeah. I've merged them.

orhun (approved):
Awesome, thanks!

PR #937, opened 2024-10-27
fix: don't create regex inside a loop

kemitix:
## Description

Moves the creation of an array of `TextProcessor` objects, that includes the contruction of a `Regex` to the outside of a loop. The array is only ever user by reference, so reusing a single instance will produce the same effects.

## Motivation and Context

The Lints CI check is failing due to the creation of the same Regex within a loop.

## How Has This Been Tested?

The change is within the integration tests, and those continue to pass.

image

## Types of Changes

- [ ] Other 

## Checklist:

- [ ] ~I have updated the documentation accordingly.~
- [ ] ~I have added tests to cover my changes.~


orhun (approved):
Thanks!

PR #918, opened 2024-10-10
feat(nix): add a basic Nix environment

blackheaven:
## Description

I add a basic Nix environment.

## Motivation and Context

Provide a reproducible dev environment.

## How Has This Been Tested?

I've run it on my NixOS.

## Types of Changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation (no code change)
- [ ] Refactor (refactoring production code)
- [x] Other 

Add a new dev tool.

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] I have updated the documentation accordingly.
- [ ] I have formatted the code with [rustfmt](https://github.com/rust-lang/rustfmt).
- [ ] I checked the lints with [clippy](https://github.com/rust-lang/rust-clippy).
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #939, opened 2024-10-29
feat: add Nix CI

blackheaven:
## Description

Add full Nix build and associated CI.

## Motivation and Context

Following [my previous Nix PR](https://github.com/orhun/git-cliff/pull/918#pullrequestreview-2372038928), I was suggested to add the CI part.

It had took a full Nix build support (instead of just an environment definition.

## How Has This Been Tested?

Locally via `nix flake check -Lv`.

## Types of Changes

- [x] Other 

## Checklist:


flake.nix
```diff
@@ -15,7 +15,40 @@
...
+            "--skip=run_os_command"
+            "--skip=git_tags"
+            "--skip=git_upstream_remote"
+            "--skip=includes_root_commit"
+            "--skip=resolves_existing_tag_with_name_and_message"
+            "--skip=resolves_tag_when_no_tags_exist"
+            "--skip=test_should_retain_commit"
+            "--skip=commit_search"
+            "--skip=get_latest_commit"
+            "--skip=get_latest_tag"
```
orhun, line 36: Why are we skipping these tests?

blackheaven: IIUC, they run filesystems/git-based tests.

I can't investigate if you want.

orhun: Yes, would be nice to run them as a part of the Nix build :)

blackheaven: You can't really do that out-of-the-box.

Nix flake drops `.git` directory at build time.

A solution which could be used would be to create a (dynamically or not) a git repository for the test (we could reuse `git-cliff`, but GitHub does not allow you to download and archive with the full history).

WDYT?

orhun: I see, then in that case let's skip them via `--skip "repo::test"` (instead of individually skipping them) which is the same thing that we do for the [Arch package](https://gitlab.archlinux.org/archlinux/packaging/packages/git-cliff/-/blob/main/PKGBUILD?ref_type=heads#L35).

blackheaven: done!

I won't lie, it was challenging.

git-cliff-core/Cargo.toml
```diff
@@ -78,6 +78,7 @@ dyn-clone = "1.0.17"
 urlencoding = "2.1.3"
 cacache = { version = "13.0.0", features = ["mmap"], default-features = false }
 time = "0.3.36"
+test-tag = "0.1.4"
```
orhun, line 81: Why this is needed?

flake.nix
orhun: Hmm, I'm a bit confused. Why are we adding a new rust flag instead of just `cargo test -- --skip repo::test --skip command`?

git-cliff-core/Cargo.toml
blackheaven: It's not, just a leftover of a previous attempt, thanks.

flake.nix
blackheaven: ah, I thought you wanted something more fine coarse.

orhun: This is also fine, but the `--skip` approach would be much simpler I think :)

orhun (approved):
Awesome, thank you!

blackheaven:
Thanks for your guidance!

PR #1058, opened 2025-02-26
DRAFT: submodule recursion

yudjinn:
## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/git-cliff-core/src/commit.rs
+++ b/git-cliff-core/src/commit.rs
@@ -8,6 +8,8 @@ use crate::error::{
 	Error as AppError,
 	Result,
 };
+use crate::repo::Repository;
+use git2::Submodule;
 #[cfg(feature = "repo")]
 use git2::{
 	Commit as GitCommit,
@@ -248,6 +250,29 @@ impl Commit<'_> {
 			}
 			Err(e) => Err(AppError::ParseError(e)),
 		}
+	}
+
+	/// Returns wether the commit changes the SHA of a submodule
+	pub fn contains_submodule_update<'a>(
+		&self,
+		repo: &'a Repository,
+	) -> Option<Submodule<'a>> {
+		let commits = match repo.commits(Some(&self.id.to_string()), None, None) {
+			Ok(it) => it,
+			Err(_) => return None,
+		};
+		let output = repo.commit_changed_files_no_cache(&commits[0]);
+		for o in &output {
+			for r in match repo.submodules() {
+				Ok(it) => it,
+				Err(_) => return None,
+			} {
+				if r.path().to_path_buf() == *o {
+					return Some(r);
+				}
+			}
+		}
+		None
 	}
 
 	/// Preprocesses the commit using [`TextProcessor`]s.
--- a/git-cliff/src/lib.rs
+++ b/git-cliff/src/lib.rs
@@ -35,12 +35,18 @@ use git_cliff_core::error::{
 	Result,
 };
 use git_cliff_core::release::Release;
-use git_cliff_core::repo::Repository;
+use git_cliff_core::repo::{
+	GitSubmodule,
+	Repository,
+};
+use git_cliff_core::tag::Tag;
 use git_cliff_core::{
 	DEFAULT_CONFIG,
 	IGNORE_FILE,
 };
 use glob::Pattern;
+use indexmap::IndexMap;
+use std::collections::HashMap;
 use std::env;
 use std::fs::{
 	self,
@@ -95,6 +101,8 @@ fn process_repository<'a>(
 	let skip_regex = config.git.skip_tags.as_ref();
 	let ignore_regex = config.git.ignore_tags.as_ref();
 	let count_tags = config.git.count_tags.as_ref();
+	let recurse_submodules =
+		*config.git.recurse_submodules.as_ref().unwrap_or(&false);
 	tags.retain(|_, tag| {
 		let name = &tag.name;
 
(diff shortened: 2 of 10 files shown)


git-cliff-core/src/commit.rs
```diff
@@ -250,6 +252,29 @@ impl Commit<'_> {
 		}
 	}
 
+	/// Returns wether the commit changes the SHA of a submodule
```
orhun, line 255:
```suggestion
	/// Returns whether the commit changes the SHA of a submodule
```

git-cliff/src/lib.rs
```diff
@@ -167,56 +175,7 @@ fn process_repository<'a>(
...
-			}
-			if let (Some(tag1), Some(tag2)) = (
-				tags.get_index(tag_index).map(|(k, _)| k),
-				tags.get_index(tag_index + 1).map(|(k, _)| k),
-			) {
-				commit_range = Some(format!("{tag1}..{tag2}"));
-			}
-		}
-	}
+	parse_commits(args, &tags, &mut commit_range, repository)?;
```
orhun, line 178: Why this has been extracted to a new function?

git-cliff/src/lib.rs
```diff
@@ -303,6 +279,36 @@ fn process_repository<'a>(
 			releases.push(Release::default());
 		}
 	}
+	if recurse_submodules {
```
orhun, line 282: I guess we are updating the `releases` vector here based on the commits in the submodules.

It would be nice to extract this to a function that takes `&mut releases` maybe.

git-cliff/src/lib.rs
```diff
@@ -303,6 +279,36 @@ fn process_repository<'a>(
...
+		for (_submodule, _commits) in submodule_commits.values() {
+			log::trace!("Recursing {:?}.", _submodule.path().to_str());
+			let _range = format!(
```
orhun, line 285: This doesn't seem to be used.

git-cliff/src/lib.rs
```diff
@@ -303,6 +279,36 @@ fn process_repository<'a>(
...
+				"{}..{}",
+				_commits.first().unwrap().id,
+				_commits.last().unwrap().id
+			);
+			let _repo_path = _submodule.path().to_path_buf();
+			let _repo = Repository::init(_repo_path.clone())?;
```
orhun, line 291: You can get rid of the compiler error by doing this for now:

```rust
let repo = Box::leak(Box::new(Repository::init(repo_path.clone())?));
```

It's a bit bad, but this is also what I'm doing for the other repositories 💀

orhun (commented):
Good stuff, I left some comments that I think help moving further.

Tested it quickly, but didn't see this in action yet. Can you also check again?

git-cliff/src/lib.rs
yudjinn: interim derp, thought I might reuse it but I dont need to 😓

lehmanju:
Oh yes, great, that's a feature I am very interested in. I volunteer for testing, we use this at work for changelog generation with about 200 submodules. Right now we use the repository flag to generate a merged changelog for each submodule. 
However, version 2.8 did somehow break it so I am looking forward to true submodule support.

There is one particular issue on CI environments I ran into. If you don't specify `fetch-depth: 0` in Github Actions checkout then the submodule history is completely missing and it only contains the last commit. Maybe there is room for further improvement to fetch the necessary submodule commits on demand without fetching the complete history/refs.

orhun:
It's great to hear there is some interest for this feature :)  

> However, version 2.8 did somehow break it so I am looking forward to true submodule support.

Ah, it's probably related to the monorepo related changes. Would love to take a look into it if you have time to submit an issue about this. But I agree that proper submodule support would be indeed better here.

>  Maybe there is room for further improvement to fetch the necessary submodule commits on demand without fetching the complete history/refs.

Interesting idea...

lehmanju:
Couldn't help myself and started tinkering with this code ;). Here are some thoughts:

- the idea does seem to be: 
  - *iterate over all commits and save commits that change a submodule*
  - *gather all submodule commits for these toplevel commits and append them to a release*
- this does not work because:
  - only toplevel commit ids are gathered, but to iterate through submodule commits submodule commit ids need to be known. this requires reading the diff content, maybe `commit_changed_submodules_no_cache` .
  - instead of doing these two steps separately, recursive commit explosion can be done directly while iterating through toplevel commits. submodule commits are then just appended to the same release tag/repository with an additional `submodule`  value (optional)
  - a template can then process a release grouping/filtering commits based on submodule.

orhun:
Interesting... I didn't take a deeper look into this code but your take makes sense. I think we can wait for @yudjinn for incorporate some changes into this PR or feel free to put up a new draft based on those.

yudjinn:
I poked at it some more, but in short, @lehmanju is correct that we do need to get submodule changes as well. I am currently planning on doing that as a Vec based on the new/old filename that the commit sees, and then just diffing releases based on the first and last in that vec, but if you have a better idea to do that explosion earlier in one shot, I'd love to see it

orhun:
> And since this line already iterates through the diff, I'd extend it to look into the file contents to get the submodule commit id.
> However, I'm stuck at how to ask libgit2 to get me the file contents if I have a diff?

Not sure how it would be possible to get the submodule information from a `Diff` but it is possible to get the file contents via calling [`Diff::print`](https://docs.rs/git2/latest/git2/struct.Diff.html#method.print). Here you can also filter by `DiffDelta` (and `DiffFile`).

Here is an example from `git2`: https://github.com/rust-lang/git2-rs/blob/d1ae3b6c2d1200e7d82468af447fa66259225ecf/examples/log.rs#L201-L209

git-cliff/src/lib.rs
```diff
@@ -273,10 +281,34 @@ fn process_repository<'a>(
 	let mut previous_release = Release::default();
 	let mut first_processed_tag = None;
 	let repository_path = repository.path()?.to_string_lossy().into_owned();
+	let mut submodule_commits: HashMap<PathBuf, (&mut Repository, Vec<Commit>)> =
```
orhun, line 284: Why do we have this complex type? 🤔

yudjinn: So we don't have to reinit The repo every time we want to parse commits, but tbh I put some of that together after a very full day so I plan on redoing it a bit. The pathbuf being the key is just to avoid having to impl comparisons for submodules

orhun: Got it. it would be nice to refactor/simplify this a bit for sure :)

yudjinn: I think I might just swap it to be a tuple and replace the end commit instead. Plan is that for any given submodule, the releases we include should be inclusive and in aggregate. So if a single release has multiple commits bumping the submodule underneath it, just aggregate all of those changes

yudjinn:
> I'd probably explode the commits between submodule _before_ and _after_ states somewhere in:
> 
> https://github.com/orhun/git-cliff/blob/63129ce0c481c00c06bc9ee6271da76e99a6e933/git-cliff-core/src/repo.rs#L116-L150
> 
> 
> https://github.com/orhun/git-cliff/blob/63129ce0c481c00c06bc9ee6271da76e99a6e933/git-cliff-core/src/repo.rs#L275-L285
> 
> 
> And since this line already iterates through the diff, I'd extend it to look into the file contents to get the submodule commit id.
> However, I'm stuck at _how to ask libgit2 to get me the file contents if I have a diff_? [DiffFile](https://docs.rs/git2/latest/git2/struct.DiffFile.html) doesn't have a getter to get the contents.

the only way I was able to do it was checking the `old_path` and `new_path` against the paths from `repo.submodules()`

orhun:
Now implemented in #1082

PR #1061, opened 2025-02-27
feature: add option git.require-conventional

### LINKED ISSUE #536: add option to fail on unconventional commits

### Feature description

An option to fail generating the changelog if the git history contains commits that do not comply to the conventional commits format would allow using git-cliff as a quality gate. Currently additional tools are needed to implement such behaviour.

### Desired solution

Add a new configuration parameter:
```toml
# fail if the git history contains commits that are not conventional (default: false)
fail_on_unconventional = true
```

If git-cliff encounters any unconventional commits, it should print the hashes of all unconventional commits to stderr and exit with a non-zero return code. Outputting the commits gives developers a clear indication as to what they need to fix.
```sh
$ git-cliff
Encountered unconventional commits: 12345678, abcdefg
```

### Alternatives considered

There exist other tools that can check compliance with conventional commits. However having to use multiple tools bloats the pipeline.

### Additional context

_No response_

Cyclonit:
## Description

closes #536

This PR adds a new option `git.require-conventional`, which requires that all commits in the changelog are conventional. If the history contains any unconventional commits, an error message listing the offending commits is emitted and the changelog generation fails.

The new option takes precedence over `git.filter-unconventional`.

Example output:
```
ERROR git_cliff_core::changelog > The following commits are not conventional: ["b5da60869701cb51f77282fa426a4d458152b02a", "2a9b8746b6398f287e6d44b3a2343e1aecfe9252", "b6d90f01839716df7402f51c1ff0b65c224d099f"]
```

## Motivation and Context

I am looking for a way to enforce commit guidelines prior to merging contributions. This serves to maintain a satisfactory level of quality in the git history and subsequently the changelog.

## How Has This Been Tested?

Added new test fixtures.

## Types of Changes

- [ ] Other 

## Checklist:


git-cliff-core/src/changelog.rs
```diff
@@ -144,6 +144,27 @@ impl<'a> Changelog<'a> {
...
 				.collect::<Vec<Commit>>();
 		});
+
+		if self.config.git.require_conventional.unwrap_or(false) {
+			debug!("Verifying that all commits are conventional.");
+			let mut unconventional_ids = Vec::new();
+
+			self.releases.iter().for_each(|release| {
+				release.commits.iter().for_each(|commit| {
+					println!("{:?}", commit.conv);
```
orhun, line 154: for debugging purposes?

orhun:
Hey, this looks good, going towards the right direction. I'd throw the error in `git-cliff` binary instead of `git-cliff-core`.

git-cliff-core/src/changelog.rs
Cyclonit: yes

Cyclonit:
So, I think this is ready for a proper review. :)

> Hey, this looks good, going towards the right direction. I'd throw the error in `git-cliff` binary instead of `git-cliff-core`.

I reworked how the logging and error work. Is this more to your liking? I am not sure how I would go about throwing the error in `git-cliff` instead. Within `changelog.rs`, `changelog.process_commits()` may return an error now, just like `changelog.add_remote_data()` does.

https://github.com/orhun/git-cliff/pull/1061/files#diff-1c527f230d9fe61c0ffca65b8c03fe724a07753b8caeea26c05271e2688a9813R47-R53

git-cliff-core/src/error.rs
```diff
@@ -111,6 +111,12 @@ pub enum Error {
 	/// Error that may occur while constructing patterns.
 	#[error("Pattern error: `{0}`")]
 	PatternError(#[from] glob::PatternError),
+	/// Error that
```
orhun, line 114: We should complete this doc comment, maybe sth like this:

```suggestion
	/// Error that may occur when unconventional commits are found.
	/// See `require_conventional` option for more information.
```

Or anything that fits to your liking.

git-cliff-core/src/changelog.rs
```diff
@@ -144,6 +147,26 @@ impl<'a> Changelog<'a> {
...
 		});
+
+		if self.config.git.require_conventional.unwrap_or(false) {
+			debug!("Verifying that all commits are conventional.");
+			let mut unconventional_count = 0;
+
+			self.releases.iter().for_each(|release| {
+				release.commits.iter().for_each(|commit| {
+					if commit.conv.is_none() {
+						error!("Commit {} is not conventional.", &commit.id[..7]);
```
orhun, line 158: How about we also log the commit message here?

```suggestion
						error!(
							"Commit {id} is not conventional. ({message})",
							id = &commit.id[..7],
							message = commit
								.message
								.lines()
								.next()
								.unwrap_or_default()
								.trim()
						);
```

orhun (commented):
LGTM, added some minor comments.

Can you also add this entry to the defaul changelog `config/cliff.toml` so that it is easier to use when you do `--init`? (It should be `false` as default)

git-cliff-core/src/changelog.rs
Cyclonit: I was considering it, but I was unsure how this ends up looking in the log. Given the following commit message, only the first line ends up being logged:

```txt
This is unconventional on multiple lines.

Line 2

Some Footer: user3 <user3@example.com>
```

```txt
Commit ffdc50e is not conventional. (This is unconventional on multiple lines.)
```

Would it make sense to log the message on separate lines underneath instead?

Cyclonit: Having multi-line error messages is bad for automated processing of log messages, but I doubt that would be a problem for git-cliff, because it is a pure CI tool.

orhun: Yup, then we can do format it like that when there is a multi line commit and print everything in a single line when it's a single line commit.

Alternatively I was thinking to add `colored` to make it prettier (e.g. make the commit ID orange etc,), but I'm not sure.

Cyclonit: I changedit to the new format for both single- and multi-line commits. I think this is fine and easyon the eyes.

```txt
ERROR git_cliff_core::changelog > Commit df5e6aa is not conventional:
    | This is unconventional on a single line.
ERROR git_cliff_core::changelog > Commit ffdc50e is not conventional:
    | This is unconventional on multiple lines.
    |
    | Line 2
    |
    | Some Footer: user3 <user3@example.com>
```

git-cliff-core/src/changelog.rs
```diff
@@ -144,6 +147,41 @@ impl<'a> Changelog<'a> {
 				})
 				.collect::<Vec<Commit>>();
 		});
+
+		self.check_require_conventional()?;
```
orhun, line 151: We should only call this when `require_conventional` is set to true.

git-cliff-core/src/changelog.rs
```diff
@@ -144,6 +147,41 @@ impl<'a> Changelog<'a> {
...
+
+		Ok(())
+	}
+
+	fn check_require_conventional(&mut self) -> Result<()> {
```
orhun, line 156:
```suggestion
    /// Checks the commits and returns an error if any unconventional commits are found.
	fn check_conventional_commits(&mut self) -> Result<()> {
```

.github/actions/run-fixtures-test/action.yml
```diff
@@ -39,16 +39,32 @@ runs:
...
+          current_date=$(date +"$IN_DATE_FORMAT")
+          sed -i "s/<<DATE>>/$current_date/g" expected.md
+          if [ ! diff --strip-trailing-cr output.md expected.md ]; then
+            echo "The generated changelog does not match the expected changelog."
+            exit 1
+          fi
+        fi
+      shell: bash
+
+    - name: Compare the return code with the expected return code
```
orhun, line 63: I'm thinking if we should support comparing the output log as well. It would really help to see what those error messages look like when unconventional commits are found.

Cyclonit: Good point, but somewhatt more complicated I assume. I wouldn't diff the entire log with some expected log, but only some messages.

Cyclonit: I have played around with this a bit and I couldn't find a clean solution. Matching on the entire output is too much. This would lead to random test failures whenever any log message is changed. The next step would be to have a file with expected log messages. But that too introduces problems. If a line contains any variable at all (e.g. the commit id), it will not match the expected output unless we do some magic string replacement first.

While I agree that this would be a nice thing to have, I'd opt for not tackling it in this PR.

orhun: Good call. Created #1075.

git-cliff-core/src/changelog.rs
```diff
@@ -144,6 +147,43 @@ impl<'a> Changelog<'a> {
...
+					error!(
+						"Commit {id} is not conventional:\n{message}",
+						id = &commit.id[..7],
+						message = commit
+							.message
+							.lines()
+							.map(|line| { format!("    | {}", line.trim()) })
+							.collect::<Vec<String>>()
+							.join("\n")
+					);
```
orhun, line 176: How about we print in a single line when the commit is not multine?

```suggestion
					error!(
						"Commit {id} is not conventional: {message}",
						id = &commit.id[..7],
						message = {
							if commit.message.lines().count() == 1 {
								format!("{}", commit.message.trim())
							} else {
								commit
									.message
									.lines()
									.map(|line| format!("\n    | {}", line.trim()))
									.collect::<Vec<String>>()
									.join("")
							}
						}
					);
```

Results in:

```
 ERROR git_cliff_core::changelog > Commit 9cba393 is not conventional:
    | Merge pull request #7 from gitolith/feat/custom-error-handling
    |
    | Custom Errors
 ERROR git_cliff_core::changelog > Commit b5f0136 is not conventional: Merge branch 'master'
 ERROR git_cliff_core::changelog > Commit 223cfc9 is not conventional:
    | Merge pull request #9 from orhun/chore/fix-ci
    |
    | Fix CI
 ERROR git_cliff_core::changelog > Commit 74f605d is not conventional: Merge branch 'master' into feat/parse-commits
```

I find this better because we are not wasting newlines for commits that are single line.

What do you think?

orhun (commented):
Needs a rebase on `main` and I left a comment about formatting the output.

git-cliff-core/src/changelog.rs
Cyclonit: I prefer sticking to the multi-line approach. Changing the look of an error message because some messages are shorter than others makes it more difficult to parse at a glance.

Also I doubt many of our users pay by line, thus having a few more lines in the output won't hurt anyone ;)

Cyclonit:
@orhun, unless you insist on logging single line commit messages differently, this should be ready to merge.

orhun (approved):
Great stuff, thanks!

PR #1082, opened 2025-03-07
feat: recurse into submodules

lehmanju:
### Usage

If `recurse_submodules` is set to `true`, you can include submodule commits into your changelog.

For example:

```toml
[git]
recurse_submodules = true

[changelog]
body = """
{% for submodule_path, commits in submodule_commits %}
    ### {{ submodule_path | upper_first }}
    {% for group, commits in commits | group_by(attribute="group") %}
        #### {{ group | upper_first }}
        {% for commit in commits %}
            - {{ commit.message | upper_first }}\
        {% endfor %}
    {% endfor %}
{% endfor %}\n
"""
```

### Implementation details

Adds a feature which allows git-cliff to recurse into submodules of a repository. Between two subsequent releseases in the main repository, all submodule commits in between are expanded. For each submodule, there is an entry in `submodule_commits` mapping the submodules path to a list of commits. This feature is off by default (recurse_submodules=false). Nested submodules are not supported.


yudjinn:
Nice! So for a given commit range of the containing repo, this returns all submodules that changed within those commits, and their deltas between them?

lehmanju:
Well, this iteration sort of actually works! However, the submodule path is not correct ...
[context.json](https://github.com/user-attachments/files/19132065/context.json)

I used openstack/openstack for testing btw, that repository does seem like the ultimate stresstest.

lehmanju:
So that is fixed as well, only config option, documentation and code cleanup/review are missing.

orhun:
This is some good stuff. Let me know when this is ready for review :)

git-cliff-core/Cargo.toml
```diff
@@ -55,7 +55,7 @@ serde_json = "1.0.137"
 bincode = "2.0.0-rc.3"
 serde_regex = "1.1.0"
 tera = "1.20.0"
-indexmap = { version = "2.7.1", optional = true }
+indexmap = { version = "2.7.1", features = ["serde"] }
```
lehmanju, line 58: this isn't optional anymore since `Release` uses it for submodule commits. maybe use hashmap instead? my reasoning was that its used in `Repository` as well, so keep it consistent

git-cliff-core/src/changelog.rs
```diff
@@ -835,6 +836,7 @@ mod test {
 				sort_commits:             Some(String::from("oldest")),
 				link_parsers:             None,
 				limit_commits:            None,
+				recurse_submodules:       None,
```
lehmanju, line 839: not sure whether the name fits, it recurses only with depth 1 but I don't think there is a use case for higher depths.

git-cliff-core/src/commit.rs
```diff
@@ -177,8 +177,8 @@ impl From<String> for Commit<'_> {
 }
 
 #[cfg(feature = "repo")]
-impl<'a> From<&GitCommit<'a>> for Commit<'a> {
-	fn from(commit: &GitCommit<'a>) -> Self {
+impl From<&GitCommit<'_>> for Commit<'_> {
```
lehmanju, line 180: the lifetimes of `GitCommit` and `Commit` are not related

git-cliff-core/src/release.rs
```diff
@@ -31,35 +32,37 @@ use serde_json::value::Value;
...
-	pub timestamp:  i64,
+	pub timestamp:         i64,
 	/// Previous release.
-	pub previous:   Option<Box<Release<'a>>>,
+	pub previous:          Option<Box<Release<'a>>>,
 	/// Repository path.
-	pub repository: Option<String>,
+	pub repository:        Option<String>,
+	/// Submodule commits.
+	pub submodule_commits: Option<IndexMap<String, Vec<Commit<'a>>>>,
```
lehmanju, line 51: Map of relative submodule path to a list of commits.

lehmanju (commented):
not sure about some naming conventions and what should be tested/where.

git-cliff-core/src/release.rs
orhun: Can you add a comment explaining what the `IndexMap` contains? It's not clear to me what that `String` is supposed to be.

git-cliff-core/src/repo.rs
```diff
@@ -90,6 +93,14 @@ impl Repository {
 		Ok(path)
 	}
 
+	/// Returns the initial path of the repository.
+	///
+	/// In case of a submodule this is the relative path to the toplevel
+	/// repository.
+	pub fn initial_path(&self) -> &PathBuf {
```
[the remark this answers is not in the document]
orhun: I recommend the following:

- Rename the existing `path` function to `root_path` since I think that's more fitting.
- Use `path` instead of `initial_path` here.

WDYT?

git-cliff/Cargo.toml
```diff
@@ -60,7 +66,7 @@ url.workspace = true
 pathdiff = "0.2.3"
 
 [dependencies.git-cliff-core]
-version = "2.8.0" # managed by release.sh
+version = "2.8.0"          # managed by release.sh
```
orhun, line 69: Revert these formatting changes pls

git-cliff/src/lib.rs
```diff
@@ -77,6 +78,46 @@ fn check_new_version() {
...
+	let first_commit = release
+		.commits
+		.first()
+		.map(|commit| &commit.id)
+		.and_then(|commit_id| repository.find_commit(commit_id));
+	let last_commit = release
+		.commits
+		.last()
+		.map(|commit| &commit.id)
+		.and_then(|commit_id| repository.find_commit(commit_id));
```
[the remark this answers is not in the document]
orhun: Yup, the ordering should be like that and it does not change. Not sure what you mean by the second part, does that affect us?

git-cliff-core/Cargo.toml
orhun: If we use `hashmap` then the items are not ordered which is probably not something that we want.

Also why enable the `serde` feature?

git-cliff-core/src/changelog.rs
orhun: I think that's fine (also sounds cool :D)

orhun (commented):
Looks nice!

Some things to do:

- Add [fixture tests](https://github.com/orhun/git-cliff/tree/main/.github/fixtures) to ensure the functionality
- Update documentation about the new configuration option

lehmanju:
Thanks for the review! I'll have time to address your comments in about a week.

git-cliff-core/Cargo.toml
lehmanju: serde is needed now because `Release`  needs to be serializable and as a consequence `indexmap` .

git-cliff/src/lib.rs
```diff
@@ -77,6 +81,56 @@ fn check_new_version() {
...
+	let first_commit = release
+		.previous
+		.as_ref()
+		.and_then(|previous_release| previous_release.commit_id.clone())
+		.and_then(|commit_id| repository.find_commit(&commit_id));
+	let last_commit = release
+		.commit_id
+		.clone()
+		.and_then(|commit_id| repository.find_commit(&commit_id));
+	let commit_range = first_commit.zip(last_commit);
```
lehmanju, line 99: This commit range generation was initially wrong, I needed to use the tag commit id from the previous and current release.

git-cliff/src/lib.rs
```diff
@@ -300,6 +355,9 @@ fn process_repository<'a>(
 			previous_release.previous = None;
 			release.previous = Some(Box::new(previous_release));
 			previous_release = release.clone();
+			if recurse_submodules {
```
lehmanju, line 358: Because previous release is needed, I moved this to the bottom of the for loop. Else it wouldn't be initialized.

git-cliff-core/src/release.rs
```diff
@@ -31,35 +33,40 @@ use serde_json::value::Value;
...
-	pub previous:   Option<Box<Release<'a>>>,
+	pub previous:          Option<Box<Release<'a>>>,
 	/// Repository path.
-	pub repository: Option<String>,
+	pub repository:        Option<String>,
+	/// Submodule commits.
+	///
+	/// Maps submodule path to a list of commits.
+	#[serde(rename = "submodule_commits")]
+	pub submodule_commits: Option<HashMap<String, Vec<Commit<'a>>>>,
```
lehmanju, line 55: Changed to Hashmap because indexmap serde implementation doesn't preserve order anyway and order doesn't matter.

git-cliff-core/src/changelog.rs
```diff
@@ -186,6 +145,58 @@ impl<'a> Changelog<'a> {
 		Ok(())
 	}
 
+	fn process_commit_list(
```
lehmanju, line 148: This is new, because commit preprocessing now needs to be done for release commits AND submodule commits.

lehmanju (commented):
Did a bunch of fixes:
- added fixture test
- fixed first/last commit bug (submodule commit range now correct)
- fixed commit preprocessing for submodule commits

@orhun would appreciate another review.

lehmanju:
There does seem to be something wrong with the fixture tests in general: https://github.com/orhun/git-cliff/actions/runs/13907737767/job/38914649566#step:3:913

git-cliff-core/src/release.rs
lehmanju: removed the Option<> wrapping and used an empty hashmap instead. results in a cleaner API during templating.

lehmanju:
The more I test this in a real scenario the more corner cases I find:
unreleased and first release don't contain any submodule commits although they should

orhun (commented):
Looking good! Left some comments that I believe will make the implementation/documentation better.

The fixtures should be fixed in #1099 - ignore the ones that are failing for now.

website/docs/templating/context.md
```diff
@@ -64,6 +64,15 @@ following context is generated to use for templating:
   "commit_id": "a440c6eb26404be4877b7e3ad592bfaa5d4eb210 (release commit)",
   "timestamp": 1625169301,
   "repository": "/path/to/repository",
+  "submodule_commits": {
+    "<relative/submodule_path/in/repository>": [
```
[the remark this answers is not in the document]
lehmanju: Do you mean, if git-cliff is invoked from a repository subdirectory that is not a submodule or from the submodule itself?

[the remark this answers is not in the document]
lehmanju: added another fixture test for include-path. if git-cliff is invoked inside a submodule, it will only be able to process the submodule itself. if git-cliff is invoked inside a normal monorepo directory, it should work as expected. the path that is used to find the submodule is gathered from the commit itself, so it doesn't depend on where git-cliff is invoked.

lehmanju:
@orhun If CI turns out good (which it should), this is ready for a (hopefully) final review.

- docs updated
- fixture tests improved
- edge cases (submodule added/removed) working
- releases without commit id working

.github/fixtures/test-submodules-include-path/cliff.toml
orhun, line 1: This fixture config can now be further simplified thanks to #1087 

Can you remove the default values and just leave the ones that affect the fixture result?

.github/fixtures/test-submodules/cliff.toml
orhun, line 1: This config can be simplified, see my comment above.

orhun (commented):
Thank you, this looks great! Only a couple of tweaks left:

- I left a review about the fixture configs, they can be simplified.
- Can you update the default config (`config/cliff.toml`) and include the `recurse_submodules` there?
- Lastly, we need to rebase on `main` :)

And sorry for late review, I've been traveling...

lehmanju:
Any chance to get a new release by Monday? This would really help me at my job.

orhun:
I was already planning a release soon, I'll try to push it sooner :)

PR #1090, opened 2025-03-12
refactor: initialize config structs with default values

### LINKED ISSUE #1087: Initialize config structs with default values

### Feature description

Currently, all config structs contain `Option<>`. These options have some value if it was set in the configuration file or environment variables, and `None` otherwise. Default values are not taken into consideration at this point.

Whenever an option is used in the application code, the consumer must take care of unpacking the `Option<>` and substituting a default value if needed. If an option is used in multiple places, this unpacking needs to be repeated every time.

### Desired solution

All options should be initialised with their respective default values when loading the config. This would have two benefits:

1. All default values can be defined in one place.
2. Handling of config options is simplied throughout the code.

### Alternatives considered

none

### Additional context

_No response_

Cyclonit:
## Description

This change uses the embedded configuration to initialize configuration options with default values whenever possible. This moves the concern of defining default values into the embedded config and reduces complexity when using the configuration.

## Motivation and Context

Closes #1087.

Using the embedded config.toml to initialize the config structs allows us to use well defined defaults in the code where previously we had to repeatedly unwrap `Option<>` in lots of places. The embedded config.toml was chosen because it is the config used by git-cliff itself and should follow all of our best practices.

This change also allowed for minimizing the fixtures' config files. Only values that are different to the defaults remain. Thus fixtures are less cluttered and easier to maintain.

## How Has This Been Tested?

Regression testing was done using the existing fixtures and integration tests.

## Types of Changes

- [ ] Other 

## Checklist:


orhun (approved):
Nice! The config loading is pretty convenient, I added a comment about that.

orhun:
Uh-oh, fixtures seems to be broken 😵

.github/fixtures/test-fixtures-locally.sh
```diff
@@ -32,4 +32,4 @@ fi
 
 # Show results
 echo -e "\n---Run git-cliff---" >&2
-cargo run --manifest-path "$SCRIPT_DIR/../../Cargo.toml" -- -vv --config "$FIXTURES_DIR/cliff.toml" "${@:2}"
+cargo run --manifest-path "$SCRIPT_DIR/../../Cargo.toml" -- -vv --config "$FIXTURES_DIR/cliff.toml" "${@:2}" > "$FIXTURES_DIR/output.md"
```
orhun, line 35: What's the reasoning for this change? Is the file being used anywhere or just for testing purposes?

I'd prefer to see the output in the terminal tbh - maybe `cat` it afterwards?

.github/fixtures/test-bitbucket-integration/expected.md
```diff
@@ -17,4 +21,4 @@
 ### New Contributors
 * @orhun <orhun@archlinux.org> made their first contribution
 
-<!-- generated by -cliff -->
```
orhun, line 20: Oh wow, any idea what happened here before?

.github/fixtures/test-commit-preprocessors/expected.md
```diff
@@ -4,15 +4,15 @@ All notable changes to this project will be documented in this file.
 
 ## [0.1.0] - 2022-04-06
 
-### Feat
+### <!-- 0 -->🚀 Features
```
orhun, line 7: Why the ordering changes?

orhun (commented):
Just to double check: the fixture configs has been trimmed because we now have those values in the default config, right?

Cyclonit:
> Just to double check: the fixture configs has been trimmed because we now have those values in the default config, right?

Yes exactly. Imho duplicating default values over and over again is bad prwctice because it makes it difficult to find what is different.

.github/fixtures/test-bitbucket-integration/expected.md
Cyclonit: Both the template and expected output had the broken string.

https://github.com/orhun/git-cliff/pull/1090/files/a29045d02bcd67e5fcb4fe7d5904b5fffc1036ff#diff-4c1d6e9d77c9f3447022f6155d1d1fa6a722fa96e0e8194fe557a9931879403bL43

.github/fixtures/test-commit-preprocessors/expected.md
Cyclonit: The default config has the ordering defined using the `<!-- N -->` in the group names. I simply reordered the expected outputs' groups to align to it. Otherwise I'dhave to add custom `commit_parsers` to the fixture jzst to preserve the old ordering. I didn't consider that to be important.

orhun (approved):
Awesome, thanks!

Issue #1041, opened 2025-02-07
Add commit_range to template context


jdrst:
Hey,

i would like to add links to diffs of certain files to the changelog. 
Is there a way to get the commit range git cliff uses or was called with (in my case `git cliff commitA..commitB`)? As far as i could see it's not part of the context.

Thanks

orhun:
Good idea! We simply need to add `commit_range` to `Release` struct and update the other parts accordingly. See #721 for an example.

Marking this as good-first-issue, would you be interested in contributing?

jdrst:
i'll try to make some time next week and have a look.

orhun:
Perfect, assigned to you! Let me know if you need any assistance.

jdrst:
Hey, 

i finally could make the time to have a look. Unless i'm overlooking something this might not be as straight forward as thought. Here are my observations:

Currently the commit range is passed when [getting the commits](https://github.com/orhun/git-cliff/blob/63129ce0c481c00c06bc9ee6271da76e99a6e933/git-cliff-core/src/repo.rs#L116) for changelog creation, we don't have ownership of the repo there. 
As far as i could tell, we might not even "know" the actual commits used, at that time, because they might be internal to the revwalker. What we (afterwards) do have, is a list of sorted commits that is produced. So it might be possible to add something [here](https://github.com/orhun/git-cliff/blob/63129ce0c481c00c06bc9ee6271da76e99a6e933/git-cliff/src/lib.rs#L275) akin to
```rust
//...
	let first_commit = commits.first().unwrap().id();
	let last_commit = commits.last().unwrap().id();
	let commit_range =
		CommitRange::new(&first_commit.to_string(), &last_commit.to_string());
	for git_commit in commits.iter().rev() {
		let release = releases.last_mut().unwrap();
                //..
		release.commit_range = Some(commit_range.clone());
// ...
```
with
```rust
/// Commit range (first..last)
pub struct CommitRange {
	from: String,
	to:   String,
}
```

From what i could tell this is currently not easily testable though, right?

Let me know your thoughts.

Cyclonit:
Can you elaborate on the use case for this request? Do you want to generate a diff between a change's commit and the previous release?

orhun:
> Do you want to generate a diff between a change's commit and the previous release?

I think yes... I cannot think of another use case.

> As far as i could tell, we might not even "know" the actual commits used, at that time, because they might be internal to the revwalker. 

Hmm yeah, good point.

[...]

PR #1123, opened 2025-04-03
feat(context): add release commit range

jdrst:
## Description

Adds the used `commit_range` for each release. 
`commit_range` includes the full length commit ids (`commit_range.from` and `commit_range.to`) and the short ids (`commit_range.from_short` and `commit_range.to_short`) 
Also added related docs (and documentation for the sort_commits default)

## Motivation and Context

See #1041

## How Has This Been Tested?

Via integration_test and three new fixtures.

## Types of Changes

- [ ] Other

## Checklist:

/edit: added a third fixture. no more caveats + documentation and rustfmt/clippy

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,7 +30,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
 dependencies = [
  "cfg-if",
- "getrandom",
+ "getrandom 0.2.15",
  "once_cell",
  "version_check",
  "zerocopy",
@@ -126,9 +126,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.89"
+version = "1.0.98"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "86fdf8605db99b54d3cd748a44c6d04df638eb5dafb219b135d0149bd0db01f6"
+checksum = "e16d2d3311acee920a9eb8d33b8cbc1787ce4a264e85f964c2404b969bdcd487"
 
 [[package]]
 name = "arraydeque"
@@ -158,9 +158,9 @@ dependencies = [
 
 [[package]]
 name = "async-trait"
-version = "0.1.82"
+version = "0.1.88"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a27b8a3a6e1a44fa4c8baf1f653e4172e81486d4941f2237e20dc2d0cf4ddff1"
+checksum = "e539d3fca749fcee5236ab05e93a52867dd549cc157c8cb7f99595f3cedffdb5"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -211,19 +211,20 @@ dependencies = [
 
 [[package]]
 name = "bincode"
-version = "2.0.0-rc.3"
+version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f11ea1a0346b94ef188834a65c068a03aec181c94896d481d7a0a40d85b0ce95"
+checksum = "36eaf5d7b090263e8150820482d5d93cd964a81e4019913c972f4edcc6edb740"
 dependencies = [
  "bincode_derive",
  "serde",
+ "unty",
 ]
 
 [[package]]
 name = "bincode_derive"
-version = "2.0.0-rc.3"
+version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7e30759b3b99a1b802a7a3aa21c85c3ded5c28e1c83170d82d70f08bbf7f3e4c"
+checksum = "bf95709a440f45e986983918d0e8a1f30a9b1df04918fc828670606804ac3c09"
 dependencies = [
  "virtue",
 ]
@@ -369,9 +370,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.5.27"
+version = "4.5.37"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "769b0145982b4b48713e01ec42d61614425f27b7058bda7180a3a41f30104796"
+checksum = "eccb054f56cbd38340b380d4a8e69ef1f02f1af43db2f0cc817a4774d80ae071"
 dependencies = [
  "clap_builder",
  "clap_derive",
@@ -379,9 +380,9 @@ dependencies = [
 
 [[package]]
 name = "clap_builder"
-version = "4.5.27"
+version = "4.5.37"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1b26884eb4b57140e4d2d93652abfa49498b938b3c9179f9fc487b0acc3edad7"
+checksum = "efd9466fac8543255d3b1fcad4762c5e116ffe808c8a3043d4263cd4fd4862a2"
 dependencies = [
  "anstream",
  "anstyle",
@@ -392,18 +393,18 @@ dependencies = [
 
 [[package]]
 name = "clap_complete"
-version = "4.5.42"
+version = "4.5.47"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "33a7e468e750fa4b6be660e8b5651ad47372e8fb114030b594c2d75d48c5ffd0"
+checksum = "c06f5378ea264ad4f82bbc826628b5aad714a75abf6ece087e923010eb937fb6"
 dependencies = [
  "clap",
 ]
 
 [[package]]
 name = "clap_derive"
-version = "4.5.24"
+version = "4.5.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "54b755194d6389280185988721fffba69495eed5ee9feeee9a599b53db80318c"
+checksum = "09176aae279615badda0765c0c0b3f6ed53f4709118af73cf4655d85d1530cd7"
 dependencies = [
  "heck",
  "proc-macro2",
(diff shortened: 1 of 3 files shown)


jdrst:
working on it :D

jdrst:
i am unsure how to fix that timeout. is this flaky?

orhun:
That links check should be fine... I'll take care of it later

website/docs/usage/adding-commits.md
```diff
@@ -27,3 +27,5 @@ git commit -m "$commit_msg"
 ```
 
 The commit SHA will be empty as default when `--with-commit` is used. Specify the hash with a message separated by single whitespace for setting the commit SHA. e.g. `--with-commit "8f55e69eba6e6ce811ace32bd84cc82215673cb6 feat: add X"`
+
+Custom commits will have no effects on the commit range.
```
orhun, line 31: This is a bit unclear for the reader - maybe update it like:

```
:::info

 These added commits will have empty [`commit_range`](docs/etc/etc) variable etc.

::: 
```

Or something along those lines.

git-cliff-core/src/commit_range.rs
orhun, line 1: I would love to rename this to `range` for brevity :)

git-cliff-core/src/commit_range.rs
```diff
@@ -0,0 +1,27 @@
+use serde::{
+	Deserialize,
+	Serialize,
+};
+
+use crate::commit::Commit;
+
+#[derive(Default, Debug, Clone, PartialEq, Serialize, Deserialize)]
+/// Commit range (from..to or from_short..to_short)
```
orhun, line 9:
```suggestion
/// Commit range (from..to or from_short..to_short)
#[derive(Default, Debug, Clone, PartialEq, Serialize, Deserialize)]
```

git-cliff-core/src/commit_range.rs
```diff
@@ -0,0 +1,27 @@
...
+pub struct CommitRange {
+	from:       String,
+	to:         String,
+	from_short: String,
+	to_short:   String,
```
orhun, line 14: Can you add comments explaining what these fields are?

git-cliff-core/src/commit_range.rs
```diff
@@ -0,0 +1,27 @@
...
+}
+
+impl CommitRange {
+	/// Creates a new [`CommitRange`] from [`crate::commit::Commit`].
+	pub fn new(from: &Commit, to: &Commit) -> Self {
+		Self {
+			from:       from.id.clone(),
+			to:         to.id.clone(),
+			from_short: from.short_id.clone(),
+			to_short:   to.short_id.clone(),
```
orhun, line 24: I'm thinking maybe we can remove the `short` variants... the templating can be used to trim them anyways.

e.g.

```
{{ commit_range.to | truncate(length=7, end="") }}
``` 

Also I have another idea which might change the whole implementation but... maybe adding this to the `commit` object would be cleaner?

e.g.

```
{{ commit.range.to | truncate(length=7, end="") }}
``` 

What do you think?

orhun (commented):
Thanks for the implementation!

I had some comments and ideas :)

git-cliff-core/src/commit_range.rs
jdrst: > I'm thinking maybe we can remove the short variants... the templating can be used to trim them anyways.

Short ids aren't necessarily unique when just truncated. I'd consider using the actual short id coming from git as more robust.

> Also I have another idea which might change the whole implementation but... maybe adding this to the commit object would be cleaner?

I'm not sure i can follow? The commit has no idea about a range, or does it? Or do you mean, instead of having the range on the release, have it on every commit in the release? I think i'd consider that counterintuitive.

orhun: > Short ids aren't necessarily unique when just truncated. I'd consider using the actual short id coming from git as more robust.

I'd like to double down on this though... We don't have 2 variants (short, long) in the `commit` object itself and it might lead to inconsistencies (in very rare cases) if we go with short-long here. So I'd prefer we only go with the long variant here.

Thoughts?

orhun:
We need to rebase on `main` and fix conflicts 😊

git-cliff-core/src/commit_range.rs
jdrst: I'd also like to double down :D (one last try). 
What kind of inconsistencies?
I'd rather change e.g. the commit in the context to be an object with long+short (that'd be breaking though).

orhun: > What kind of inconsistencies?

You know, we'll be supporting short hashes in one variable (`commit_range`) but not for the other (`commit`) which is not great.

The reason why I don't like supporting short commit hashes is because 7 characters is not long enough and it's more probable that two entirely different commits might belong to the same short hash.

Even worse, you can override the first 7 digits of a commit hash with a tool like [`lucky-commit`](https://github.com/not-an-aardvark/lucky-commit):

```
$ git log
1f6383a Some commit

$ lucky_commit

$ git log
0000000 Some commit
```

So you can have a repo with all `000000` ids, see [this git history](https://github.com/not-an-aardvark/lucky-commit/commits/main/).

jdrst: > You know, we'll be supporting short hashes in one variable (commit_range) but not for the other (commit) which is not great.

That i absolutely understand, hence my proposition to also add it to the commit itself. But one could also argue that `commit` and `commit_range` are different use cases with different needs. 

> The reason why I don't like supporting short commit hashes is because 7 characters is not long enough and it's more probable that two entirely different commits might belong to the same short hash.

That is exactly my point, just from a different perspective :D. This PR does make use of [git_object_short_id](https://libgit2.org/docs/reference/main/object/git_object_short_id.html) to circumvent that ([see here](https://github.com/jdrst/git-cliff/blob/1b8ecf8f3c527d5ce57f2538d36797231f21f1b6/git-cliff-core/src/commit.rs#L186))

Sorry for being so persistent. You can just put your foot down and say "Please remove the short id"!

Cyclonit: I'll try playing tie-breaker ;)

I'd remove the short variants too. We are talking about configuration here. I don't see a point in ommitting a few bytes here. Putting in the long hash doesn't hurt and is more expressive.

Having only one set simplifies both the code and the documentation too.

orhun: Hey again, needed to take some break due to a small injury but I'm hoping to get this merged before the release soon :)

So I think we should go without short hashes here and rebase on `main` to fix the conflicts. The rest should be ready.

@jdrst do you mind incorporating these changes? 😊

jdrst:
Hey, no problem! I'm currently on vacation though. I could offer some time within the week after the next (so somewhere between 28.4 and 2.5). Does that still fit your timeframe? 
If not, i think you have editing rights, so if it's really pressing you could do it yourself. The rebase was uncomplicated, i just didn't push it yet :/

orhun:
Hey that works, I also tried to do changes here but I can't push to your branch :/

```
 ! [remote rejected]   jdrst/main -> jdrst/main (permission denied)
error: failed to push some refs to 'https://github.com/jdrst/git-cliff.git'
```

git-cliff-core/src/range.rs
orhun, line 1: We can move the entirety to the `commit` module I think.

git-cliff/src/lib.rs
```diff
@@ -397,6 +398,22 @@ fn process_repository<'a>(
 		}
 	}
 
+	// Set the commit ranges for all releases
+	for release in &mut releases {
```
orhun, line 402: Is it possible that we use an already existing loop for this instead of looping through releases again?

e.g. we can set this in

https://github.com/orhun/git-cliff/blob/aec41befc701e786c01ec4ba5399c67df1981779/git-cliff/src/lib.rs#L302

Maybe

jdrst:
> Hey that works, I also tried to do changes here but I can't push to your branch :/
> 
> ```
>  ! [remote rejected]   jdrst/main -> jdrst/main (permission denied)
> error: failed to push some refs to 'https://github.com/jdrst/git-cliff.git'
> ```

That is weird..
image

I just turned it off and on again. Maybe it broke because i accidentally force pushed no changes in between and this PR got closed automatically?

git-cliff/src/lib.rs
jdrst: Will have a look. From what i remember it would've needed a lot of branching/complicated logic because we don't know all of the commits (or the release) we're in until everything is added (and then we can only add it before we switch to the previous release).

I think i opted for this because it is O(n) with n being the number of releases). That should still be fine even with lots of releases.

lehmanju: Have a look at `process_submodules`. It does the same thing because it also needs to have the first and last commit of a release. Maybe you can merge the two loops?

orhun:
No I didn't work :/ Maybe you have branch protection on on `main` or something, I don't really know...

jdrst:
> No it didn't work :/ Maybe you have branch protection on on `main` or something, I don't really know...

no idea.. i just forked. no branch protection active. the hint even states 

> If checked, users with write access to orhun/git-cliff can add new commits to your main branch.
> You can always change this setting later.

i just added you as contributor..

orhun:
Ok I just pushed a commit to there... it doesn't show up in this PR though. What is going on...

orhun:
Ok I messed up something

orhun:
Can we maybe create a new branch and open a new PR from there? Creating a PR from the `main` branch makes things super confusing...

jdrst:
> Can we maybe create a new branch and open a new PR from there? Creating a PR from the `main` branch makes things super confusing...

aight, will do next week. you could upload a patch here or smth.

orhun:
The changes are in the `main` branch here: https://github.com/jdrst/git-cliff/tree/main

jdrst:
> The changes are in the `main` branch here: https://github.com/jdrst/git-cliff/tree/main

currently rebasing, cleaning up and incorporating requested changes, will pr soon.
i squashed your commit into mine and added you as co-author. i hope that's okay?

orhun:
yup! thanks

PR #1137, opened 2025-04-28
fix: validate bumped version against tag pattern

### LINKED ISSUE #1095: Add offline/local version of --bumped-version

### Feature description

This feature would enable calculation of the bumped version without requiring an authentication token (e.g. GITLAB_TOKEN) and without need to access the remote. It purely consults the local git history and tags to make its calculation.

### Desired solution

It would be nice if --bumped-version could be local/offline by default, but alternatively, one could add a flag that sets that preference, e.g. --offline or --local.

### Alternatives considered

I don't think there is an alternative, other than trying a different utility, but then we would have to define our "bump rules" for two different tools and keep them in sync.

### Additional context

_No response_

peiqing6888:
When using --bumped-version with a tag pattern configured, git-cliff now validates the generated version against the pattern and returns an error if it doesn't match.

Fixes #1095

## Description

Added validation for the bumped version against the configured tag pattern. When using `--bumped-version` with a tag pattern configured, git-cliff will now check if the generated version matches the pattern. If it doesn't match, it will return an error with a descriptive message instead of silently creating a non-compliant version.

## Motivation and Context

Previously, when using `--bumped-version` with a tag pattern configured, git-cliff would generate a version (e.g. "0.1.0") without validating it against the configured tag pattern. This could lead to confusion when the generated version doesn't match the project's versioning scheme.

This change ensures that the generated version complies with the project's versioning pattern, providing better consistency and immediate feedback when there's a mismatch.

## How Has This Been Tested?

Added a new test case `test_bump_version_with_tag_pattern` that verifies:
- Version generation fails with appropriate error when the pattern doesn't match (e.g. pattern: "foo-[0-9]+\.[0-9]+\.[0-9]+")
- Version generation succeeds when the pattern matches (e.g. pattern: "v[0-9]+\.[0-9]+\.[0-9]+")

The test covers both success and failure scenarios, ensuring proper error handling and pattern validation.

## Types of Changes

## Checklist:

### DIFF
--- a/git-cliff-core/src/changelog.rs
+++ b/git-cliff-core/src/changelog.rs
@@ -577,6 +577,18 @@ impl<'a> Changelog<'a> {
 			if last_release.version.is_none() {
 				let next_version = last_release
 					.calculate_next_version_with_config(&self.config.bump)?;
+
+				// Validate against tag pattern if configured
+				if let Some(tag_pattern) = &self.config.git.tag_pattern {
+					if !tag_pattern.is_match(&next_version) {
+						return Err(Error::ChangelogError(format!(
+							"Bumped version '{}' does not match configured tag pattern '{}'",
+							next_version,
+							tag_pattern
+						)));
+					}
+				}
+
 				debug!("Bumping the version to {next_version}");
 				last_release.version = Some(next_version.to_string());
 				last_release.timestamp = SystemTime::now()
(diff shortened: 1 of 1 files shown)


git-cliff-core/src/changelog.rs
```diff
@@ -577,6 +577,18 @@ impl<'a> Changelog<'a> {
...
 					.calculate_next_version_with_config(&self.config.bump)?;
+
+				// Validate against tag pattern if configured
+				if let Some(tag_pattern) = &self.config.git.tag_pattern {
+					if !tag_pattern.is_match(&next_version) {
+						return Err(Error::ChangelogError(format!(
+							"Bumped version '{}' does not match configured tag pattern '{}'",
+							next_version,
+							tag_pattern
+						)));
```
orhun, line 588:
```suggestion
						return Err(Error::ChangelogError(format!(
							"Bumped version '{next_version}' does not match \
							 configured tag pattern '{tag_version}'",
						)));
```

orhun (commented):
Hello, thanks for the PR!

This looks good, however I'm not sure how it relates to #1095 - did you maybe link to an incorrect issue there?

Issue #452, opened 2024-01-17
Add commit statistics to the changelog


orhun:
**Is your feature request related to a problem? Please describe.**

[`cargo-smart-release`](https://github.com/Byron/cargo-smart-release) has a cool feature that adds the commits statistics to the changelog:

> Commit Statistics
> 7 commits contributed to the release.
> 83 days passed between releases.
> 1 commit was understood as [conventional](https://www.conventionalcommits.org/).
> 0 issues like '(#ID)' were seen in commit messages

**Describe the solution you'd like**

Implement the same functionality in `git-cliff` and make it configurable via template.

I'm thinking it can be accessed via `{{ statistics }}` variable.

**Describe alternatives you've considered**

None.

**Additional context**

See `CommitStatistics` struct in https://github.com/Byron/cargo-smart-release/blob/main/src/changelog/section/segment.rs

melMass:
I tried a few tools like git-cliff before but it's by far my favorite one!
This issue is the last "critical" thing I miss from it. Mainly to be able to spot missed/ignored commits.
Thanks for the great work

ognis1205:
Can I take this one?

orhun:
Yup!

orhun:
I don't have a specific list in mind... would love to see what the implementation looks like and then decide based on that :)

Issue #1148, opened 2025-05-19
Make git short sha available


hmvp:
### Is there an existing issue or pull request for this?

- [x] I have searched the existing issues and pull requests

### Feature description

I want to generate links to the gitlab webinterface  for commits like `[{{ commit.id }}](https://gitlab.something.com/user/project/commits/{{ commit.id }})` but the full id is quite long and I rather use the short sha as the git cli provides it...

### Desired solution

I would like to be able use `commit.short_id` or something next to `commit.id` in templates 

### Alternatives considered

A template filter that converts the long sha into a short one.  This will not work as expected because git does some magic to adapt the length of the short sha so that there is low chance on collisions.

orhun:
Ah got it, you want SHA-1

It would be nice to add a custom filter (similar to [`upper_first`](https://github.com/orhun/git-cliff/blob/7148b2dbff518eb71c276f3c256db49970730863/git-cliff-core/src/template.rs#L61)) then!

PR #1151, opened 2025-05-30
feat: add commit statistics to the changelog

ognis1205:
## Description

This adds support for computing and exposing aggregated commit statistics:

 - Total number of commits
 - Duration (in days) between the first and last commit
 - Number of conventional commits
 - Link references and their counts
 - Days since the previous release

The data is accessible in templates using the {{ statistics }} variable.

## Implementation Notes

- The core logic is located in [`Release::aggregate_statistics`](https://github.com/orhun/git-cliff/blob/63c3fea6efc7233d25544887be6d78f3c3509b09/git-cliff-core/src/release.rs#L244).
- The aggregation is performed based on the snapshot of `commits` at the time of execution.
- Statistics are computed after `Changelog::process_commits`, meaning they reflect the state of commits after they have been split, filtered, and transformed (e.g., via `into_conventional` and/or `parse_links`) by any processing logic.

## Motivation and Context

This change is inspired by the cargo-smart-release tool, which provides similar release statistics.
Including commit statistics in the changelog offers better insight into the release scope and development activity.
This change addresses [FEATURE REQUEST #452] and makes the statistics available in a configurable way through templates.

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/.github/fixtures/test-release-statistics/cliff.toml
+++ b/.github/fixtures/test-release-statistics/cliff.toml
@@ -1,0 +1,46 @@
+# git-cliff ~ configuration file
+# https://git-cliff.org/docs/configuration
+
+[changelog]
+# A Tera template to be rendered for each release in the changelog.
+# See https://keats.github.io/tera/docs/#introduction
+body = """
+{% if version %}\
+    ## [{{ version | trim_start_matches(pat="v") }}] - {{ timestamp | date(format="%Y-%m-%d") }}
+{% else %}\
+    ## [unreleased]
+{% endif %}\
+{% for group, commits in commits | group_by(attribute="group") %}
+    ### {{ group | upper_first }}
+    {% for commit in commits %}
+        - {{ commit.message | upper_first }}\
+    {% endfor %}
+{% endfor %}
+### Commit Statistics\n
+- {{ statistics.commit_count }} commit(s) contributed to the release.
+- {{ statistics.commits_timespan | default(value=0) }} day(s) passed between the first and last commit.
+- {{ statistics.conventional_commit_count }} commit(s) parsed as conventional.
+- {{ statistics.links | length }} linked issue(s) detected in commits.
+{%- if statistics.links | length > 0 %}
+	{%- for link in statistics.links %}
+        {{ "  " }}- [{{ link.text }}]({{ link.href }}) (referenced {{ link.count }} time(s))
+	{%- endfor %}
+{%- endif %}
+{%- if statistics.days_passed_since_last_release %}
+	- {{ statistics.days_passed_since_last_release }} day(s) passed between releases.
+{%- endif %}\n\n
+"""
+
+[git]
+# An array of regex based parsers for extracting data from the commit message.
+# Assigns commits to groups.
+# Optionally sets the commit's `scope` and can decide to exclude commits from further processing.
+commit_parsers = [
+    { message = "^feat", group = "Features", default_scope = "app" },
+    { message = "^fix", group = "Bug Fixes", scope = "cli" },
+]
+# An array of link parsers for extracting external references, and turning them into URLs, using regex.
+link_parsers = [
+    { pattern = "#(\\d+)", href = "https://github.com/orhun/git-cliff/issues/$1"},
+    { pattern = "RFC(\\d+)", text = "ietf-rfc$1", href = "https://datatracker.ietf.org/doc/html/rfc$1"},
+]
--- a/.github/fixtures/test-release-statistics/expected.md
+++ b/.github/fixtures/test-release-statistics/expected.md
@@ -1,0 +1,57 @@
+## [unreleased]
+
+### Bug Fixes
+
+- Remove duplication (#452)
+- Ensure URI parsing is RFC3986-compliant
+
+### Features
+
+- Add release statistics feature(#452)
+- Make git short shat available(#1148)
+
+### Test
+
+- Add tests
+
+### Commit Statistics
+
+- 5 commit(s) contributed to the release.
+- 0 day(s) passed between the first and last commit.
+- 5 commit(s) parsed as conventional.
+- 3 linked issue(s) detected in commits.
+  - [#452](https://github.com/orhun/git-cliff/issues/452) (referenced 2 time(s))
+  - [#1148](https://github.com/orhun/git-cliff/issues/1148) (referenced 1 time(s))
+  - [ietf-rfc3986](https://datatracker.ietf.org/doc/html/rfc3986) (referenced 1 time(s))
+- <<DAYS SINCE LAST RELEASE>> day(s) passed between releases.
+
+## [0.2.0] - 2022-04-06
+
+### Bug Fixes
+
+- Fix feature 2
+
+### Features
+
+- Add feature 2
+
+### Commit Statistics
+
+- 2 commit(s) contributed to the release.
+- 0 day(s) passed between the first and last commit.
+- 2 commit(s) parsed as conventional.
+- 0 linked issue(s) detected in commits.
+
+## [0.1.0] - 2022-04-06
+
+### Bug Fixes
+
+- Fix feature 1
+
+### Commit Statistics
+
+- 1 commit(s) contributed to the release.
+- 0 day(s) passed between the first and last commit.
+- 1 commit(s) parsed as conventional.
+- 0 linked issue(s) detected in commits.
+
--- a/git-cliff-core/src/changelog.rs
+++ b/git-cliff-core/src/changelog.rs
@@ -150,6 +150,7 @@ impl<'a> Changelog<'a> {
 						.filter_map(|line| {
 							let mut c = commit.clone();
 							c.message = line.to_string();
+							c.links.clear();
 							if c.message.is_empty() {
 								None
 							} else {
@@ -217,6 +218,7 @@ impl<'a> Changelog<'a> {
 					true
 				}
 			})
+			.map(|release| release.with_statistics())
 			.collect();
 		for skipped_tag in &skipped_tags {
 			if let Some(release_index) = self.releases.iter().position(|release| {
@@ -565,10 +567,12 @@ impl<'a> Changelog<'a> {
 					.calculate_next_version_with_config(&self.config.bump)?;
 				debug!("Bumping the version to {next_version}");
 				last_release.version = Some(next_version.to_string());
-				last_release.timestamp = SystemTime::now()
-					.duration_since(UNIX_EPOCH)?
-					.as_secs()
-					.try_into()?;
+				last_release.timestamp = Some(
+					SystemTime::now()
+						.duration_since(UNIX_EPOCH)?
+						.as_secs()
+						.try_into()?,
+				);
 				return Ok(Some(next_version));
 			}
 		}
@@ -689,8 +693,10 @@ mod test {
 	use regex::Regex;
 
 	use super::*;
+	use crate::commit::Signature;
 	use crate::config::{
-		Bump, ChangelogConfig, CommitParser, Remote, RemoteConfig, TextProcessor,
+		Bump, ChangelogConfig, CommitParser, LinkParser, Remote, RemoteConfig,
+		TextProcessor,
 	};
 
 	fn get_test_data() -> (Config, Vec<Release<'static>>) {
@@ -706,7 +712,22 @@ mod test {
 				### {{ group }}{% for group, commits in commits | group_by(attribute="scope") %}
 				#### {{ group }}{% for commit in commits %}
 				- {{ commit.message }}{% endfor %}
-				{% endfor %}{% endfor %}"#,
+				{% endfor %}{% endfor %}
+				### Commit Statistics
+
+				- {{ statistics.commit_count }} commit(s) contributed to the release.
+				- {{ statistics.commits_timespan | default(value=0) }} day(s) passed between the first and last commit.
+				- {{ statistics.conventional_commit_count }} commit(s) parsed as conventional.
+				- {{ statistics.links | length }} linked issue(s) detected in commits.
+				{%- if statistics.links | length > 0 %}
+					{%- for link in statistics.links %}
+						{{ "  " }}- [{{ link.text }}]({{ link.href }}) (referenced {{ link.count }} time(s))
+					{%- endfor %}
+				{%- endif %}
+				{%- if statistics.days_passed_since_last_release %}
+					- {{ statistics.days_passed_since_last_release }} day(s) passed between releases.
+				{%- endif %}
+				"#,
 				),
 				footer:         Some(String::from(
 					r#"-- total releases: {{ releases | length }} --"#,
@@ -878,7 +899,12 @@ mod test {
 				topo_order:               false,
 				topo_order_commits:       true,
 				sort_commits:             String::from("oldest"),
-				link_parsers:             [].to_vec(),
+				link_parsers:             vec![LinkParser {
+					pattern: Regex::new("#(\\d+)")
+						.expect("issue reference regex should be valid"),
+					href:    String::from("https://github.com/$1"),
+					text:    None,
+				}],
 				limit_commits:            None,
 				recurse_submodules:       None,
 				include_paths:            Vec::new(),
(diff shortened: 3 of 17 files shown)


ognis1205:
Hey @orhun,

I've implemented a feature to include release statistics. I don't think there's any need to update the documentation for this. Please take a look when you have time.

orhun:
I had a quick glance at this and it looks pretty good! I'll find some time to do a more throughout review soon.

> I don't think there's any need to update the documentation for this. 

We need to at least update the [context](https://git-cliff.org/docs/templating/context) page for the newly added `statistics` object.

Also maybe we can add a new example under [`examples/`](https://github.com/orhun/git-cliff/tree/main/examples) such as `statistics.toml` for demonstrating this new template. Or maybe in [tips & tricks](https://git-cliff.org/docs/tips-and-tricks/)? What do you think?

ognis1205:
Hey @orhun , thanks for taking a look!

I agree — updating the [context](https://git-cliff.org/docs/templating/context) page for the new statistics object makes sense, and I’ll take care of that as part of this PR.

For the [examples/](https://github.com/orhun/git-cliff/tree/main/examples), I like the idea of adding a new file under `examples/statistics.toml`, and also referencing it in [tips & tricks](https://git-cliff.org/docs/tips-and-tricks/) to improve visibility. That said, I think it might be cleaner to do those in a follow-up PR, if that sounds good to you.

Let me know your thoughts!

orhun:
Yes, sounds reasonable to me 👍🏼

ognis1205:
FYI: @orhun 

Updated `website/docs/templating/context.md`

git-cliff-core/src/release.rs
orhun, line 1: Can you move the changes in this module (the new `LinkCount`, `Statistics` structs and `aggregate_statistics` function) to a separate module such as `statistics`? I think it would be much cleaner that way.

git-cliff-core/src/release.rs
```diff
@@ -31,6 +35,41 @@ use serde::{
...
+
+/// Aggregated statistics about commits in the release.
+#[derive(Default, Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct Statistics {
+	/// The total number of commits included in the release.
+	pub commit_count:                   usize,
+	/// The time span, in days, from the first to the last commit in the
+	/// release. Only present if there is more than one commit.
+	#[serde(skip_serializing_if = "Option::is_none")]
+	pub commit_duration_days:           Option<i64>,
```
orhun: I feel like we can rename this to something better. Maybe `commits_timespan`?

git-cliff-core/src/release.rs
```diff
@@ -31,6 +35,41 @@ use serde::{
...
+	/// The number of commits that follow the Conventional Commits
+	/// specification.
+	pub conventional_commit_count:      usize,
+	/// The number of times each link was referenced in commit messages.
+	pub link_counts:                    Vec<LinkCount>,
```
orhun: Although it is the count of `LinkCount`, maybe calling this `link_count` would be better. `link_counts` sounds a bit odd.

What do you think?

git-cliff-core/src/release.rs
```diff
@@ -31,6 +35,41 @@ use serde::{
...
+	/// The total number of links referenced in all commit messages.
+	pub total_link_count:               usize,
```
orhun: I'm a bit confused about this... it is not the number of `link_counts`, is it?

git-cliff-core/src/release.rs
```diff
@@ -96,6 +137,16 @@ impl Release<'_> {
 		self.calculate_next_version_with_config(&Bump::default())
 	}
 
+	/// Returns a copy of the release with aggregated statistics populated.
```
orhun: Copy sounds like the `Copy` trait in Rust. Can you rephrase this comment?

git-cliff-core/src/release.rs
```diff
@@ -177,6 +228,81 @@ impl Release<'_> {
...
+	/// its commits:
+	///
+	/// - Counts the total number of commits.
+	/// - Determines the number of days between the first and last commit.
+	/// - Counts the number of commits that follow the Conventional Commits
+	///   specification.
+	/// - Tallies how many times each link appears across all commit messages.
+	/// - Calculates the number of days since the previous release, if
+	///   available.
+	fn aggregate_statistics(&self) -> Statistics {
```
orhun: Maybe we can change this function to `impl From<Release> for Statistics` (and move to `statistics` module) since it's simply constructing a `Statistics` from a `Release`. What do you think?

git-cliff-core/src/commit.rs
```diff
@@ -420,13 +423,23 @@ impl Commit<'_> {
...
 				});
 			}
 		}
-		Ok(self)
+		// NOTE: Deduplication is done here based on (text, href) to avoid duplicate
+		// links. This approach preserves the original order of appearance and parser
+		// application, at the cost of slightly reduced performance compared to using
+		// a HashSet directly during collection.
+		// It also ensures that parse_links remains idempotent even if called
+		// multiple times.
```
orhun: We need to answer questions like "Why deduplication is needed?" and maybe most importantly "why it wasn't needed before?"

How does this affect the existing workflow of the users + is it breaking?

git-cliff-core/src/commit.rs
```diff
@@ -408,7 +410,8 @@ impl Commit<'_> {
 	/// Sets the [`links`] of the commit.
 	///
 	/// [`links`]: Commit::links
-	pub fn parse_links(mut self, parsers: &[LinkParser]) -> Result<Self> {
+	pub fn parse_links(mut self, parsers: &[LinkParser]) -> Self {
+		let mut links: Vec<Link> = vec![];
```
orhun:
```suggestion
		let mut links = Vec::new();
```

.github/fixtures/test-release-statistics/cliff.toml
```diff
@@ -0,0 +1,48 @@
...
+{% for group, commits in commits | group_by(attribute="group") %}
+    ### {{ group | upper_first }}
+    {% for commit in commits %}
+        - {{ commit.message | upper_first }}\
+    {% endfor %}
+{% endfor %}
+### Commit Statistics
+    * {{ statistics.commit_count }} commits contributed to the release.
+	{%- if statistics.commit_duration_days is defined %}
+	* {{ statistics.commit_duration_days }} days spanned between the first and last commit.
```
orhun: This sounds better imo:

```suggestion
	* {{ statistics.commit_duration_days }} days passed between the first and last commit.
```

.github/fixtures/test-release-statistics/cliff.toml
```diff
@@ -0,0 +1,48 @@
...
+	{%- endif %}
+	* {{ statistics.conventional_commit_count }} commit was understood as conventional.
```
orhun:
```suggestion
	* {{ statistics.conventional_commit_count }} commits were parsed as conventional.
```

.github/fixtures/test-release-statistics/cliff.toml
```diff
@@ -0,0 +1,48 @@
...
+	* {{ statistics.total_link_count }} issues like '(#ID)' were seen in commit messages.
```
orhun: Btw I'm struggling to understand how this metric is useful... can you give me an example where you'd want to see this?

git-cliff-core/src/release.rs
ognis1205: I agree, that sounds much cleaner. I'll move them to a separate `statistics` module.

git-cliff-core/src/release.rs
ognis1205: Sounds good — I'll rename it to `commits_timespan`.

git-cliff-core/src/release.rs
ognis1205: You're right — it's not the number of `link_counts` entries. It's actually the sum of the `count` property from each `LinkCount`.

git-cliff-core/src/release.rs
ognis1205: Good point, I'll change the wording to avoid confusion with the `Copy` trait.

git-cliff-core/src/release.rs
ognis1205: Sounds good — implementing `From<Release>` for `Statistics` definitely feels more idiomatic in Rust!

git-cliff-core/src/commit.rs
ognis1205: Great point! This was originally introduced to maintain consistency while supporting the `split_commits` option and to avoid overcounting. But you're right — if we simply count the number of `Link`s appearing in `links`, we can construct `LinkCount` without deduplication. That’s also safer in terms of avoiding a breaking change. Thanks for catching this! I’ll revise the logic to remove deduplication.

.github/fixtures/test-release-statistics/cliff.toml
ognis1205: Sure! First, I started with an implementation to reproduce the kind of output described in the [issue](https://github.com/orhun/git-cliff/issues/452), like:

> Commit Statistics  
> 7 commits contributed to the release.  
> 83 days passed between releases.  
> 1 commit was understood as conventional.  
> 0 issues like '(#ID)' were seen in commit messages.

As I worked on this, I introduced the `LinkCount` struct to provide more flexibility. Regarding `total_link_count`, I thought it could be useful in cases where many links appear in commit messages — listing every individual `LinkCount` might get too verbose. Also, although I’m not deeply familiar with the Tera template engine, I assumed it might be cumbersome to calculate the sum of all `LinkCount::count` values from within the template if only the list of `LinkCount`s was available.

That said, I haven’t come up with a strong practical example where `total_link_count` is essential, so I’m neutral on whether it should stay. I’m happy to go with whatever you think is best.

git-cliff-core/src/release.rs
ognis1205: I named it in the plural form (`link_counts`) because it's a `Vec<LinkCount>`, and I assumed it would be iterated over in templates. Using a plural name gives a subtle hint that the value is iterable, which I think is helpful from a usability perspective—even if the English feels a bit awkward.

That said, I agree it could be clearer. How do you feel about renaming it to `link_count_list` or `links` instead? It keeps the iterable implication while sounding more natural.

ognis1205:
@orhun

Thank you for the review! I’ll make the changes as you suggested.
I also left a few replies to your questions — would appreciate it if you could take a look when you have a moment.
I'll ping you again once it's ready for another round of review.

.github/fixtures/test-release-statistics/cliff.toml
orhun: Got it.. I guess then it might be useful.

FWIW Tera has a [`length`](https://keats.github.io/tera/docs/#length) function to get the count of things.

git-cliff-core/src/release.rs
orhun: Ooh, got it now! We should probably make this more clear in the docs

git-cliff-core/src/release.rs
orhun: `links` sounds like a object which contains the `Link` objects. `link_count_list` is not clear imo.

Also `links` == `link_count`, no? :D

ognis1205:
@orhun

I took another look and I think I now understand the point of confusion more clearly.

What I’m aggregating as `total_link_count` is **NOT** the number of **distinct** links that appear in the release, but rather the total number of **link appearances, including duplicates**. In other words, if the same link appears multiple times, it gets counted multiple times. If this isn’t a very meaningful metric, I’m happy to remove it.

As for the naming of `link_counts`, I think there's still room for discussion. For example, the usage looks like this:

```tera
{%- for link in statistics.link_counts %}
  ** [{{ link.text }}]({{ link.href }}) ({{ link.count }} time{% if link.count > 1 %}s{% endif %} referenced)
{%- endfor %}
```

If we were to name this `link_count` in singular form, wouldn’t that feel counter-intuitive?

ognis1205:
Hey @orhun ,

All issues raised in the review—other than the naming of `link_counts`—have been addressed. The only remaining point is whether `link_counts` is the most appropriate name:

There may have been some misunderstandings due to shortcomings in my documentation or explanation, so I'd like to take a moment here to clarify what `link_counts` represents.

`statistics.link_counts` is a `Vec<LinkCount>`, and `LinkCount` is a struct defined as follows:

```rust
pub struct LinkCount {
    /// Text of the link.
    pub text:  String,
    /// URL of the link
    pub href:  String,
    /// The number of times this link was referenced.
    pub count: usize,
}
```

In other words, this holds the `text` and `href` of each individual `Link` that appears in a release, along with the number of times that link was referenced, i.e., `count`.

 For example, consider the following:
 
 ```md
 ## [unreleased]

### Bug Fixes

- Remove duplication (#452)
- Ensure URI parsing is RFC3986-compliant

### Features

- Add release statistics feature(#452)
- Make git short shat available(#1148)

### Test

- Add tests

### Commit Statistics
* 3 issues like '(#ID)' were seen in commit messages. <-- total number of links appeared in the release
** [#452](https://github.com/orhun/git-cliff/issues/452) (2 times referenced) <-- each LinkCount's count, i.e., 2 times
** [ietf-rfc3986](https://datatracker.ietf.org/doc/html/rfc3986) (1 time referenced) <-- each LinkCount's count, i.e., 1 time
** [#1148](https://github.com/orhun/git-cliff/issues/1148) (1 time referenced) <-- each LinkCount's count, i.e., 1 time
 ```
 
I hope this clears things up.

> links sounds like a object which contains the Link objects. link_count_list is not clear imo.
> Also links == link_count, no? :D

Given the explanation above, using `link_count` for a `Vec<LinkCount>` sounds to me like it could be mistaken for a single `LinkCount` struct. Does that sound odd to a native speaker as well? For example, the usage looks like this:

```
{%- for link in statistics.link_counts %}
  ** [{{ link.text }}]({{ link.href }}) ({{ link.count }} time{% if link.count > 1 %}s{% endif %} referenced)
{%- endfor %}
```

If we were to name this `link_count` in singular form, wouldn’t that feel counter-intuitive?

git-cliff-core/src/release.rs
ognis1205: I actually explained this in more detail in the comment below—feel free to take a look when you have a moment! 🙂

website/docs/templating/context.md
```diff
@@ -176,6 +189,19 @@ If [`conventional_commits`](/docs/configuration/git#conventional_commits) is set
       "from": "(id of the first commit used for this release)",
       "to": "(id of the last commit used for this release)",
   },
+  "statistics": {
+    "commit_count": 1,
+    "commits_timespan": 0,
+    "conventional_commit_count": 0,
+    "link_counts": [
```
orhun: This is almost identical to the `commits.links` object... so let's just rename it to `links`

.github/fixtures/test-release-statistics/expected.md
```diff
@@ -0,0 +1,60 @@
...
+- Make git short shat available(#1148)
+
+### Test
+
+- Add tests
+
+### Commit Statistics
+* 5 commits contributed to the release.
+* 0 days passed between the first and last commit.
+* 5 commits were understood as conventional.
```
orhun: `understood` -> `parsed`

Can you also update the rest of the examples / code about this?

.github/fixtures/test-release-statistics/expected.md
```diff
@@ -0,0 +1,60 @@
...
+* 3 issues like '(#ID)' were seen in commit messages.
+** [#452](https://github.com/orhun/git-cliff/issues/452) (2 times referenced)
+** [#1148](https://github.com/orhun/git-cliff/issues/1148) (1 time referenced)
+** [ietf-rfc3986](https://datatracker.ietf.org/doc/html/rfc3986) (1 time referenced)
+* -19088 days passed between releases.
```
orhun: Uhh... `-19088` days? :D

.github/fixtures/test-release-statistics/expected.md
```diff
@@ -0,0 +1,60 @@
...
+- Add tests
+
+### Commit Statistics
+* 5 commits contributed to the release.
+* 0 days passed between the first and last commit.
+* 5 commits were understood as conventional.
+* 3 issues like '(#ID)' were seen in commit messages.
+** [#452](https://github.com/orhun/git-cliff/issues/452) (2 times referenced)
+** [#1148](https://github.com/orhun/git-cliff/issues/1148) (1 time referenced)
+** [ietf-rfc3986](https://datatracker.ietf.org/doc/html/rfc3986) (1 time referenced)
```
orhun: Please update the config accordingly to get this result:

```suggestion
** [#452](https://github.com/orhun/git-cliff/issues/452) (referenced 2 time(s))
** [#1148](https://github.com/orhun/git-cliff/issues/1148) (referenced 1 time(s))
** [ietf-rfc3986](https://datatracker.ietf.org/doc/html/rfc3986) (references 1 time(s))
```

orhun:
Left a comment. I think we can simply rename that to `links` now

ognis1205:
Hey @orhun ,

I’ve addressed all the points that were raised in the review.
Let me know if there’s anything else you’d like me to revise or clarify.

.github/fixtures/test-release-statistics/expected.md
```diff
@@ -0,0 +1,59 @@
...
+- Add tests
+
+### Commit Statistics
+* 5 commit(s) contributed to the release.
+* 0 day(s) passed between the first and last commit.
+* 5 commit(s) were parsed as conventional.
+* 3 issue(s) like '(#ID)' were seen in commit messages.
+** [#452](https://github.com/orhun/git-cliff/issues/452) (referenced 2 time(s))
+** [#1148](https://github.com/orhun/git-cliff/issues/1148) (referenced 1 time(s))
+** [ietf-rfc3986](https://datatracker.ietf.org/doc/html/rfc3986) (referenced 1 time(s))
```
orhun: I'm not sure if this formatting is correct. It doesn't render correctly for me on GitHub:

image

Let's use `-` and indented `-` instead? So something like:

- Test
  - Sub
  - Sub

.github/fixtures/test-release-statistics/expected.md
```diff
@@ -0,0 +1,59 @@
...
+
+- Add release statistics feature(#452)
+- Make git short shat available(#1148)
+
+### Test
+
+- Add tests
+
+### Commit Statistics
+* 5 commit(s) contributed to the release.
```
orhun: Can you add a newline here?

```suggestion
### Commit Statistics

* 5 commit(s) contributed to the release.
```

.github/fixtures/test-release-statistics/cliff.toml
```diff
@@ -0,0 +1,48 @@
...
+        - {{ commit.message | upper_first }}\
+    {% endfor %}
+{% endfor %}
+### Commit Statistics
+* {{ statistics.commit_count }} commit(s) contributed to the release.
+{%- if statistics.commits_timespan is defined %}
+	* {{ statistics.commits_timespan }} day(s) passed between the first and last commit.
+{%- endif %}
+* {{ statistics.conventional_commit_count }} commit(s) {% if statistics.conventional_commit_count > 1 %}were{% else %}was{% endif %} parsed as conventional.
+* {{ statistics.links | length }} issue(s) like '(#ID)' {% if statistics.links | length != 1 %}were{% else %}was{% endif %} seen in commit messages.
```
orhun: This is too cumbersome... I think we can simplify this message/template.

An example:

Before:

> 3 issue(s) like '(#ID)' were seen in commit messages.

After:

> 3 linked issue(s) detected in commits.

WDYT?

.github/fixtures/test-release-statistics/cliff.toml
```diff
@@ -0,0 +1,48 @@
...
+{% endif %}\
+{% for group, commits in commits | group_by(attribute="group") %}
+    ### {{ group | upper_first }}
+    {% for commit in commits %}
+        - {{ commit.message | upper_first }}\
+    {% endfor %}
+{% endfor %}
+### Commit Statistics
+* {{ statistics.commit_count }} commit(s) contributed to the release.
+{%- if statistics.commits_timespan is defined %}
```
orhun: These `is defined` checks can be simplified as:

```suggestion
{%- if statistics.commits_timespan %}
```

Also, is there a chance that these are not defined?

I think we can further simplify this template.

.github/fixtures/test-release-statistics/cliff.toml
```diff
@@ -0,0 +1,48 @@
...
+* {{ statistics.conventional_commit_count }} commit(s) {% if statistics.conventional_commit_count > 1 %}were{% else %}was{% endif %} parsed as conventional.
+* {{ statistics.links | length }} issue(s) like '(#ID)' {% if statistics.links | length != 1 %}were{% else %}was{% endif %} seen in commit messages.
+{%- if statistics.links | length > 0 %}
+	{%- for link in statistics.links %}
+		** [{{ link.text }}]({{ link.href }}) (referenced {{ link.count }} time(s))
+	{%- endfor %}
+{%- endif %}
+{%- if statistics.days_passed_since_last_release is defined %}
+	* {{ statistics.days_passed_since_last_release }} day(s) passed between releases.
+{%- endif %}\n\n
```
orhun: See my comment above about `is defined` checks.

git-cliff-core/src/changelog.rs
```diff
@@ -1290,6 +1466,14 @@ chore(deps): fix broken deps
 			#### app
 			- merge #5
 
+			### Commit Statistics
+			* 8 commit(s) contributed to the release.
+			* 6 day(s) passed between the first and last commit.
+			* 8 commit(s) were parsed as conventional.
+			* 1 issue(s) like '(#ID)' was seen in commit messages.
+			** [#5](https://github.com/5) (referenced 1 time(s))
+			* -578 day(s) passed between releases.
```
orhun: See my comment about formatting above

orhun (commented):
Nice, we're getting there!

.github/fixtures/test-release-statistics/expected.md
ognis1205: Sounds good — changing it accordingly!

.github/fixtures/test-release-statistics/cliff.toml
ognis1205: Agreed, it's a bit redundant. I'll simplify the message and remove the conditional logic.

ognis1205:
Hey @orhun ,

I've addressed all the feedback in this PR. When you have a moment, I'd really appreciate another look. Thanks!

git-cliff-core/src/commit.rs
```diff
@@ -428,7 +428,8 @@ impl Commit<'_> {
 	/// Sets the [`links`] of the commit.
 	///
 	/// [`links`]: Commit::links
-	pub fn parse_links(mut self, parsers: &[LinkParser]) -> Result<Self> {
+	pub fn parse_links(mut self, parsers: &[LinkParser]) -> Self {
+		self.links.clear();
```
orhun: This shouldn't be necessary:

```suggestion
```

Can you double check?

orhun (approved):
Just one small thing :D

git-cliff-core/src/commit.rs
ognis1205: I think this process is necessary. If the `split_commits` option is set:

- https://github.com/orhun/git-cliff/blob/76d3e819b126837d7d8541c18c083cdd9f80ec62/git-cliff-core/src/changelog.rs#L158  
- https://github.com/orhun/git-cliff/blob/76d3e819b126837d7d8541c18c083cdd9f80ec62/git-cliff-core/src/changelog.rs#L170  

The `Changelog::process_commit` (i.e., `Commit::process`) function is called twice in these two locations, which causes the links to be accumulated/counted twice.  If we don't clear the `links` here, then we may want to explicitly reset it right after the link handling line:

- https://github.com/orhun/git-cliff/blob/76d3e819b126837d7d8541c18c083cdd9f80ec62/git-cliff-core/src/changelog.rs#L166

```rust
c.links = Vec::new();
```

This might make it clearer that the links are intentionally cleared before processing. What do you think?

ognis1205: As mentioned in my commit message, I’ve moved the `links.clear()` call out of `Commit::parse_links` and into `Changelog::process_commit_list` instead, for the reasons described above — to make the reset explicit and scoped to where link accumulation could actually become problematic.

This results in `Commit::links` being accumulated twice unless it's cleared in between. In practice, this leads to changelogs like:

```diff
- [#5](https://github.com/5) (referenced 1 time(s))
+ [#5](https://github.com/5) (referenced 2 time(s))
```

By moving the clear operation to `Changelog::process_commit_list`, we can avoid this duplication without introducing hidden side effects in `Commit::parse_links`, keeping the logic more predictable and context-aware.

Let me know what you think!

ognis1205: Hi @orhun ,

Just a friendly reminder about this PR — no rush at all, but I wanted to check in and see if you had a chance to take a look. Let me know if there’s anything you’d like me to adjust or clarify! Thanks!

git-cliff-core/src/release.rs
```diff
@@ -46,7 +47,7 @@ pub struct Release<'a> {
 	#[serde(rename = "commit_id")]
 	pub commit_id:         Option<String>,
 	/// Timestamp of the release in seconds, from epoch.
-	pub timestamp:         i64,
+	pub timestamp:         Option<i64>,
```
orhun, line 33: One last question, why this was changed to an `Option`?

ognis1205: Thanks for the question. Good question!

Previously, when `timestamp` was defined as `i64` instead of `Option<i64>`, it was initialized with the default value of `0` when `Release::new` or its equivalent was called. This led to an issue where the `days_passed_since_last_release` in the statistics calculation could become negative — as seen here:

https://github.com/orhun/git-cliff/blob/74c33918f85abd356bd1448586640cb791f9ba3b/git-cliff-core/src/statistics.rs#L105

This was pointed out in the comment:

https://github.com/orhun/git-cliff/pull/1151#discussion_r2138346817

To avoid this, I changed `timestamp` to `Option<i64>`, and when the value is `None`, we now fall back to using the current time, as seen here:

https://github.com/orhun/git-cliff/blob/74c33918f85abd356bd1448586640cb791f9ba3b/git-cliff-core/src/statistics.rs#L110

This ensures that `days_passed_since_last_release` is calculated correctly, even in cases where the release timestamp isn’t explicitly set.

orhun: Alright, that's fair, thanks for the explanation!

orhun:
We need a rebase then it's all good to go

ognis1205:
Thanks so much for your thoughtful and patient reviews @orhun — I really appreciate it!
I’ll try to make the next PR smoother based on what I’ve learned this time. I’ll ping you again once the rebase is done.

rustfmt.toml
orhun, line 1: This merge doesn't look right... Can you double check the changes here vs `main`?

ognis1205: Oh, you’re right — I meant to squash and rebase, but it looks like I messed something up. I’ll investigate!

ognis1205: Hi @orhun, I’ll likely go with cherry-picking from a clean branch to simplify the history, since rebase has become a bit messy. It might take a little time, but I’ll ping you as soon as it’s done. Thanks again for your patience!

orhun: all good, thanks!

ognis1205:
Hi @orhun,

I’ve finished squashing and rebasing the commits, and also ran `cargo fmt` afterward.
The PR should now be ready to merge. Let me know if anything else is needed!

ognis1205:
@orhun 

Thanks so much for your thoughtful and patient review — really appreciate it!
I’m glad to hear you think it’ll be useful!

PR #1165, opened 2025-06-04
doc: Fix inconsistency in documentation of Gitea integration

wetneb:
The text says that the extra information is added in a `gitea` nested object, but it's in fact a `remote` object according to the examples.

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/website/docs/integration/bitbucket.md
+++ b/website/docs/integration/bitbucket.md
@@ -96,7 +96,7 @@ https://bitbucket.org/{{ remote.bitbucket.owner }}/{{ remote.bitbucket.repo }}/c
 
 ### Commit authors
 
-For each commit, Bitbucket related values are added as a nested object (named `bitbucket`) to the [template context](/docs/templating/context):
+For each commit, Bitbucket related values are added as a nested object (named `remote`) to the [template context](/docs/templating/context):
 
 ```json
 {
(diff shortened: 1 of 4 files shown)


orhun:
They are actually added as both, e.g. for GitHub:

```json
        "github": {
          "username": "LitoMore",
          "pr_title": "docs(readme): fix twitter badge",
          "pr_number": 1164,
          "pr_labels": [],
          "is_first_time": false
        },
        "remote": {
          "username": "LitoMore",
          "pr_title": "docs(readme): fix twitter badge",
          "pr_number": 1164,
          "pr_labels": [],
          "is_first_time": false
        },
```

For context, the `remote` was added in #822

So in this case we should probably expand the docs to say something like "`remote` or `gitea`", what do you think?

wetneb:
Hmm, yeah then I don't know how to formulate in a way that's not confusing… if you write "`gitea` or `remote`" then as a reader I think I would ask myself "how do I know which?"… Is there a preferred one that should be used to access this information? I would recommend just documenting that and ignoring the other.

orhun:
Makes sense. Then we need to change this to `remote` in all of the other docs (GitHub, GitLab, etc.) as well.

orhun (approved):
Thanks!

PR #1185, opened 2025-06-21
refactor(config): implement FromStr instead of Config::parse_from_str()

joshka:
Rename parse() to load()


orhun (approved):
LGTM, thanks!

Issue #810, opened 2024-08-23
New contributors needs the merge_number after the !


joergklein:
Please take a look at cliff.toml and GitLab (Discussion #730)

new contributors Needs the merge_number after the !
There seems to be a bug with this, I'm not sure what is going on. Feel free to create a separate issue and I can look into it.

In the meantime you can add a guard like this:

```toml
{% for contributor in contributors %}
  {% if contributor.pr_number -%}\
    * [@{{ contributor.username }}](https://gitlab.com/{{ contributor.username }}) made their first contribution in [!{{ contributor.pr_number }}](<REPO>/-/merge_requests/{{ contributor.pr_number }})\
  {% endif -%}\
{%- endfor %}\n
```

Is it possible to put this lines into the code below? The error message witch I get is:

```text
 git-cliff -r . > CHANGELOG.md
 WARN  git_cliff_core::changelog > You are using an experimental feature! Please report bugs at <https://git-cliff.org/issues>
 ERROR git_cliff                 > Template render error:
Variable `contributors` not found in context while rendering 'template'
Cleaning up project directory and file based variables
00:00
ERROR: Job failed: exit code 1
```

cliff.toml

```toml
# git-cliff ~ configuration file
# https://git-cliff.org/docs/configuration

[remote.gitlab]
owner = "user"
repo = "test"
token = ""

[changelog]
# template for the changelog body
# https://keats.github.io/tera/docs/#introduction
header = """
# Changelog\n
This is an automated changelog based on the commits in this repository.\n

### Releases

Check the notes in the [releases](https://gitlab.com/user/test/-/releases) for more information.
"""
# https://keats.github.io/tera/docs/#introduction
body = """
{% if version -%}\
    ## [{{ version | trim_start_matches(pat="v") }}] - {{ timestamp | date(format="%Y-%m-%d") }}
{% else %}\
    ## [Unreleased]
{% endif -%}\

{% for group, commits in commits | group_by(attribute="group") %}
    ### {{ group | upper_first }}
    {% for commit in commits %}
        - {% if commit.breaking %}[**breaking**] {% endif %}{{ commit.message | upper_first | trim }} ([{{ commit.id | truncate(length=7, end="") }}](<REPO>/commit/{{ commit.id }}))\
            {% if commit.gitlab.username and commit.gitlab.username != remote.gitlab.owner %} by [@{{ commit.gitlab.username }}](https://gitlab.com/{{ commit.gitlab.username }}){%- endif -%}
[...]

PR #1186, opened 2025-06-21
Fix GitLab Author Metadata: Use `username` Instead of `author_name`, …

JuryA:
…Add `author_username` Trait, and Improve Release Metadata Extraction (#1)

## Description

This pull request introduces support for retrieving the author's username from pull requests across different remote platforms (GitHub, GitLab, and Bitbucket). It also updates the release metadata logic to incorporate this new functionality. Below is a summary of the most important changes:

### Enhancements to `RemotePullRequest` Trait:

* Added a new method `author_username` to the `RemotePullRequest` trait, which returns the username of the pull request's author. A default implementation returning `None` is provided for backward compatibility. (`git-cliff-core/src/remote/mod.rs`, [git-cliff-core/src/remote/mod.rsR118-R121](diffhunk://#diff-9d830e0be5fc20135aee3ba39934c06263208881475d426a20c85a3b7dbba385R118-R121))

### Platform-Specific Implementations:

* **GitHub**: 
  - Introduced a new struct `GitHubPullRequestUser` to represent the pull request author.
  - Updated the `GitHubPullRequest` struct to include a `user` field of type `Option<GitHubPullRequestUser>`.
  - Implemented the `author_username` method for `GitHubPullRequest` to return the author's username. (`git-cliff-core/src/remote/github.rs`, [[1]](diffhunk://#diff-a9ca597c71fe138c7f11d73a4b7696b61f7a47feeacde88f640e8349518b1a03R98-R104) [[2]](diffhunk://#diff-a9ca597c71fe138c7f11d73a4b7696b61f7a47feeacde88f640e8349518b1a03R122-R123) [[3]](diffhunk://#diff-a9ca597c71fe138c7f11d73a4b7696b61f7a47feeacde88f640e8349518b1a03R144-R147)

* **GitLab**: Implemented the `author_username` method for `GitLabMergeRequest` to return the author's username. (`git-cliff-core/src/remote/gitlab.rs`, [git-cliff-core/src/remote/gitlab.rsR197-R200](diffhunk://#diff-26d62ab23081d1553984f3abec6f6f59d168d32493c43dbc7162431c7239f1b8R197-R200))

* **Bitbucket**: Implemented the `author_username` method for `BitbucketPullRequest` to return the author's username. (`git-cliff-core/src/remote/bitbucket.rs`, [git-cliff-core/src/remote/bitbucket.rsR152-R155](diffhunk://#diff-66cced4c6dfbc084d1fa09a0ca9cf2ea6f5c60957f8767fc9a4ee0473d1421aaR152-R155))

### Updates to Release Metadata Logic:

* Modified the `update_release_metadata` macro to use the `author_username` method when setting the username in release metadata. This ensures that the author's username is correctly captured from the associated pull request, falling back to the existing logic if unavailable. (`git-cliff-core/src/remote/mod.rs`, [git-cliff-core/src/remote/mod.rsL357-R364](diffhunk://#diff-9d830e0be5fc20135aee3ba39934c06263208881475d426a20c85a3b7dbba385L357-R364))

### Test Adjustments:

* Updated test cases in `git-cliff-core/src/release.rs` to include the `user` field with a `None` value for pull requests, ensuring compatibility with the new `GitHubPullRequest` structure. (`git-cliff-core/src/release.rs`, [[1]](diffhunk://#diff-56d2e9b48d9c6d1fee904b269749489630519d3cfd9c2fc6e746057316c6a09dR554) [[2]](diffhunk://#diff-56d2e9b48d9c6d1fee904b269749489630519d3cfd9c2fc6e746057316c6a09dR565) [[3]](diffhunk://#diff-56d2e9b48d9c6d1fee904b269749489630519d3cfd9c2fc6e746057316c6a09dR576) [[4]](diffhunk://#diff-56d2e9b48d9c6d1fee904b269749489630519d3cfd9c2fc6e746057316c6a09dR587) [[5]](diffhunk://#diff-56d2e9b48d9c6d1fee904b269749489630519d3cfd9c2fc6e746057316c6a09dR598)

## Motivation and Context

The primary motivation was to resolve the issue with loading metadata from GitLab.

#810, #687

## How Has This Been Tested?

I’m not much of a programmer—especially not in Rust—but I was able to identify the root of the issue. The tests still need to be completed, and a proper code review is required. (There may be other things as well, but I’m probably not equipped to handle those.)

N/A

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/git-cliff-core/src/remote/mod.rs
+++ b/git-cliff-core/src/remote/mod.rs
@@ -115,6 +115,10 @@ pub trait RemotePullRequest: DynClone {
 	fn labels(&self) -> Vec<String>;
 	/// Merge commit SHA.
 	fn merge_commit(&self) -> Option<String>;
+	/// Username of the author of the pull request.
+	fn author_username(&self) -> Option<String> {
+		None
+	}
 }
 
 dyn_clone::clone_trait_object!(RemotePullRequest);
@@ -354,7 +358,10 @@ macro_rules! update_release_metadata {
 							pr.merge_commit() == Some(v.id().clone()) ||
 								pr.merge_commit() == sha_short
 						});
-						commit.$remote.username = v.username();
+						let username = pull_request
+							.and_then(|pr| pr.author_username())
+							.or_else(|| v.username());
+						commit.$remote.username = username.clone();
 						commit.$remote.pr_number = pull_request.map(|v| v.number());
 						commit.$remote.pr_title =
 							pull_request.and_then(|v| v.title().clone());
--- a/git-cliff-core/src/release.rs
+++ b/git-cliff-core/src/release.rs
@@ -551,6 +551,7 @@ mod test {
 					merge_commit_sha: Some(String::from(
 						"1d244937ee6ceb8e0314a4a201ba93a7a61f2071",
 					)),
+					user:             None,
 					labels:           vec![PullRequestLabel {
 						name: String::from("rust"),
 					}],
@@ -561,6 +562,7 @@ mod test {
 					merge_commit_sha: Some(String::from(
 						"21f6aa587fcb772de13f2fde0e92697c51f84162",
 					)),
+					user:             None,
 					labels:           vec![PullRequestLabel {
 						name: String::from("rust"),
 					}],
@@ -571,6 +573,7 @@ mod test {
 					merge_commit_sha: Some(String::from(
 						"35d8c6b6329ecbcf131d7df02f93c3bbc5ba5973",
 					)),
+					user:             None,
 					labels:           vec![PullRequestLabel {
 						name: String::from("deps"),
 					}],
@@ -581,6 +584,7 @@ mod test {
 					merge_commit_sha: Some(String::from(
 						"4d3ffe4753b923f4d7807c490e650e6624a12074",
 					)),
+					user:             None,
 					labels:           vec![PullRequestLabel {
 						name: String::from("deps"),
 					}],
@@ -591,6 +595,7 @@ mod test {
 					merge_commit_sha: Some(String::from(
 						"5a55e92e5a62dc5bf9872ffb2566959fad98bd05",
 					)),
+					user:             None,
 					labels:           vec![PullRequestLabel {
 						name: String::from("github"),
 					}],
--- a/git-cliff-core/src/remote/bitbucket.rs
+++ b/git-cliff-core/src/remote/bitbucket.rs
@@ -149,6 +149,10 @@ impl RemotePullRequest for BitbucketPullRequest {
 	fn merge_commit(&self) -> Option<String> {
 		Some(self.merge_commit.hash.clone())
 	}
+
+	fn author_username(&self) -> Option<String> {
+		self.author.login.clone()
+	}
 }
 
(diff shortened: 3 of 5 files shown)


JuryA:
@orhun Hi, I’ve proposed a fix, but this is only my third contribution to a Rust project, so I may need some guidance.

I’m a DevOps Engineer rather than a developer, so I apologize in advance if anything is off. :)

I’d really appreciate it if you could take a look—and feel free to correct or suggest improvements where needed.

git-cliff-core/src/remote/mod.rs
```diff
@@ -354,7 +358,10 @@ macro_rules! update_release_metadata {
 							pr.merge_commit() == Some(v.id().clone()) ||
 								pr.merge_commit() == sha_short
 						});
-						commit.$remote.username = v.username();
+						let username = pull_request
+							.and_then(|pr| pr.author_username())
```
ognis1205, line 362: Since commit.remote.username is a field intended to represent the commit author, I believe it would be problematic to associate the PR author here if they are different. Linking the PR author as the commit author in this context might lead to confusion or misrepresentation.

ognis1205 (commented):
Just to confirm, is this PR meant to address [issue #687](https://github.com/orhun/git-cliff/issues/687)?

From my understanding, the root cause seems to be related to deserialization of the GitLab API response. If that's correct, it seems that the current changes might not fully resolve the issue — but please let me know if I misunderstood anything.

git-cliff-core/src/remote/mod.rs
JuryA: The main issue was that the `username` was incorrectly read as `author_name`, which resulted in the metadata being absent in most cases. I noticed this when I wondered why it wasn’t returning `@username`. I then called the API in the same way as `git-cliff`, and in that case, the information was correct. To confirm, I ran a test and compiled with these changes, and voilà—the metadata appeared! There may still be some room for fine-tuning, but the outcome was already much better than before. Therefore, I am confident that this is the root cause of the problem.

JuryA: This is definitely caused by this swap:

```
"gitlab": {
"username": null,
"pr_title": null,
"pr_number": null,
"pr_labels": [],
"is_first_time": false
},
```

JuryA: > Linking the PR author as the commit author in this context might lead to confusion or misrepresentation.

Maybe you’re right! In that case, we can probably use the correct username from the commit—it should also be available there. The main problem is that `author_name` can be anything; it is not a unique identifier that can be easily used to match records.

ognis1205: Thanks for the replay.

My original concern was more about the semantic meaning of the `commit.remote.username` field. Even if we now retrieve the `username`, if that value corresponds to the PR author and not the commit author, I believe it could still lead to incorrect attribution.

https://github.com/orhun/git-cliff/blob/260636682334c6d29ec45fb3a9afc216374ae548/git-cliff-core/src/remote/mod.rs#L94-L95

orhun:
I'm not sure how this is fixing #687 - can you take a step back and explain the actual problem behind the missing metadata and how this PR is fixing it?

PR #1193, opened 2025-06-28
ci: remove deprecated actions and use stable

joshka:
nightly is only necessary for formatting. For most things in git-cliff
stable is sufficient. So use stable for everything else.

actions-rs/toolchain was archived years ago. Generally use
dtolnay/rust-toolchain as a replacement

actions-rs/cargo is similarly archived. Use plain cargo commands

taike-e/install-action automatically uses cargo-binstall to install
tools not setup for install-action directly. So there's no need to
install binstall to install cargo-msrv

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -26,126 +26,93 @@ jobs:
   publish-binaries:
     name: Publish binaries
     needs: generate-changelog
-    runs-on: ${{ matrix.build.os }}
+    runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
-        build:
-          - {
-              NAME: linux-x64-glibc,
-              OS: ubuntu-22.04,
-              TARGET: x86_64-unknown-linux-gnu,
-              NPM_PUBLISH: true,
-              PYPI_PUBLISH: true,
-            }
-          - {
-              NAME: linux-x64-musl,
-              OS: ubuntu-22.04,
-              TARGET: x86_64-unknown-linux-musl,
-              NPM_PUBLISH: false,
-              PYPI_PUBLISH: true,
-            }
-          - {
-              NAME: linux-x86-glibc,
-              OS: ubuntu-22.04,
-              TARGET: i686-unknown-linux-gnu,
-              NPM_PUBLISH: false,
-              PYPI_PUBLISH: false,
-            }
-          - {
-              NAME: linux-x86-musl,
-              OS: ubuntu-22.04,
-              TARGET: i686-unknown-linux-musl,
-              NPM_PUBLISH: false,
-              PYPI_PUBLISH: true,
-            }
-          - {
-              NAME: linux-arm64-glibc,
-              OS: ubuntu-22.04,
-              TARGET: aarch64-unknown-linux-gnu,
-              NPM_PUBLISH: true,
-              PYPI_PUBLISH: true,
-            }
-          - {
-              NAME: linux-arm64-musl,
-              OS: ubuntu-22.04,
-              TARGET: aarch64-unknown-linux-musl,
-              NPM_PUBLISH: false,
-              PYPI_PUBLISH: true,
-            }
-          - {
-              NAME: win32-x64-mingw,
-              OS: windows-2022,
-              TARGET: x86_64-pc-windows-gnu,
-              NPM_PUBLISH: false,
-              PYPI_PUBLISH: false,
-            }
-          - {
-              NAME: win32-x64-msvc,
-              OS: windows-2022,
-              TARGET: x86_64-pc-windows-msvc,
-              NPM_PUBLISH: true,
-              PYPI_PUBLISH: true,
-            }
-          - {
-              NAME: win32-x86-msvc,
-              OS: windows-2022,
-              TARGET: i686-pc-windows-msvc,
-              NPM_PUBLISH: false,
-              PYPI_PUBLISH: true,
-            }
-          - {
-              NAME: win32-arm64-msvc,
-              OS: windows-2022,
-              TARGET: aarch64-pc-windows-msvc,
-              NPM_PUBLISH: true,
-              PYPI_PUBLISH: false,
-            }
-          - {
-              NAME: darwin-x64,
-              OS: macos-14,
-              TARGET: x86_64-apple-darwin,
-              NPM_PUBLISH: true,
-              PYPI_PUBLISH: true,
-            }
-          - {
-              NAME: darwin-arm64,
-              OS: macos-14,
-              TARGET: aarch64-apple-darwin,
-              NPM_PUBLISH: true,
-              PYPI_PUBLISH: true,
-            }
+        include:
+          - name: linux-x64-glibc
+            os: ubuntu-latest
+            target: x86_64-unknown-linux-gnu
+            publish_npm: true
+            publish_pypi: true
+          - name: linux-x64-musl
+            os: ubuntu-latest
+            target: x86_64-unknown-linux-musl
+            publish_npm: false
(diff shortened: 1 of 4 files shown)


.github/workflows/cd.yml
```diff
@@ -34,95 +34,83 @@ jobs:
           - {
               NAME: linux-x64-glibc,
               OS: ubuntu-22.04,
-              TOOLCHAIN: stable,
```
orhun, line 37: Can we keep this matrix variable? I think it would be useful if I want to use `nightly` for some reason in the future.

.github/workflows/cd.yml
```diff
@@ -142,11 +130,9 @@ jobs:
               --allow-unauthenticated musl-tools
           fi
       - name: Install Rust toolchain
-        uses: actions-rs/toolchain@v1
+        uses: dtolnay/rust-toolchain@stable
         with:
-          toolchain: ${{ matrix.build.TOOLCHAIN }}
-          target: ${{ matrix.build.TARGET }}
-          override: true
+          targets: ${{ matrix.build.TARGET }}
```
orhun: Do you know if this action also overrides the following commands?

Sadly there is not way to test this without creating a new release and I had many problems with it...

orhun (commented):
Thanks, looking great!

I'm not so sure about the changes in `cd.yml` though.

.github/workflows/cd.yml
joshka: With the change to use dtolnay/rust-toolchain, this variable is not used to do anything, so it makes the workflow confusing to have it specified. It's easy to add in the future if needed, but generally YAGNI.

If there is a per combination change that's used once though, a future change could instead do something like `if: ${{ matrix.build.TOOLCHAIN != '' }}` ... instead of repeating the toolchain for each item.

.github/workflows/cd.yml
joshka: yes, it overrides future commands that use cargo. I think it may not override the actions/cargo below (which I missed from this PR). From memory that installs rust instead of using the default rust.

.github/workflows/cd.yml
```diff
@@ -26,140 +26,93 @@ jobs:
...
-            }
-          - {
-              NAME: darwin-arm64,
-              OS: macos-14,
-              TOOLCHAIN: stable,
-              TARGET: aarch64-apple-darwin,
-              NPM_PUBLISH: true,
-              PYPI_PUBLISH: true,
-            }
+        include:
```
joshka, line 33: See https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixinclude

.github/workflows/cd.yml
```diff
@@ -168,122 +121,119 @@ jobs:
           OUT_DIR=release/completions/ cargo run --release --bin git-cliff-completions
           OUT_DIR=release/man/ cargo run --release --bin git-cliff-mangen
           for bin in 'git-cliff' 'git-cliff-completions' 'git-cliff-mangen'; do
-            if [ "${{ matrix.build.OS }}" = "windows-2022" ]; then
+            if [ "${{ contains(matrix.os, 'windows') }}" = "true" ]; then
```
joshka, line 124: This can be further simplified and deduplicated by splitting the os and runner in the matrix above, either by adding a `runner` variable, or by adding an `is_windows` variable. I didn't worry about this as it's clear enough here.

There's possibly also a bash test that checks for `true` instead of string comparisonI but I haven't looked.

.github/workflows/cd.yml
```diff
@@ -26,140 +26,93 @@ jobs:
...
-          toolchain: ${{ matrix.build.TOOLCHAIN }}
-          target: ${{ matrix.build.TARGET }}
-          override: true
-      - name: Build (linux/macos)
-        if: matrix.build.OS != 'windows-2022'
-        uses: actions-rs/cargo@v1
-        with:
-          use-cross: true
-          command: build
-          args: --release --locked --target ${{ matrix.build.TARGET }}
```
joshka, line 156: I think cross is no longer necessary because the arm stuff is being built on the right runner now.

I'd suggest doing a quick check to validate this by turning off all the release parts of this change in another PR and just checking that all the build succeeds.

.github/workflows/cd.yml
```diff
@@ -26,140 +26,93 @@ jobs:
...
+            target: i686-unknown-linux-gnu
+            publish_npm: false
+            publish_pypi: false
+          - name: linux-x86-musl
+            os: ubuntu-latest
+            target: i686-unknown-linux-musl
+            publish_npm: false
+            publish_pypi: true
+            install_musl: true
+          - name: linux-arm64-glibc
```
joshka, line 56: See https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#standard-github-hosted-runners-for-public-repositories for available runners

.github/workflows/cd.yml
```diff
@@ -168,122 +121,119 @@ jobs:
...
-          file: git-cliff-${{ env.RELEASE_VERSION }}-${{ matrix.build.TARGET }}*
+          file: git-cliff-${{ env.RELEASE_VERSION }}-${{ matrix.target }}*
           file_glob: true
           overwrite: true
           tag: ${{ github.ref }}
           release_name: "Release v${{ env.RELEASE_VERSION }}"
           body: "${{ needs.generate-changelog.outputs.release_body }}"
       - name: Publish to GitHub (pre-release)
-        if: ${{ contains(github.ref, '-') }}
+        if: contains(github.ref, '-')
```
joshka, line 163: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsif
>When you use expressions in an if conditional, you can, optionally, omit the ${{ }} expression syntax because GitHub Actions automatically evaluates the if conditional as an expression. However, this exception does not apply everywhere.

.github/workflows/cd.yml
joshka: in the followup commit, I removed the actions-rs/cargo usage

.github/workflows/cd.yml
orhun: We can still use it in `dtolnay/rust-toolchain` with an extra parameter AFAIK though.

.github/workflows/cd.yml
orhun: I'm not so sure about not needing `cross`. This was extremely fragile to put together and testing it takes a lot of time.

joshka: If the build commands run without cross, why would cross be needed?

.github/workflows/cd.yml
joshka: Yes but why? all the targets use stable. If you need to do something different, then add it then. Do you have an anticipated use case for where nightly might be necessary?

.github/workflows/cd.yml
joshka: Checking this in https://github.com/orhun/git-cliff/pull/1198

orhun: yeah I guess we can backport those changes from that PR

.github/workflows/cd.yml
orhun: that's a cool shorthand

.github/workflows/cd.yml
orhun: Yeah, sometimes some deps magically do not work with stable Rust in the CI (it happened before).

It happens during the release so I'd prefer if we can do that now instead of hastily doing that after realizing the release is broken or something.

Sorry if I'm a bit pushy, I apparently have PTSD from one of the past releases lol

PR #1198, opened 2025-07-01
ci: check builds without cross

joshka:
Followup to https://github.com/orhun/git-cliff/pull/1193#discussion_r2176219304

This is a do not merge PR, it's just there to check whether cross is necessary for any of the build targets.


joshka:
Ok, so it looks like this works. The changes to the other PR are:
- add a concurrency check to make it possible to iterate on the PR. For release triggered PRs this is likely unnecessary
- set rust-toolchain.toml to use stable, so that running cargo doesn't choose the wrong thing
- install the 32 bit tools for x86 target (gcc-multilib g++-multilib libc6-dev-i386). Not sure if libc6 is needed, probably not
- Make x86 musl use clang-19 explicitly to build instead of musl-gcc

All of this seems pretty logical. So cross isn't necessary to build all the targets that you're building here.

It's likely the matrix could be simplified a fair bit easily.

joshka:
TODO - back merge the changes in this into the other PR

orhun:
So the verdict is we don't need cross I guess 👀

Issue #1093, opened 2025-03-13
^ in commit_parsers with skip_lines not working


erikthi:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description of the bug

I am trying to generate a changelog for unconventional commits. The commits may potentially contain more than one change, and I would like to use the commit parsers to grab each relevant part and make a separate entry for them in the changelog. 

### Steps To Reproduce

Configuration like this
```
{% else %}\
    ## [unreleased]
{% endif %}\
{% for group, commits in commits | group_by(attribute="group") %}
    ### {{ group | striptags | trim | upper_first }}
    {% for commit in commits %}
        - {% if commit.scope %}*({{ commit.scope }})* {% endif %}\
            {% if commit.breaking %}[**breaking**] {% endif %}\
            {{ commit.message | upper_first }}\
    {% endfor %}
{% endfor %}\n
"""

footer = """
<!-- generated by git-cliff -->
"""
# remove the leading and trailing s
trim = true

[git]
conventional_commits = false
filter_unconventional = false
split_commits = true
commit_parsers = [
  { message = "^fix:", group = "Fixed" },
]
filter_commits = true
topo_order = false
sort_commits = "oldest"

```
```

cd $(mktemp -d)
git init .
git commit --allow-empty -m "Unconventional commit" -m "fix: first line of the commit body"
git cliff
```

This yields an empty changelog. 

### Expected behavior

I would the changelog to contain:
### Fixed

- Fix: first line of the commit body

This is also what happens if I change 

```
commit_parsers = [
  { message = "^fix:", group = "Fixed" },
]
```
to
```
commit_parsers = [
  { message = "fix:", group = "Fixed" },
]
```

### Software information

- Operating system: Mac OS 14.6.1
- Rust version: 1.85.0
- Project version: Tested on both 2.5.0 and 2.8.0

orhun:
Hello, thanks for reporting! 👋🏼 

In your case, the generated commit looks like:

```
    Unconventional commit

    fix: first line of the commit body
```

1. `^fix:` won't catch this, since the `message` does not start with "fix"
2. `fix:`: this won't also do it.
3. You need to use `.*fix` to match on the whole commit message.

```toml
[git]
commit_parsers = [
  { message = ".*fix:", group = "Fixed" },
]
```

P.S. you can't match on specific fields of the commit since it's unconventional + conventional commit parsing is disabled (`conventional_commits=false`).

erikthi:
[...]

PR #1201, opened 2025-07-03
feat(changelog): implement commit processing order as set of steps

dslemusp:
## Description

Currently each commit in the commit list is processed before splitting. This causes an issue when `split_commit` is set to true. Commits in the description are not being processed if the commit message contains a skipped type set in the `commit_parsers`. For instance if the output of a squash and merge start with a `chore:` and the `chore:` is set to skip in the `commit_parsers` the description won't be parsed even if it contains `feat:` or `fix:` or any other non skipped field.

```
// Squash and merge Commit Sample
chore(ci): Update ci runners

fix(ci): a fix --> This is never parsed
chore(ci): another chore --> This is never parsed
```

## Motivation and Context

When working with squash and merge it is useful to have the message of the commit as the overview of the changes and the description as the details of the changes. However it should be possible to skip parsing the message as the description of the commit contains all the info, so trying to avoid repetitions.

## How Has This Been Tested?

I have adjusted the current unit test to reflect the scenario that I am describing

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/git-cliff-core/src/changelog.rs
+++ b/git-cliff-core/src/changelog.rs
@@ -204,9 +204,9 @@ impl<'a> Changelog<'a> {
             .filter_map(|commit| {
                 let mut commit_into_conventional = Ok(commit.clone());
                 if git_config.conventional_commits {
-                    if !git_config.require_conventional
-                        && git_config.filter_unconventional
-                        && !git_config.split_commits
+                    if !git_config.require_conventional &&
+                        git_config.filter_unconventional &&
+                        !git_config.split_commits
                     {
                         commit_into_conventional = commit.clone().into_conventional();
                     } else if let Ok(conv_commit) = commit.clone().into_conventional() {
@@ -397,8 +397,8 @@ impl<'a> Changelog<'a> {
                 release
                     .previous
                     .as_ref()
-                    .and_then(|release| release.version.as_ref())
-                    == Some(skipped_tag)
+                    .and_then(|release| release.version.as_ref()) ==
+                    Some(skipped_tag)
             }) {
                 if let Some(previous_release) = self.releases.get_mut(release_index + 1) {
                     previous_release.previous = None;
@@ -427,12 +427,10 @@ impl<'a> Changelog<'a> {
     #[cfg(feature = "github")]
     fn get_github_metadata(&self, ref_name: Option<&str>) -> Result<crate::remote::RemoteMetadata> {
         use crate::remote::github;
-        if self.config.remote.github.is_custom
-            || self
-                .body_template
-                .contains_variable(github::TEMPLATE_VARIABLES)
-            || self
-                .footer_template
+        if self.config.remote.github.is_custom ||
+            self.body_template
+                .contains_variable(github::TEMPLATE_VARIABLES) ||
+            self.footer_template
                 .as_ref()
                 .is_some_and(|v| v.contains_variable(github::TEMPLATE_VARIABLES))
         {
@@ -479,12 +477,10 @@ impl<'a> Changelog<'a> {
     #[cfg(feature = "gitlab")]
     fn get_gitlab_metadata(&self, ref_name: Option<&str>) -> Result<crate::remote::RemoteMetadata> {
         use crate::remote::gitlab;
-        if self.config.remote.gitlab.is_custom
-            || self
-                .body_template
-                .contains_variable(gitlab::TEMPLATE_VARIABLES)
-            || self
-                .footer_template
+        if self.config.remote.gitlab.is_custom ||
+            self.body_template
+                .contains_variable(gitlab::TEMPLATE_VARIABLES) ||
+            self.footer_template
                 .as_ref()
                 .is_some_and(|v| v.contains_variable(gitlab::TEMPLATE_VARIABLES))
         {
@@ -538,12 +534,10 @@ impl<'a> Changelog<'a> {
     #[cfg(feature = "gitea")]
     fn get_gitea_metadata(&self, ref_name: Option<&str>) -> Result<crate::remote::RemoteMetadata> {
         use crate::remote::gitea;
-        if self.config.remote.gitea.is_custom
-            || self
-                .body_template
-                .contains_variable(gitea::TEMPLATE_VARIABLES)
-            || self
-                .footer_template
+        if self.config.remote.gitea.is_custom ||
+            self.body_template
+                .contains_variable(gitea::TEMPLATE_VARIABLES) ||
+            self.footer_template
                 .as_ref()
                 .is_some_and(|v| v.contains_variable(gitea::TEMPLATE_VARIABLES))
         {
@@ -593,12 +587,10 @@ impl<'a> Changelog<'a> {
         ref_name: Option<&str>,
     ) -> Result<crate::remote::RemoteMetadata> {
         use crate::remote::bitbucket;
-        if self.config.remote.bitbucket.is_custom
-            || self
-                .body_template
-                .contains_variable(bitbucket::TEMPLATE_VARIABLES)
-            || self
-                .footer_template
+        if self.config.remote.bitbucket.is_custom ||
+            self.body_template
+                .contains_variable(bitbucket::TEMPLATE_VARIABLES) ||
+            self.footer_template
                 .as_ref()
                 .is_some_and(|v| v.contains_variable(bitbucket::TEMPLATE_VARIABLES))
         {
@@ -646,12 +638,10 @@ impl<'a> Changelog<'a> {
         ref_name: Option<&str>,
     ) -> Result<crate::remote::RemoteMetadata> {
         use crate::remote::azure_devops;
-        if self.config.remote.azure_devops.is_custom
-            || self
-                .body_template
-                .contains_variable(azure_devops::TEMPLATE_VARIABLES)
-            || self
-                .footer_template
+        if self.config.remote.azure_devops.is_custom ||
+            self.body_template
+                .contains_variable(azure_devops::TEMPLATE_VARIABLES) ||
+            self.footer_template
                 .as_ref()
                 .is_some_and(|v| v.contains_variable(azure_devops::TEMPLATE_VARIABLES))
         {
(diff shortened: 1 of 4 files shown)


orhun:
This was changed in #556 to fix another issue and was reported again in #1093 - can you maybe take a look and double check this?

dslemusp:
Yeah actually it seems to me that the initial behaviour and order is what I would expect to have. As I said, right now the commits are treated as one during preprocessing even if the `split_commit` flag is enabled. This results on description being skipped because the commit message is ignored in the preprocessor.

orhun:
Thanks for checking. I'd lean towards making this configurable instead of changing the behavior again. We could maybe add a new `processing_order` option like follows:

```toml
[git]
processing_order = ["preprocess", "split", "proces", "postprocess", "etc"]
```

So that you can configure the behavior. What do you think?

dslemusp:
I like it!

dslemusp:
How many processing steps do we have?. I see `split`, `preprocess`, `commit_parsing`, `link_parsing`, `into_conventional`

ognis1205:
@dslemusp @orhun 

I assume the current step names like `preprocess`, `process`, and `postprocess` are just temporary placeholders for now, since this is still in a draft stage. That said, if we move forward with making the order configurable, I think it’d be good to avoid names like those — they imply a fixed sequence, which might be misleading if the steps can be reordered. It’d be great if we could find names that reflect the actual behavior more clearly.

dslemusp:
I agree. I would suggest we use similar (if not same) names as they are defined in the cliff.toml (e.g. `commit_preprocessor` instead of `preprocess` ). I am working on updating the PR but it might take some time!. Feel free to think about the naming in the mean time ;P

orhun:
> I would suggest we use similar (if not same) names as they are defined in the cliff.toml

Yes, very good point. I'd like that 👍🏼 

> I am working on updating the PR but it might take some time!

Great, thank you! Let me know if you need any help/guidance in the meantime. I think we should also update the title of this PR accordingly :D

dslemusp:
Ok I found some time today I came up with this. Maybe there is still some cleanup to do as this makes https://github.com/orhun/git-cliff/blob/c156fc5edb405737b62a802ed9b1e91f9789ef9b/git-cliff-core/src/commit.rs#L208 not needed.

dslemusp:
@orhun Any follow up on this?. I would like to have this functionality merged

git-cliff-core/src/changelog.rs
```diff
@@ -128,31 +112,141 @@ impl<'a> Changelog<'a> {
...
+                ProcessingStep::SplitCommits => {
+                    debug!("Splitting commits...");
+                    if git_config.split_commits {
+                        *commits = Self::apply_split_commits(commits);
+                    } else {
+                        debug!("Split commits is disabled, skipping...");
+                    }
+                }
+            }
+        }
```
dslemusp: A thing to keep in mind is that if all the steps are not listed in the `processing_order` they will be skipped. So maybe we need to check that the list always contains all the steps (no matter the order)

git-cliff-core/src/config.rs
```diff
@@ -82,9 +82,54 @@ pub struct ChangelogConfig {
     pub output: Option<PathBuf>,
 }
 
+/// Processing steps for the commits
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum ProcessingStep {
```
ognis1205, line 88: Suggestions:

1.	Make the enum variant names and config keys consistent and explicit about what the slot does. Prefer either explicit Verb + Object action names or neutral stage names so the intent is clear. Example mappings:

 - `CommitParsers` -> `ParseCommits`
 - `CommitPreprocessors` -> `ModifyCommits`
 - `IntoConventional` -> `ConventionalizeCommits` or `NormalizeCommits`
 - `LinkParsers` -> `ParseLinks`
 - `SplitCommits` -> `SplitCommits`

2.	Rename `CommitPreprocessors` to a name that reflects it being an ordering slot rather than implying it always runs "before" something. Options:

 - If message transformations/modification are the main behavior: `TransformCommits` or `ModifyCommits`.

ognis1205 (commented):
Another important point: since there are already issues arising from the interactions of repository and directory-related configs, I think we should be very careful about introducing new functionality into the config. Both the necessity and the exact specification should be carefully evaluated before adding more to this area.

git-cliff-core/src/config.rs
dslemusp: Thx for the suggestions, I think it is indeed nice to have consistent naming. However I think the naming is more a fundamental change as I am using the current naming definitions from the configuration (see https://git-cliff.org/docs/configuration/git). I can indeed modify the names with the caveat of using multiple names for the same (e.g. `ModifyCommits` in place of `CommitPreprocessors` for this step https://git-cliff.org/docs/configuration/git#commit_preprocessors).
It seems that @orhun agreed on using the same names as in the config.toml (https://github.com/orhun/git-cliff/pull/1201#issuecomment-3063303337).

dslemusp:
> Another important point: since there are already issues arising from the interactions of repository and directory-related configs, I think we should be very careful about introducing new functionality into the config. Both the necessity and the exact specification should be carefully evaluated before adding more to this area.

I tend to agree, but in this case the idea is to keep feature parity and backwards comparability in case the `processing_steps` field is not present in the cliff.toml
When I have time I refactor the tests to leave them as they were on the base and add new ones for the feature. This to also check in a way that there are no regressions.

git-cliff-core/src/config.rs
orhun: Yup, I think it makes more sense to use the names that we are already familiar.

orhun:
Sorry for the delay on this... but most things look good already. Just need to find some time to do proper review.

Pretty excited for this one

Btw can you fix the conflicts with `main`?

orhun:
going to include this in 2.12.0

dslemusp:
I have rebased and reverted the test to assert the original results. However I cannot manage to make that the parser skip the commit with the footer. I check inside the parser and both in `main` and in my implementation the parser does not skip the commit. Still trying to figure it out. Any help understandig the issue is appreciated

ognis1205:
> I check inside the parser and both in `main` and in my implementation the parser does not skip the commit.

Just to clarify:

When you say it doesn't skip the commit in `main` either, do you mean tests are failing on the `main` branch?
Locally, with the latest `main`, all tests seem to pass for me. Or is the idea that the current tests don't cover this case properly?

git-cliff-core/src/changelog.rs
```diff
@@ -90,132 +90,16 @@ impl<'a> Changelog<'a> {
...
-            if let Some(commit) = Self::process_commit(commit, git_config, summary) {
-                if git_config.split_commits {
-                    for line in commit.message.lines() {
-                        let mut c = commit.clone();
-                        c.message = line.to_string();
-                        c.links.clear();
-                        if c.message.is_empty() {
-                            continue;
-                        }
-                        if let Some(c) = Self::process_commit(&c, git_config, summary) {
```
ognis1205, line 187: > Still trying to figure it out. Any help understandig the issue is appreciated

I think you're mainly referring to failures in `changelog::test::changelog_generator_split_commits`.

The changes in this PR introduce a breaking change even when the default commit processing order is used.

I haven't traced the logs in full detail, but in the previous implementation, after commit splitting, each split commit message still went through `Commit::process_commit`, which in turn calls things like `Commit::preprocess` and `Commit::into_conventional_commit`.

In the current implementation in this PR, after preprocess is executed, the corresponding processing step doesn't seem to be applied.

This would likely explain the following diff:

```sh
 #### other
 - support unconventional commits
 - this commit is preprocessed
>- use footer
>- footer text
 - make awesome stuff look better
 
 #### ui
 - make good stuff
 
 ### Commit Statistics
 
<- 18 commit(s) contributed to the release.
<- 12 day(s) passed between the first and last commit.
<- 17 commit(s) parsed as conventional.
>- 20 commit(s) contributed to the release.
>- 13 day(s) passed between the first and last commit.
>- 19 commit(s) parsed as conventional.
```

orhun:
>  However I cannot manage to make that the parser skip the commit with the footer. I check inside the parser and both in main and in my implementation the parser does not skip the commit. 

Can you elaborate this a bit? Are you talking about the tests or the actual implementation?

> The changes in this PR introduce a breaking change even when the default commit processing order is used.

I think this feature should cover our existing behavior and tests as default. So I would expect it to be compatible or at least configurable to satisfy the test fixtures.

Btw I had a quick look to review this but realized there are some leftover commented out code pieces and println statements in the code. Should we mark this as draft until it's ready? We should update the docs/website about this as well.

git-cliff-core/src/changelog.rs
dslemusp: Correct I am talking about the `changelog::test::changelog_generator_split_commits`. My approach is to make the feature backwards compatible such that the current tests should also pass.

git-cliff-core/src/changelog.rs
```diff
@@ -1469,6 +1355,13 @@ mod test {
     #[test]
     fn changelog_generator_split_commits() -> Result<()> {
         let (mut config, mut releases) = get_test_data();
+        config.git.processing_order.order = vec![
+            ProcessingStep::CommitPreprocessors,
+            ProcessingStep::SplitCommits,
+            ProcessingStep::IntoConventional,
+            ProcessingStep::CommitParsers,
+            ProcessingStep::LinkParsers,
+        ];
```
ognis1205, line 1364:
```suggestion
        config.git.processing_order.order = vec![
            ProcessingStep::CommitPreprocessors,
            ProcessingStep::IntoConventional,
            ProcessingStep::CommitParsers,
            ProcessingStep::SplitCommits,
            ProcessingStep::IntoConventional,
            ProcessingStep::CommitParsers,
            ProcessingStep::LinkParsers,
        ];
```

I've identified the part that needs to be adjusted in order to make the existing unit tests pass.

In the previous implementation (depending on the configuration), the processing flow inside `Changelog::process_commit` -> `Commit::process` was:

`preprocess` -> `into_conventional` -> `parse`

However, in this PR the flow becomes:

`preprocess` -> `parse`

Since `into_conventional` is no longer invoked, the footer regex matcher is not executed. Without calling `into_conventional`, the `footer` is never parsed.

ognis1205 (changes requested):
With a small adjustment we can make the current tests pass again. However, the existing commit processing order is not strictly linear, it changes depending on configuration and context. Because of that, I'm not yet fully confident that this patch alone is sufficient or entirely correct from a behavioral standpoint.

ognis1205:
Thanks again for the work on this PR!

Just checking in. Are you still planning to continue working on it? Since it's a fairly large change, it would be good to know how you'd like to proceed.

If it helps, we could also co-author the remaining work or take it over to move things forward.

orhun:
#1485

PR #1226, opened 2025-07-27
docs: mention the side-bar

Lewiscowles1986:
## Description

Adds some blurb to the main installation page, as it is the default navigated page from the docs if I click installation.

## Motivation and Context

I Cannot be the only person that thought "why is there a blank installation page". The other pages are really good too!

I'm totally cool if this cannot be merged, but I was really expecting cargo install or install binaries to be on this page, so it caused a momentary disconnect that I had to look for where to find the information, instead of it being right there for me.

A list of links might be nice, but; for-now, this would at least allow continuous flow.

## How Has This Been Tested?

I just wrote text in a docs page.

Just what the page looks like for me when I visited.

Original installation page

## Types of Changes

- [X] Documentation (no code change)

## Checklist:

### DIFF
--- a/website/docs/installation/index.md
+++ b/website/docs/installation/index.md
@@ -3,6 +3,8 @@ sidebar_position: 2
 ---
 # Installation
 
+Please see the side-bar to navigate to a specific installation mechanism. This page has information about the packaging status from repo-ology.
+
 <details>
   <summary>Packaging status</summary>
 


Lewiscowles1986:
While I could change https://github.com/orhun/git-cliff/actions/runs/16552784738/job/46810035103; I don't think it really belongs in this patch set.

website/docs/installation/index.md
```diff
@@ -3,6 +3,8 @@ sidebar_position: 2
 ---
 # Installation
 
+Please see the side-bar to navigate to a specific installation mechanism. This page has information about the packaging status from repo-ology.
```
orhun, line 6:
```suggestion
Please see the side-bar to navigate to a specific installation mechanism. This page has information about the packaging status from repology.
```

PR #1249, opened 2025-09-09
chore(examples): improve Github templates

tessus:
## Description

The `New contributors` section should be within (under) the release section, but they were on the same level (`##`).
This change adjusts the heading level to `###`.

There was no empty line between the list of new contributors and the next release.
There should also be an empty line before the footer. This change takes care of both of these cases.

## Motivation and Context

see https://github.com/orhun/systeroid/pull/199#issuecomment-3269307935

## How Has This Been Tested?

cargo test
manual test

## Types of Changes

- [x] Other: depending on the point of view it can be a fix or an improvement ;-)

## Checklist:


examples/github-keepachangelog.toml
```diff
@@ -45,6 +45,10 @@ body = """
       [#{{ contributor.pr_number }}]({{ self::remote_url() }}/pull/{{ contributor.pr_number }}) \
     {%- endif %}
 {%- endfor %}\n
+
+{%- if github.contributors | filter(attribute="is_first_time", value=true) | length != 0 %}
+{% endif %}
+
```
orhun: I left a comment about this in another PR.

Can we leave this out for now and just do the header changes here?

tessus: unfortunately now the replies are also spread across 2 other PRs...

> Can we leave this out for now and just do the header changes here?

sure. without it, there won't be an empty line between the body or the footer, neither will there be one after the new contributors section.

Those are just cosmetic changes anyway. They are not needed for the markdown to be rendered correctly.

It's just annoying when you read the raw text, because it's less legible.

I can remove these 4 lines.

tessus: argh!!! I committed and pushed, but guess what.... the file was not staged. ;-) 🤔 
hmmm. I guess I must have been somewhere else. sorry about that.

3 days later.... here it is.

tessus:
Btw, something else, but I thought I can still ask here in the PR. (unless you want me to open an issue)

I've noticed that git-cliff ignores the `XDG_CONFIG_HOME` env var on macOS and only considers `Library/Application Support` as the config dir, which is primarily used by GUI apps.
TBH, cli apps should still use the `~/.config` directory on macOS unless `XDG_CONFIG_HOME` is set.
I tried to open an issue in the `dirs` crate repo, but it has been archived.
I also checked `dirs_next`, which is used in rustypaste-cli, but that one's dead too.
Not sure what going on, but the dir crates seem all unmaintained.

Either way, I was wondering whether we could make a similar change here in git-cliff as in rustypaste-cli: honor `XDG_CONFIG_HOME` and fallback to `~/.config`.

orhun:
ok, here is the thing: we also use this in the other templates so we should update them too, if necessary (just search for `## New Contributors` in the codebase)

Also the CI is failing because the fixture result has changed. You need to update the `expected.md` for the failed fixture I guess.

Sorry that it took so long. (excuse: I'm in China)

orhun:
> Either way, I was wondering whether we could make a similar change here in git-cliff as in rustypaste-cli: honor XDG_CONFIG_HOME and fallback to ~/.config.

Yeah, reasonable! Please do it!

tessus:
>  we should update them too

will do!

> Also the CI is failing because the fixture result has changed. You need to update the expected.md for the failed fixture I guess.

That's a bit weird, since the fixtures use their own templates. Not sure what's going on. I'll have to investigate.

> Sorry that it took so long. (excuse: I'm in China)

No worries, my current sleep cycle is completely fucked up, so my responses also might take a while. e.g. I'm going to bed soon, so I will look into it in around 14h when I take a break from work...

Man, have fun in China!

> Yeah, reasonable! Please do it!

I'll create a separate PR.

orhun:
> That's a bit weird, since the fixtures use their own templates. Not sure what's going on. I'll have to investigate.

Hmm I think you are right... there must be another issue there. I realized we are hitting rate limits in CI, maybe that's why? (simply setting `GITHUB_TOKEN` would fix that)

> No worries, my current sleep cycle is completely fucked up, so my responses also might take a while. e.g. I'm going to bed soon, so I will look into it in around 14h when I take a break from work...

No worries/rush!

<sup>
Man... sleep is really important. I recently started to get advice from ChatGPT about beating jetlag. New Zealand was pretty rough because of that...
</sup>

> Man, have fun in China!

Thanks! It's been [great](https://fosstodon.org/@orhun/115219549622080266). Now I'm omw to Ankara.

tessus:
> we should update them too

Done, but it was only on other file: `github.toml`

While editing this file, something else came to mind. What do you think of the following?

```diff
diff --git a/examples/github.toml b/examples/github.toml
index fa95719..386fe0d 100644
--- a/examples/github.toml
+++ b/examples/github.toml
@@ -12,7 +12,7 @@
 body = """
 ## What's Changed

-{%- if version %} in {{ version }}{%- endif -%}
+{%- if version %} in {{ version | trim_start_matches(pat="v") }}{%- endif -%}
 {% for commit in commits %}
   {% if commit.remote.pr_title -%}
     {%- set commit_message = commit.remote.pr_title -%}
```

For user facing version representation the `v` can always be stripped in my opinion. 

> I realized we are hitting rate limits in CI, maybe that's why? (simply setting GITHUB_TOKEN would fix that)

👍

tessus:
Hmm, maybe I have to propagate the value, e.g. calling the fixtures action with an argument.

P.S.: No need for another parameter in the fixture-action. It seems that gh propagates env vars (which makes this less work. ;-)).

tessus:
oops, I've added your fix to my fix (the one you asked mw to remove). I had too many terminal sessions open.
Please let me know, if I should remove 6e5942b28f56cb9f8213ed476193093c57ba2d85

orhun:
> While editing this file, something else came to mind. What do you think of the following?

Yeah, that would be nice I think.

orhun (approved):
LGTM

We can also do the `v`-stripping and rename this PR to "improve GitHub templates" or something

tessus:
what about the 3 dots in the compare links?

orhun:
yeah that too

orhun:
nice try diddy

PR #1184, opened 2025-06-20
style: update formatting rules

joshka:
chore: update formatting rules

Use spaces instead of tabs for indentation, wrap at 100 characters, and
remove the setting that aligns struct fields vertically as this
generally harms readability.

These changes make the code more consistent with the Rust community's
conventions and improve readability by not truncating lines too early.

---
Justification:

https://www.youtube.com/watch\?v\=V7PLxL8jIl8

This brings things more in line with pretty much every rust project out there. Hard tabs are used in around 0.1% of rust projects:
- `hard_tabs = true`: https://github.com/search?q=path%3A%22rustfmt.toml%22+%22hard_tabs+%3D+true%22+AND+%28NOT+is%3Afork%29&type=code (1.9k results)
- projects with a rustfmt.toml: https://github.com/search?q=path%3A%22rustfmt.toml%22+AND+%28NOT+is%3Afork%29&type=code (14.6k results)
- rust projects: https://github.com/search?q=path%3A%2F%28%5E%7C%5C%2F%29cargo%5C.toml%24%2F+AND+%28NOT+is%3Afork%29&type=code (1.4M results)

PR #1259, opened 2025-09-18
feat(cli): honor XDG_CONFIG_HOME on macOS

tessus:
## Description

git-cliff ignores the `XDG_CONFIG_HOME` env var on macOS and only considers `Library/Application Support` as the config dir, which is primarily used by GUI apps.

This change allows the default config file to be placed in the following 3 locations on macOS:

- `XDG_CONFIG_HOME/git-cliff/cliff.toml`
- `~/.config/git-cliff/cliff.toml`
- `~/Library/Application Support/git-cliff/cliff.toml`

The first one found is used.

## Motivation and Context
see https://github.com/orhun/git-cliff/pull/1249#issuecomment-3301875097

## How Has This Been Tested?

manually and with debug output

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/git-cliff-core/src/config.rs
+++ b/git-cliff-core/src/config.rs
@@ -492,7 +492,6 @@ impl Config {
     ///
     /// If the config file is not found in its standard locations, [`None`] is returned.
     pub fn retrieve_config_path() -> Option<PathBuf> {
-        let mut path: Option<PathBuf> = None;
         for supported_path in [
             #[cfg(target_os = "macos")]
             Some(Config::retrieve_xdg_config_on_macos().join(DEFAULT_CONFIG)),
@@ -502,12 +501,11 @@ impl Config {
         .filter_map(|v| v.as_ref())
         {
             if supported_path.exists() {
-                path = Some(supported_path.to_path_buf());
                 debug!("Using configuration file from: {:?}", supported_path);
-                break;
+                return Some(supported_path.to_path_buf());
             }
         }
-        path
+        None
     }
 }
 
--- a/git-cliff-core/src/config.rs
+++ b/git-cliff-core/src/config.rs
@@ -1,4 +1,3 @@
-use std::env;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::{fmt, fs};
@@ -482,7 +481,7 @@ impl Config {
     /// If it is not set, it will fall back to `~/.config`
     #[cfg(target_os = "macos")]
     pub fn retrieve_xdg_config_on_macos() -> PathBuf {
-        let config_dir = env::var("XDG_CONFIG_HOME").map_or_else(
+        let config_dir = std::env::var("XDG_CONFIG_HOME").map_or_else(
             |_| dirs::home_dir().unwrap_or_default().join(".config"),
             PathBuf::from,
         );
--- a/git-cliff-core/src/config.rs
+++ b/git-cliff-core/src/config.rs
@@ -1,4 +1,3 @@
-#[cfg(target_os = "macos")]
 use std::env;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
@@ -12,6 +11,8 @@ use serde::{Deserialize, Serialize};
 use crate::embed::EmbeddedConfig;
 use crate::error::Result;
 use crate::{command, error};
+
+use crate::DEFAULT_CONFIG;
 
 /// Default initial tag.
 const DEFAULT_INITIAL_TAG: &str = "0.1.0";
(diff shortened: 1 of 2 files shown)


git-cliff/src/lib.rs
```diff
@@ -525,11 +525,20 @@ pub fn run_with_changelog_modifier(
     // Set path for the configuration file.
     let mut path = args.config.clone();
     if !path.exists() {
-        if let Some(config_path) =
-            dirs::config_dir().map(|dir| dir.join(env!("CARGO_PKG_NAME")).join(DEFAULT_CONFIG))
+        for possible_path in [
```
orhun: `possible_path` -> `supported_path`

git-cliff/src/lib.rs
orhun, line 1: Maybe it would be better if we extract this into a new function (guarded by `#[cfg(target_os = "macos")]` and `#[cfg(not(target_os = "macos"))]` instead of further complicating the logic.

I really don't like how big `lib.rs` got over time. I need to refactor that sometime.

git-cliff-core/src/config.rs
```diff
@@ -1,3 +1,5 @@
+#[cfg(target_os = "macos")]
+use std::env;
```
orhun: You could also import this inside of the function. I think that would be cleaner.

git-cliff/src/lib.rs
tessus: hmmm, I can look into this. Will do later this evening. I still need the previous logic on macOS as well, which is why I created the array and loop over it.

tessus: However I look at this, the only real option is to create a new function in config.rs, similar to `retrieve_xdg_config_on_macos`.

This would hide the code in lib.rs.

However, the logic in the function wouldn't change, since I believe the current one is already the most succint.

Shall I move the code to a function `retrieve_config_dir`? However, the new code in lib.rs is not that much in the first place. It's only spread over a few lines due to the formating.
If I hide the code in a function it would have to return `Option<PathBuf>`, which then requires another construct in lib.rs that won't be on one line either.

tessus: I've now created a separate function as you suggested. Not sure, if it is better than before, but the `lib.rs` is certainly a few lines shorter. ;-)

tessus:
I don't have rust nightly installed, thus I cannot format the code.

It's the first time I have seen that a project uses nightly to check the formatting. Isn't this a bit fragile? Either way, I can install nightly later today or maybe you can format it if you have it installed.

git-cliff/src/lib.rs
orhun: lemme take a look

git-cliff-core/src/config.rs
```diff
@@ -469,6 +471,44 @@ impl Config {
...
+        let config_dir = std::env::var("XDG_CONFIG_HOME").map_or_else(
+            |_| dirs::home_dir().unwrap_or_default().join(".config"),
+            PathBuf::from,
+        );
+        config_dir.join("git-cliff")
+    }
+
+    /// Find the path of the config file.
+    ///
+    /// If the config file is not found in its standard locations, None is returned.
```
orhun:
```suggestion
    /// If the config file is not found in its standard locations, [`None`] is returned.
```

git-cliff-core/src/config.rs
```diff
@@ -469,6 +471,44 @@ impl Config {
...
+    pub fn retrieve_config_path() -> Option<PathBuf> {
+        let mut path: Option<PathBuf> = None;
```
orhun: Instead of assigning to a variable, can you return `Some(T)` whenever a path is found? This way we can also get rid of `break` as well.

I'm not sure if early-return is idiomatic or not, but I'd like to have it here.

orhun (approved):
Looking pretty good! Just some minor points

orhun:
Yeah, you can format via `cargo +nightly fmt`. You don't need to switch to a nightly toolchain I believe.

orhun:
There is some discussion about nightly in #1184 

It's not that breaking, unless you're switching to a new edition btw.

git-cliff-core/src/config.rs
tessus: done

tessus:
> Yeah, you can format via cargo +nightly fmt. You don't need to switch to a nightly toolchain I believe.

Hmmm, nope:

```
$ cargo +nightly fmt
info: syncing channel updates for 'nightly-x86_64-apple-darwin'
info: latest update on 2025-09-19, rust version 1.92.0-nightly (7c275d09e 2025-09-18)
info: downloading component 'cargo'
info: downloading component 'clippy'
info: downloading component 'rust-docs'
info: downloading component 'rust-std'
info: downloading component 'rustc'
info: downloading component 'rustfmt'
info: installing component 'cargo'
info: installing component 'clippy'
info: installing component 'rust-docs'
info: installing component 'rust-std'
info: installing component 'rustc'
info: installing component 'rustfmt'
```

Anyway, when I ran it, much less was fixed than what was shown in the pipeline. Let's see, if it now runs through. 

P.S.: Switching is not necessary, but now it will download the nightly almost every time I use `cargo +nightly fmt` since nightly is released every day.

tessus:
Nice, the formatting seems ok. Issues with link validation are not related to this PR. So we should be good.

Issue #1274, opened 2025-10-01
fields for github remote not populated


joluizquierdo:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description of the bug

When running git-cliff and the environment variables 'GITHUB_TOKEN', 'GITHUB_REPO' are properly set, github context is not properly populated everything appear empty.

I checked if the token might be the issue, but the debug mode show that the API calls to Github are properly set, and it's actually retrieving the pull request information and commits for the api (see pictures below).

For further clarification: The branch in the tests have a few commits not released and all of those commits where merged in to the branch from a previous PR using squash strategy merge.

### Steps To Reproduce

1. Install git-cliff using cargo
2. Generate a Github token (classic) with full permissions over repo, workflow and write:discussion in grant access to the private ORG where the repository is hosted.
3. set environment variables 'GITHUB_TOKEN' and 'GITHUB_REPO'
4. run ` git-cliff --unreleased -v --context` and check for each commit 'github' field, it appears null and empty.

### Expected behavior

The context object should populate the github fields like 'pr_number' and 'username'.
Correctly.

### Screenshots / Logs

When running the command: `git-cliff --unreleased -v --context`

```text
DEBUG git_cliff_core::changelog > Adding remote data...
 DEBUG git_cliff_core::changelog > You are using an experimental feature! Please report bugs at <https://git-cliff.org/issues>
▸▹▹▹▹ Retrieving data from GitHub... (my-fake-org/my-fake-repo)                                                     DEBUG git_cliff_core::remote    > Sending request to: https://api.github.com/repos/my-fake-org/my-fake-repo/commits?per_page=100&page=0
 DEBUG git_cliff_core::remote    > Sending request to: https://api.github.com/repos/my-fake-org/my-fake-repo/commits?per_page=100&page=1
 DEBUG git_cliff_core::remote    > Sending request to: https://api.github.com/repos/my-fake-org/my-fake-repo/commits?per_page=100&page=2
 DEBUG git_cliff_core::remote    > Sending request to: https://api.github.com/repos/my-fake-org/my-fake-repo/commits?per_page=100&page=3
 DEBUG git_cliff_core::remote    > Sending request to: https://api.github.com/repos/my-fake-org/my-fake-repo/commits?per_page=100&page=4
 DEBUG git_cliff_core::remote    > Sending request to: https://api.github.com/repos/my-fake-org/my-fake-repo/commits?per_page=100&page=5
[...]

Issue #1264, opened 2025-09-21
Tracking issue: repository/directory handling


ognis1205:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description of the bug

In the snippet at: 

https://github.com/orhun/git-cliff/blob/3da8f2678b872569a2875e1ca05707bc26124c78/git-cliff/src/lib.rs#L254-L266

when there is a difference between `current_dir` and `repository.root_path()`, `include_path` is generated based on that difference.

According to `git2`’s specification for Repository::discover ([docs.rs](https://docs.rs/git2/latest/git2/struct.Repository.html#method.discover)), the repository root path is determined by walking up from the specified path until the first matching Git repository is found. This works for both bare and non-bare repositories, and in the case of submodules, the `gitdir` specified in the `.git` file is used.

Meanwhile, in `git-cliff`, the repository is set either via `--repository` or, as implemented at:

https://github.com/orhun/git-cliff/blob/3da8f2678b872569a2875e1ca05707bc26124c78/git-cliff/src/lib.rs#L706

`current_dir` is used. Moreover, the paths of files returned by `git2`:

https://github.com/orhun/git-cliff/blob/3da8f2678b872569a2875e1ca05707bc26124c78/git-cliff-core/src/repo.rs#L236

are relative to `repository.root_path()`.

Therefore, in a monorepository where `current_dir` is not itself a Git repository, the current implementation works correctly. However, in cases such as:

 - Generating a changelog inside a submodule
 - Generating changelogs for multiple repositories specified via `--repository`
 - When the `.git` directory is a symlink

extra `path_diff` entries may be introduced into `include_path`, resulting in no patterns matching any files.

As a countermeasure, the change in v2.8.0 ([commit ceda1f7](https://github.com/orhun/git-cliff/commit/ceda1f753adc3459b88417ae363410d2586a33af)) could be reverted. Instead of automatically including `current_dir` in monorepository environments, it would be better to remove this behavior and encourage users to explicitly specify `include_path`.

Related issues:

 - https://github.com/orhun/git-cliff/issues/1187
 - https://github.com/orhun/git-cliff/issues/1248
 - https://github.com/orhun/git-cliff/issues/1251
 - https://github.com/orhun/git-cliff/issues/1255

orhun:
It would be nice to add a small summary to this issue (regarding why it exists + what do we want to achieve)

ognis1205:
@orhun 

[...]

PR #1275, opened 2025-10-01
fix: Use current branch for GitHub API when no range is specified

joluizquierdo:
This ensures that unreleased commits on feature/develop branches are properly matched with their corresponding pull requests.

Fixes issue where GitHub metadata (PR numbers, titles) was not populated for unreleased commits on non-default branches.

Also fixed .gitignore to ignore target/ insteand of /target

## Description

When running git-cliff --unreleased without a specific range, it now uses
the current branch name for GitHub API calls instead of defaulting to the
repository's default branch.
## Motivation and Context

Fixing of the following issue: https://github.com/orhun/git-cliff/issues/1274

## How Has This Been Tested?

Tested against some repos in my company, where the error was discovered 

Please everything is detailed in the issue linked to the PR

## Types of Changes

- [ ] Bug fix (non-breaking change which fixes an issue). I don't know if this change might break other users usage....

## Checklist:


joluizquierdo:
Hello, like I remarked in the issue that I opened today, I have no prior expirience with rust, I'm more familiar with nodejs, python and other languages, I do know how to write code. 

Therefore, with some help of the AI, I was able to locate the fix and propose a extreamly clumpsy suggestion. 

I hope to at least point the error and save you time!

ognis1205:
cc: @orhun 

I haven't gone through this PR in detail yet, but it seems related to [#1264](https://github.com/orhun/git-cliff/issues/1264), so perhaps it could be tied together for a more comprehensive implementation. Still, this one looks fine to review as-is.

git-cliff-core/src/changelog.rs
```diff
@@ -464,30 +464,49 @@ impl<'a> Changelog<'a> {
...
         };
 
+        // If no specific ref is provided, use the current branch for remote API calls
+        let ref_name_for_remote = if ref_name.is_none() {
+            // Get current branch name
+            use std::process::Command;
+            let output = Command::new("git")
+                .args(&["rev-parse", "--abbrev-ref", "HEAD"])
+                .output();
+
```
orhun, line 474: We should avoid running `git` and use [`git2`](https://docs.rs/git2) instead :)

orhun (commented):
Thanks for the PR!

This looks mostly good and makes sense :) Left a comment.

orhun:
any progress on this?

PR #1287, opened 2025-10-10
feat(templating): Implemented split_regex, replace_regex, find_regex filters

linus-skold:
## Motivation and Context
I was making use of git-cliff for a project that was using a bit more unorthodox commit style with message and body that requires me to be able to do some more advanced parsing when I generate changelogs.

## How Has This Been Tested?

I wrote new test cases in `template.rs` but I have also run this on my own project changelogs on a project and successfully pulled all string sequences I was looking for out of the commit message to be able to process them in a way that I had to.
I mainly needed the `find_regex` function the split and replace is a bonus that came from testing different approaches.

The project i work on use Azure DevOps and we add a user story, task or such number to the Pull Request when we create them, there can be multiple ones as well, and since it sadly doesn't use conventional commits we ended up using the user story number as a group.
But sometimes a PR includes multiple user stories and then we want to pull all of those from the message with a regex as such `(#\\d+)` and then just write them all before the actual commit message that is included for a better overview.

So we could write a really simple template like this
```
body = """
{% if version %}\
    # [{{ version | trim_start_matches(pat="v") }}]
{% else %}\
    # [unreleased]
{% endif %}\
{% for group, commits in commits | group_by(attribute="group") %}
    {% if group == "Others" %}
        ## {{ group | striptags | trim }}\
    {% else %}
        {{ group | striptags | trim }}\
    {% endif %}
    {% for commit in commits %}
      {% set issues = commit.message | find_regex(pat="#\\d+")   %}\
      {% for issue in issues | slice(start=1) %}\
        {{issue}}\n
      {% endfor %}
      {% set message = commit.message | split(pat="\n") %}
      {% if message | length > 1 %}\
        - {{ message[0] | trim }}\
      {% endif %}
    {% endfor %}
{% endfor %}
"""
```

## Types of Changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation (no code change)
- [ ] Refactor (refactoring production code)
- [ ] Other 

## Checklist:

- [x] My code follows the code style of this project.
- [x] I have updated the documentation accordingly.
- [x] I have formatted the code with [rustfmt](https://github.com/rust-lang/rustfmt).
- [x] I checked the lints with [clippy](https://github.com/rust-lang/rust-clippy).
- [x] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #1296, opened 2025-10-27
feat: support context json schema dump

### LINKED ISSUE #1294: Provide JSON Schema of `Context` to allow strict type generation towards scripts

### Feature description

We are integrated Github with Linear to manager Projects cycle.

We should parse github pr metadata and generate related Linear issue from it, thus, a third-part script is required.

### Desired solution

We can use `schemars` to Generate related Schema automaticly, and post the schema to website or github repo.

### Alternatives considered

Nope

### Additional context

_No response_

greenhat616:
## Description

Close #1294. Add JSON Schema dump feature.

Current blocked by https://github.com/crate-ci/git-conventional/pull/88.

## Motivation and Context

Add a gloabl opt `--dump-context-schema` to dump current version `Context` JSON Schema.

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1006,7 +1006,7 @@ dependencies = [
  "dyn-clone",
  "expect-test",
  "futures",
- "git-conventional",
+ "git-conventional 0.12.9 (git+https://github.com/greenhat616/git-conventional)",
  "git2",
  "glob",
  "http-cache-reqwest",
@@ -1020,6 +1020,7 @@ dependencies = [
  "reqwest",
  "reqwest-middleware",
  "rust-embed",
+ "schemars",
  "secrecy",
  "semver",
  "serde",
@@ -1041,6 +1042,16 @@ version = "0.12.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f6a949b7fcc81df22526032dcddb006e78c8575e47b0e7ba57d9960570a57bc4"
 dependencies = [
+ "unicase",
+ "winnow",
+]
+
+[[package]]
+name = "git-conventional"
+version = "0.12.9"
+source = "git+https://github.com/greenhat616/git-conventional#af796e07661317c82a495b9ad4b5d7134f09a4d9"
+dependencies = [
+ "schemars",
  "serde",
  "unicase",
  "winnow",
@@ -1830,7 +1841,7 @@ version = "0.2.25"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2af0742157c04cea78f8643de0d0785a29d53c4dd08d985bc542cdd4d2ec9830"
 dependencies = [
- "git-conventional",
+ "git-conventional 0.12.9 (registry+https://github.com/rust-lang/crates.io-index)",
  "regex",
  "semver",
 ]
@@ -2270,6 +2281,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "ref-cast"
+version = "1.0.25"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f354300ae66f76f1c85c5f84693f0ce81d747e2c3f21a45fef496d89c960bf7d"
+dependencies = [
+ "ref-cast-impl",
+]
+
+[[package]]
+name = "ref-cast-impl"
+version = "1.0.25"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b7186006dcb21920990093f30e3dea63b7d6e977bf1256be20c3563a5db070da"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.106",
+]
+
+[[package]]
 name = "reflink-copy"
 version = "0.1.28"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2551,6 +2582,31 @@ dependencies = [
 ]
 
 [[package]]
+name = "schemars"
+version = "1.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "82d20c4491bc164fa2f6c5d44565947a52ad80b9505d8e36f8d54c27c739fcd0"
+dependencies = [
+ "dyn-clone",
+ "ref-cast",
+ "schemars_derive",
+ "serde",
+ "serde_json",
+]
+
+[[package]]
+name = "schemars_derive"
+version = "1.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "33d020396d1d138dc19f1165df7545479dcd58d93810dc5d646a16e55abefa80"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "serde_derive_internals",
+ "syn 2.0.106",
+]
+
+[[package]]
 name = "scopeguard"
 version = "1.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


ognis1205 (changes requested):
First, please run the formatter with:
```sh
cargo +nightly fmt --all -- --verbose
```
and then run clippy with:
```sh
cargo clippy --tests --verbose -- -D warnings
```
to ensure the code passes both checks.

While using `schemars` to generate JSON Schemas in a code-first manner from the `git-cliff-core` and `git-conventional` models isn't necessarily a bad idea, both crates already expose their internal models publicly.  Given that, adding schema-generation logic directly into the CLI might slightly drift away from the core purpose of `git-cliff,` which is primarily a changelog generator.

If you still prefer to support schema generation, you might consider isolating that functionality.  For instance, by using [remote deriving](https://graham.cool/schemars/examples/5-remote_derive), you could maintain a small separate project that depends on `git-cliff-core` and `git-conventional`, and generate schemas there instead.

This would keep the CLI focused on its main responsibility while still enabling schema generation for external tooling.

greenhat616:
I have done the `fmt` and `clippy` check.

Regarding the Schema → SDK issue: I agree this workflow is not ideal in most cases. While Cliff does not define a stable Schema version, every change to `Context` can break existing interfaces—and when using Python or TS there is no type checking at runtime. Therefore, if we use a Schema, we must pin the current Cliff version.

Regarding Schema generation, after careful consideration, I approved for defining a separate remote definition for `git-conventional`. However, the `Release` type inside `git-cliff-core` is quite complex; defining it in a third-party library would require significant effort to adapt after each release.

This PR is just a draft idea for now. Maybe a better a approach: We could gate this with a feature flag and, via some process (scripts or an xtask, etc.), generate a Schema when Cliff publishes a release and upload it to the Releases artifacts. Would that be more appropriate?

ognis1205:
cc: @orhun 

Thanks for clarifying. I understand this PR is still a draft.
However, it seems that some tests are failing. Have you run the tests locally to confirm?

Also, I’m a bit hesitant about this feature itself, as well as adding a related option to the CLI. We should consider whether this functionality is truly needed for most users. Even if it is, it might be cleaner, from a separation-of-concerns perspective, to add a separate binary under `git-cliff-core` rather than integrating it into the main CLI.

> I agree this workflow is not ideal in most cases.

As you also noted, the current implementation direction seems problematic to me as well.

> defining it in a third-party library would require significant effort to adapt after each release.
> generate a Schema when Cliff publishes a release and upload it to the Releases artifacts.

In other words, what's being proposed here effectively shifts the maintenance burden to the maintainers, before we've even established a clear need for the feature. I think we should have a more thorough discussion about the necessity and maintenance cost of this functionality before proceeding further.

Personally, I would still recommend creating a separate repository dedicated to schema generation, which could depend on both `git-cliff-core` and `git-conventional`.

orhun:
Hey, sorry for the delay on this. I'll be having a look soon hopefully

orhun:
Although I like the changes in this PR, I agree that the output of this CLI command would be too specific to the current release of `git-cliff`. In other words, it's too dynamic and it might limit our capability to update the context without worrying about user setups that depend on this. We could maybe use a `version` key, but I believe it will be incremented very quickly if we add/remove/update fields.

I'm not sure how a separate repo would look like for this. Are there any example projects that do that we could get inspired from?

Thoughts?

ognis1205:
I don't know of any open-source projects that apply an implementation or operational approach like the one presented in this PR, but I believe there are internal tools that use JSON Schema to validate data type consistency.

I could try generating the schema using `build.rs` in a designated repository with the method I mentioned earlier. In particular, since the data models in `git-cliff-core` are expected to vary depending on feature flags (even though all backend feature flags are enabled by default in `git-cliff`), I think it would make more sense to manage the JSON Schemas in a separate dedicated repository, rather than deriving `JsonSchema` directly within `git-cliff-core`.

orhun:
Yeah, I think that's a fair approach. Thanks for looking into this!

ognis1205:
@greenhat616 

By the way, I've re-read the related issue:

https://github.com/orhun/git-cliff/issues/1294

and I’m wondering — is `git-cliff` actually being used in the development cycle described there?

From my understanding, it doesn't seem like the issue's context involves using `git-cliff` directly. Could you clarify if I'm missing something?

greenhat616:
@ognis1205
> By the way, I've re-read the related issue:
> 
> https://github.com/orhun/git-cliff/issues/1294
> 
> and I’m wondering — is git-cliff actually being used in the development cycle described there?
> 
> From my understanding, it doesn't seem like the issue's context involves using git-cliff directly. Could you clarify if I'm missing something?

The limited for `git-cliff` is not able to run scripts (lua, js etc), and query third-part API to query more detailed information to generate changelog. Compared with introducing a scripting engine for a custom preprocessing stage, adding type info to the existing `Context`—so third-party tools can handle it more easily—would require fewer changes, wouldn’t it?

For example, if you’re using third-party planning/project-management tools like Linear or Feishu and need to map your journal content (preferably only PRs) to task id and titles in those tools, then using `Context` with scripting for custom processing is a great fit.

Sry, the generator is internal tool, I cannot share it public.

> I don't know of any open-source projects that apply an implementation or operational approach like the one presented in this PR, but I believe there are internal tools that use JSON Schema to validate data type consistency.
> 
> I could try generating the schema using `build.rs` in a designated repository with the method I mentioned earlier. In particular, since the data models in `git-cliff-core` are expected to vary depending on feature flags (even though all backend feature flags are enabled by default in `git-cliff`), I think it would make more sense to manage the JSON Schemas in a separate dedicated repository, rather than deriving `JsonSchema` directly within `git-cliff-core`.

Yes, I agree with this approach as well. This PR is mainly a prototype for the workflow described above. Publishing it in a separate repository, or bundling it with a Release artifact, would both be good options. External users could use this file to directly generate the `Context` for the current version. The schema metadata could be provided as a `cliff-core` feature rather than as a command flag in the cli.



> Although I like the changes in this PR, I agree that the output of this CLI command would be too specific to the current release of `git-cliff`. In other words, it's too dynamic and it might limit our capability to update the context without worrying about user setups that depend on this. We could maybe use a `version` key, but I believe it will be incremented very quickly if we add/remove/update fields.
> 
> I'm not sure how a separate repo would look like for this. Are there any example projects that do that we could get inspired from?
> 
> Thoughts?

@orhun 

Yes. Cliff is an actively evolving project, and the Context may change quite frequently. There’s no plan to provide stable type guarantees, and introducing a VERSION field to mark versions would be cumbersome. So treating this as something (a) exposed only behind a feature flag, (b) published alongside each release, or (c) maintained as a per-version schema in the repository are all acceptable trade-offs. 

Asking users to define their own interfaces to adapt to the Context is too cumbersome and hard to validate—if the type definitions change substantially, it’s difficult for type checkers like `tsc` or `pyright` to automatically surface those changes. That would make upgrading cliff versions more painful.

ognis1205:
Thanks for the clarification @greenhat616 .

I still haven't fully grasped in your use case **which specific field(s)** of `Context` you intend to use and how you’d like to add **Linear-related information (such as task IDs or titles)** to **which part of the `Context`**.

> query third-part API to query more detailed information to generate changelog.

I'm not familiar with Linear myself, but according to their docs, if you include the Linear issue ID in the PR title, Linear automatically links the PR to the issue. If that's the case, wouldn't it be possible to use this feature:

https://github.com/orhun/git-cliff/pull/1287

in `git-cliff` to convert issue IDs in PR titles into links to Linear issues during changelog generation?

Also, regarding schema validation: typically, schema or type validation is used at untrusted data boundaries. I'm not sure if `git-cliff` really falls into that category. To me, the motivation for providing a JSON Schema seems more like "avoiding the need to handcraft data models," rather than addressing data integrity or validation concerns.

If only certain fields of the context are problematic, it might be simpler to just treat the context as plain JSON and access or update the relevant fields via their paths — without deserializing the entire structure into a strict data model. After all, `git-cliff` doesn't seem to operate in an untrusted environment.

That said, it might be a good idea to first make sure this PR passes the tests and that the JSON Schema is properly generated. At the moment, it's hard to verify anything without that.

Issue #1284, opened 2025-10-04
strict require_conventional


MeitarR:
### Is there an existing issue or pull request for this?

- [x] I have searched the existing issues and pull requests

### Feature description

We want to be able to check if our commits are valid conventional and also verify that their type is recognized (one of a predetermined list)

### Desired solution

I'm not sure if this is the best solution, but maybe add a list of `require_conventional_commit_types` to the config that only when it has values and `require_conventional` is true, then it will also verify that the commit types are one from the list.

### Alternatives considered

Check it with another tool instead

orhun:
Hello, thanks for reporting/idea!

How about something like this instead:

```toml
[git]
commit_parsers = [
  { message = "test", required = true }
]
```

So the changelog generation will fail if nothing is matched. This will also work with unconventional commits if configured.

WDYT?

MeitarR:
If I get it right, this means that if you don't use all the parsers, then it will fail which is bad for me.

More context about the reason I want this feature:  
We have for most of our repos a CI job that runs for every MR and comment (and updates the message every time) with how this MR will affect the change log (using git-cliff, of course). So I want to also verify that all the commits added in the MR are valid (and if not, then fail the CI)


To continue with your idea, maybe we can add a new property `fail_on_unmatched_commit`  that will have the same condition `filter_commits`, just instead of filtering out, it will fail

orhun:
> If I get it right, this means that if you don't use all the parsers, then it will fail which is bad for me.

Not really, it will only fail if `required` is `true` for an individual parser and if it's not being used.

But yeah, I see what you mean. It will also fail if the parser isn't matched because you don't have a specific commit type in your release.

Hmm, actually, isn't that what you want? 🤔 

> To continue with your idea, maybe we can add a new property fail_on_unmatched_commit that will have the same condition filter_commits, just instead of filtering out, it will fail

Do you mean this as a part of the `commit_parsers`?

MeitarR:
> > If I get it right, this means that if you don't use all the parsers, then it will fail which is bad for me.
> 
> Not really, it will only fail if `required` is `true` for an individual parser and if it's not being used.
> 
> But yeah, I see what you mean. It will also fail if the parser isn't matched because you don't have a specific commit type in your release.
> 
> Hmm, actually, isn't that what you want? 🤔 

[...]

PR #1298, opened 2025-10-31
feat(changelog): support failing on unmatched commits

MeitarR:
## Description

Added a new config `fail_on_unmatched_commit` for `git`.

## Motivation and Context

solves #1284 

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/website/docs/configuration/git.md
+++ b/website/docs/configuration/git.md
@@ -216,12 +216,14 @@ If set to `true`, commits that are not matched by [`commit_parsers`](#commit_par
 
 ### fail_on_unmatched_commit
 
-If set to `true`, git-cliff will fail (return a non-zero exit code) when any commit included in the changelog is not matched by any of the configured `commit_parsers`.
+If set to `true`, **git-cliff** will fail (return a non-zero exit code) when any commit included in the changelog is not matched by any of the configured `commit_parsers`.
 
 This option is useful to enforce that all commits are classified and prevents silently omitting unknown commit types from release notes.
 
 :::note
+
 Commits that are explicitly skipped by a parser (for example `{ message = "^revert", skip = true }`) are not considered unmatched.
+
 :::
 
 ### tag_pattern


orhun:
Nice, we should also add a test fixture for this

git-cliff/src/args.rs
```diff
@@ -229,6 +229,9 @@ pub struct Opt {
     /// Include only the tags that belong to the current branch.
     #[arg(long, help_heading = Some("FLAGS"))]
     pub use_branch_tags: bool,
+    /// Fail on a commit that is not matched by any commit parser.
+    #[arg(short, long, help_heading = Some("FLAGS"))]
+    pub fail_on_unmatched_commit: bool,
```
orhun: Can we remove the CLI argument for now? I'm not sure if this option will be used common enough that it will need to be activated by the CLI.

website/docs/usage/args.md
```diff
@@ -11,18 +11,19 @@ git-cliff [FLAGS] [OPTIONS] [--] [RANGE]
...
+      --bumped-version            Prints bumped version for unreleased changes
+  -l, --latest                    Processes the commits starting from the latest tag
+      --current                   Processes the commits that belong to the current tag
+  -u, --unreleased                Processes the commits that do not belong to a tag
+      --topo-order                Sorts the tags topologically
+      --use-branch-tags           Include only the tags that belong to the current branch
+  -f, --fail-on-unmatched-commit  Fail on a commit that is not matched by any commit parser
+      --no-exec                   Disables the external command execution
+  -x, --context                   Prints changelog context as JSON
+      --use-native-tls            Load TLS certificates from the native certificate store
```
orhun: We can revert this change if we remove the CLI argument. (also the indentation is looking a bit off)

git-cliff-core/src/error.rs
```diff
@@ -112,6 +112,10 @@ pub enum Error {
     /// See `require_conventional` option for more information.
     #[error("Requiring all commits be conventional but found {0} unconventional commits.")]
     UnconventionalCommitsError(i32),
+    /// Error raised when commits are not matched by any commit parser and the
+    /// `fail_on_unmatched_commit` option is enabled.
```
orhun: Maybe link to the API docs like:

```suggestion
    /// [`GitConfig::fail_on_unmatched_commit`] option is enabled.
```

git-cliff-core/src/changelog.rs
```diff
@@ -157,6 +186,10 @@ impl<'a> Changelog<'a> {
             })
             .collect::<Vec<Commit>>();
 
+        if git_config.fail_on_unmatched_commit {
+            Self::check_unmatched_commits(commits)?;
+        }
```
orhun: Should we do this before or after the unconventional commits check?

git-cliff/src/args.rs
MeitarR: Actually, this is significant for my use case. I plan to use this check only on a CI check and not over the entire changelog (I don't want the changelog generation to fail because of old commits). I think it will be most of its use for other users.

orhun: Then this could be still overridden by a environment variable I believe.

I prefer to wait a bit before landing config options as CLI arguments.

git-cliff-core/src/error.rs
MeitarR: applied

git-cliff-core/src/changelog.rs
MeitarR: It's more logical to check for unmatched after checking for unconventional commits. I changed to this order

git-cliff/src/args.rs
MeitarR: ok, I can get that. So I will add an ENV var instead

MeitarR: hid the arg and added env

git-cliff/src/args.rs
```diff
@@ -339,6 +339,9 @@ pub struct Opt {
     /// Load TLS certificates from the native certificate store.
     #[arg(long, help_heading = Some("FLAGS"), hide = !cfg!(feature = "remote"))]
     pub use_native_tls: bool,
+    /// Fail on a commit that is not matched by any commit parser.
+    #[arg(short, long, env = "GIT_CLIFF_FAIL_ON_UNMATCHED_COMMIT", hide = true)]
+    pub fail_on_unmatched_commit: bool,
```
orhun: We can remove this argument alltogether. The config values already support environment variables: https://git-cliff.org/docs/configuration/#environment-configuration-overrides

So in this case, you can already override it via `GIT_CLIFF__GIT__FAIL_ON_UNMATCHED_COMMIT`

MeitarR: Ha cool! I forgot that.

Removed that commit

orhun (approved):
Looks good, only need to update the website about this new option :)

MeitarR:
> Looks good, only need to update the website about this new option :)

Added in 7e41a47

PR #1306, opened 2025-11-04
refactor(core): improve the generation API

orhun:
### New Methods

- `init_config`: initializes and writes a configuration file.
- `write_changelog`: writes the generated changelog to a file or stdout.

### Breaking Changes

- `Changelog::new` now takes ownership of `Config` (`Config` instead of `&Config`).
- `Changelog::from_context` and internal `build` methods also updated to take `Config` by value.
- `run` now returns `Result<Changelog<'static>>` instead of `Result<()>`.

### Refactors

- Configuration initialization logic moved out of `run_with_changelog_modifier()` into the new `init_config()` function.
- The main entry point (`main.rs`) now explicitly calls `check_new_version()`, `init_config()`, `run()`, and `write_changelog()` instead of relying on `run()` for all logic.
- Example renamed from `examples/run.rs` to `examples/basic.rs` and updated to use the new API (`write_changelog`).

### Minor Changes

- `check_new_version()` made public.
- `Opt` now derives `Clone` to allow reuse across function calls.

## Motivation and Context

closes #1046

## How Has This Been Tested?

Locally.

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/git-cliff/src/lib.rs
+++ b/git-cliff/src/lib.rs
@@ -486,7 +486,7 @@ fn process_repository<'a>(
 ///     Ok(())
 /// }
 /// ```
-pub fn run(args: Opt) -> Result<Changelog<'static>> {
+pub fn run<'a>(args: Opt) -> Result<Changelog<'a>> {
     run_with_changelog_modifier(args, |_| Ok(()))
 }
 
@@ -513,10 +513,10 @@ pub fn run(args: Opt) -> Result<Changelog<'static>> {
 ///     Ok(())
 /// }
 /// ```
-pub fn run_with_changelog_modifier(
+pub fn run_with_changelog_modifier<'a>(
     mut args: Opt,
     changelog_modifier: impl FnOnce(&mut Changelog) -> Result<()>,
-) -> Result<Changelog<'static>> {
+) -> Result<Changelog<'a>> {
     // Retrieve the built-in configuration.
     let builtin_config = BuiltinConfig::parse(args.config.to_string_lossy().to_string());
 


orhun:
cc @marcoieni (for release-plz downstream)

git-cliff/src/lib.rs
```diff
@@ -769,20 +767,23 @@ pub fn run_with_changelog_modifier(
                 &args,
             )?);
         }
-        Changelog::new(releases, &config, commit_range.as_deref())?
+        Changelog::new(releases, config.clone(), commit_range.as_deref())?
```
ognis1205:
```suggestion
        Changelog::new(releases, config, commit_range.as_deref())?
```

git-cliff/src/lib.rs
```diff
@@ -156,6 +156,30 @@ fn process_submodules(
...
 }
 
+/// Initializes the configuration file.
+pub fn init_config(name: Option<String>, config_path: PathBuf) -> Result<()> {
+    let contents = match name {
+        Some(ref name) => BuiltinConfig::get_config(name.to_string())?,
+        None => EmbeddedConfig::get_config()?,
+    };
+
+    let config_path = if config_path == PathBuf::from(DEFAULT_CONFIG) {
```
ognis1205:
```suggestion
    let config_path = if config_path == Path::new(DEFAULT_CONFIG) {
```

git-cliff/examples/basic.rs
```diff
@@ -4,6 +4,7 @@ use git_cliff_core::error::Result;
 
 fn main() -> Result<()> {
     let args = Opt::parse();
-    git_cliff::run(args)?;
+    let changelog = git_cliff::run(args.clone())?;
+    git_cliff::write_changelog(&args, changelog, std::io::stdout())?;
     Ok(())
```
ognis1205, line 9:
```suggestion
    git_cliff::run(args.clone())
        .and_then(|changelog| git_cliff::write_changelog(&args, changelog, std::io::stdout()))
```

git-cliff/src/lib.rs
```diff
@@ -156,6 +156,30 @@ fn process_submodules(
     Ok(())
 }
 
+/// Initializes the configuration file.
+pub fn init_config(name: Option<String>, config_path: PathBuf) -> Result<()> {
```
ognis1205:
```suggestion
pub fn init_config(name: Option<&str>, config_path: &Path) -> Result<()> {
```

git-cliff/src/lib.rs
```diff
@@ -156,6 +156,30 @@ fn process_submodules(
...
+    let contents = match name {
+        Some(ref name) => BuiltinConfig::get_config(name.to_string())?,
```
ognis1205:
```suggestion
        Some(name) => BuiltinConfig::get_config(name.to_string())?,
```

git-cliff/src/lib.rs
```diff
@@ -156,6 +156,30 @@ fn process_submodules(
...
+        None => EmbeddedConfig::get_config()?,
+    };
+
+    let config_path = if config_path == PathBuf::from(DEFAULT_CONFIG) {
+        PathBuf::from(DEFAULT_CONFIG)
+    } else {
+        config_path.clone()
```
ognis1205:
```suggestion
        config_path.to_path_buf()
```

git-cliff/src/main.rs
```diff
@@ -32,9 +34,33 @@ fn main() -> Result<()> {
...
-    // Run git-cliff
-    let exit_code = match git_cliff::run(args) {
-        Ok(()) => 0,
+    // Check if there is a new version available.
+    #[cfg(feature = "update-informer")]
+    git_cliff::check_new_version();
+
+    // Create the configuration file if init flag is given.
+    if let Some(path) = &args.init {
+        init_config(path.clone(), args.config.clone())?;
```
ognis1205:
```suggestion
        init_config(path.as_deref(), &args.config)?;
```

git-cliff/examples/basic.rs
orhun: I'd like to keep this more explicit since it's an example.

orhun:
Thanks for the great review! I applied your suggestions

git-cliff/examples/basic.rs
ognis1205: I think this is more about personal preference, so it’s totally fine as it is.

git-cliff/src/lib.rs
```diff
@@ -462,7 +486,7 @@ fn process_repository<'a>(
 ///     Ok(())
 /// }
 /// ```
-pub fn run(args: Opt) -> Result<()> {
+pub fn run(args: Opt) -> Result<Changelog<'static>> {
```
ognis1205:
```suggestion
pub fn run<'a>(args: Opt) -> Result<Changelog<'a>> {
```

git-cliff/src/lib.rs
```diff
@@ -492,33 +516,7 @@ pub fn run(args: Opt) -> Result<()> {
...
-        log::info!(
-            "Saving the configuration file{} to {:?}",
-            init_config.map(|v| format!(" ({v})")).unwrap_or_default(),
-            config_path
-        );
-        fs::write(config_path, contents)?;
-        return Ok(());
-    }
-
+) -> Result<Changelog<'static>> {
```
ognis1205:
```suggestion
pub fn run_with_changelog_modifier<'a>(
    mut args: Opt,
    changelog_modifier: impl FnOnce(&mut Changelog) -> Result<()>,
) -> Result<Changelog<'a>>
```

ognis1205 (changes requested):
Thanks for accepting the previous suggestion!

I left a suggestion regarding a point I missed in the last review. As an additional note, I changed the lifetime from `'static` to `'a` for `Changelog<'a>`. This allows the caller to decide the appropriate lifetime. Although we kept the `'a` lifetime for flexibility, in practice the returned `Changelog` is always `'static` because it owns all its data internally. This is mostly a stylistic choice, so feel free to ignore if you prefer.

I'm also quoting here the discussion from Discord about `Cow`:

> Considering `git-cliff` as a CLI tool, currently the only place calling `Changelog::new` or `Changelog::from_context` is `run_with_changelog_modifier`, so I don't think using `Cow<'a, Config>` is necessary.
>
> However, if we view `git-cliff-core` as an SDK that could be used by external applications, implementing `impl<'a> From<Config> for Cow<'a, Config>` and changing the arguments of `Changelog::new`/`Changelog::build`/`Changelog::from_context` to `config: impl Into<Cow<'a, Config>>` seems like a reasonable option.

ognis1205:
In another thread, I discussed with @orhun the idea of changing the struct to something like:

```rust
pub struct Changelog<'a> {
    /// Releases that the changelog will contain.
    pub releases: Vec<Release<'a>>,
    /// Configuration used for generating the changelog.
    pub config: Cow<'a, Config>,
    header_template: Option<Template>,
    body_template: Template,
    footer_template: Option<Template>,
    additional_context: HashMap<String, serde_json::Value>,
}
```

and updating the API in `Changelog::new` / `Changelog::from_context` to:

```rust
pub fn new(releases: Vec<Release<'a>>, config: impl Into<Cow<'a, Config>>, range: Option<&str>) -> Result<Self>
```

I think this approach, using `Cow<'a, Config>`, makes sense when considering `git-cliff-core` as an SDK, since it maintains backward compatibility. On the other hand, since I'm not personally using `git-cliff-core` as an SDK, I'd like to hear the opinion of @marcoieni , who is implementing `release-plz`.

That said, I don't want to burden everyone with extra review work, so I think it's fine to proceed with the current approach for now. If any compatibility issues arise later, we can revisit and implement this change separately, it shouldn't require too much additional effort.

ognis1205:
As @orhun mentioned earlier, in the long term, we'll likely need a broader overhaul of `git-cliff`'s `lib.rs`.

In particular, it might be worth reconsidering whether the `Changelog` struct should hold a reference to `Config`, own it directly, or not depend on it at all.

For this PR, I think we can proceed as is, but it could be good to open a separate long-term issue to track that discussion.

marcoieni:
I have no time to review this unfortunately. Feel free to do all the updates you want. I'll just fix them in release-plz. If you bump the major version of git-cliff-core when publishing the breaking changes it would be great!

orhun:
Thanks for the review again Shingo!

ognis1205:
> Thanks for the review again Shingo!

No problem :) If there’s anything I can help with, just ping me anytime.

Issue #1268, opened 2025-09-24
Migrate logging from `env_logger` to `tracing` + `tracing-subscriber`


ognis1205:
### Is there an existing issue or pull request for this?

- [x] I have searched the existing issues and pull requests

### Feature description

Currently, `env_logger` is being used in a special way (to trigger the progress bar when fetching data for remotes). This usage prevents upgrading to newer versions of `env_logger` due to API changes, and the implementation is somewhat hacky.

### Desired solution

Migrate from `env_logger` to `tracing` and `tracing-subscriber`. This would provide:

 - Spans: Representing nested events with start/end times, capturing the flow across system components.
 - Structured data: Richer data model for events and spans.
 - Causality and temporality: Better understanding of temporal relationships and causal flow.
 - Function-level instrumentation: Using the `#[instrument]` attribute to automatically capture arguments and return values in logs.

### Alternatives considered

 - Continue using `env_logger` with the current hack to support the progress bar (but this blocks upgrading env_logger).
 - Replace only the progress bar trigger with another mechanism, keeping `env_logger` for basic logging.

### Additional context

 - https://github.com/orhun/git-cliff/discussions/1191
 - https://github.com/orhun/git-cliff/blob/main/git-cliff/src/logger.rs

orhun:
Thanks for creating the issue!

Yup, this would be great.

orhun:
FWIW there is a crate called [`tracing-indicatif`](https://github.com/emersonford/tracing-indicatif) which might be a perfect fit for our case.

ognis1205:
> FWIW there is a crate called [`tracing-indicatif`](https://github.com/emersonford/tracing-indicatif) which might be a perfect fit for our case.

I haven’t looked into it deeply yet, but yeah, this does look like it could be a good fit for implementing the progress bar. I think it makes sense to move forward with this approach! Thanks for sharing!

PR #1308, opened 2025-11-21
feat(tracing): migrate logging to tracing

ognis1205:
## Description

This PR migrates the logging implementation from `env_logger` to `tracing`.

The migration preserves the existing log format while extending it to include span information, enabling more structured and contextual logging without changing the default user experience.

## Motivation and Context

This change is motivated by the following discussion and issues:

- https://github.com/orhun/git-cliff/discussions/1191
- https://github.com/orhun/git-cliff/issues/1268

The goal is to modernize `git-cliff`'s logging infrastructure by adopting `tracing`, while maintaining backward compatibility and avoiding unnecessary complexity in the core crate.

## What Has Changed

- The logger has been updated to emit spans in a format consistent with the existing log output.
- Span instrumentation is applied only to methods that affect the changelog generator's semantics (e.g. changelog, release, and commit processing).
- `git-cliff-core` now depends on `tracing` behind a feature flag.
   - When the feature is disabled, the behavior is unchanged and only the existing `log` output is produced.
   - When the feature is enabled, spans are created and included in the output.
- Remote metadata progress reporting has been fully migrated to a `tracing`-based implementation using `tracing-indicatif`.
- The progress bar implementation is intentionally kept simple.
   - While richer progress reporting is possible with `tracing`, implementing such behavior cleanly would require API changes outside of `git-cliff-core`.
   - Given the current API design, this PR avoids introducing additional complexity and leaves room for future design discussions.

## How Has This Been Tested?

- Existing unit tests were run.
- Logging output was verified manually.

N/A

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/git-cliff-core/src/lib.rs
+++ b/git-cliff-core/src/lib.rs
@@ -50,3 +50,19 @@ pub const DEFAULT_CONFIG: &str = "cliff.toml";
 pub const DEFAULT_OUTPUT: &str = "CHANGELOG.md";
 /// Default ignore file.
 pub const IGNORE_FILE: &str = ".cliffignore";
+
+/// Sets a human-readable message on the current progress bar span.
+/// This macro only has effect if the `tracing-indicatif` feature is enabled.
+#[doc(hidden)]
+#[macro_export]
+macro_rules! pb_msg {
+    ($($arg:tt)*) => {{
+        #[cfg(feature = "tracing-indicatif")]
+        {
+            use tracing::Span;
+            use tracing_indicatif::span_ext::IndicatifSpanExt;
+            let msg = format!($($arg)*);
+            Span::current().pb_set_message(&msg);
+        }
+    }};
+}
--- a/git-cliff-core/src/lib.rs
+++ b/git-cliff-core/src/lib.rs
@@ -55,7 +55,7 @@ pub const IGNORE_FILE: &str = ".cliffignore";
 /// This macro only has effect if the `tracing-indicatif` feature is enabled.
 #[doc(hidden)]
 #[macro_export]
-macro_rules! pb_msg {
+macro_rules! set_progress_message {
     ($($arg:tt)*) => {{
         #[cfg(feature = "tracing-indicatif")]
         {
--- a/git-cliff/src/logger.rs
+++ b/git-cliff/src/logger.rs
@@ -136,7 +136,7 @@ fn indicatif_progress_style() -> ProgressStyle {
     .with_key("color_start", color_start_key)
     .with_key("color_end", color_end_key)
     .with_key("dim_start", dim_start_key)
-    .with_key("eim_end", dim_end_key)
+    .with_key("dim_end", dim_end_key)
     .tick_strings(SPINNER_TICKS)
 }
 
--- a/git-cliff/src/logger.rs
+++ b/git-cliff/src/logger.rs
@@ -6,8 +6,8 @@ use indicatif::{ProgressState, ProgressStyle};
 use owo_colors::{OwoColorize, Style, Styled};
 use tracing::{Event, Level, Subscriber};
 use tracing_indicatif::IndicatifLayer;
+use tracing_subscriber::fmt::FmtContext;
 use tracing_subscriber::fmt::format::{self, FormatEvent, FormatFields};
-use tracing_subscriber::fmt::{FmtContext, FormattedFields};
 use tracing_subscriber::layer::SubscriberExt;
 use tracing_subscriber::registry::LookupSpan;
 use tracing_subscriber::util::SubscriberInitExt;
@@ -173,13 +173,6 @@ where
         if let Some(scope) = ctx.event_scope() {
             for span in scope.from_root() {
                 write!(writer, "{}", span.name().bright_black().bold())?;
-                let ext = span.extensions();
-                let fields = &ext
-                    .get::<FormattedFields<N>>()
-                    .expect("will never be `None`");
-                if !fields.is_empty() {
-                    write!(writer, "{{{}}}", fields)?;
-                }
                 write!(writer, "{}", ": ".bright_black().bold())?;
             }
         }
--- a/git-cliff/src/logger.rs
+++ b/git-cliff/src/logger.rs
@@ -157,7 +157,7 @@ where
         let metadata = event.metadata();
         let level = style_level(metadata.level());
         let target = metadata.target();
-        let max_width = max_target_width(&target);
+        let max_width = max_target_width(target);
         write!(
             &mut writer,
             "{} {} > ",
--- a/git-cliff/src/logger.rs
+++ b/git-cliff/src/logger.rs
@@ -95,7 +95,7 @@ fn color_start_key(state: &ProgressState, writer: &mut dyn fmt::Write) {
 
 /// Performs linear interpolation between two color components.
 fn lerp(a: u8, b: u8, t: f32) -> u8 {
-    (a as f32 + (b as f32 - a as f32) * t) as u8
+    ((a as f32 + (b as f32 - a as f32) * t).clamp(0.0, 255.0)) as u8
 }
 
 /// Resets ANSI styling to the terminal default.
--- a/git-cliff/src/logger.rs
+++ b/git-cliff/src/logger.rs
@@ -43,8 +43,8 @@ fn max_target_width(target: &str) -> usize {
 }
 
 /// Adds styles/colors to the given level and returns it.
-fn style_level(level: &Level) -> Styled<&'static str> {
-    match *level {
+fn style_level(level: Level) -> Styled<&'static str> {
+    match level {
         Level::ERROR => Style::new().red().bold().style("ERROR"),
         Level::WARN => Style::new().yellow().bold().style("WARN"),
         Level::INFO => Style::new().green().bold().style("INFO"),
@@ -72,7 +72,7 @@ fn shorten_target(target: &str, width: usize) -> String {
 fn elapsed_subsec_key(state: &ProgressState, writer: &mut dyn fmt::Write) {
     let seconds = state.elapsed().as_secs();
     let sub_seconds = (state.elapsed().as_millis() % 1000) / 100;
-    let _ = write!(writer, "{}.{}s", seconds, sub_seconds);
+    let _ = write!(writer, "{seconds}.{sub_seconds}s");
 }
 
 /// Emits an ANSI color escape sequence for the spinner, based on elapsed time.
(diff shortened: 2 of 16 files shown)


orhun:
10 seconds ago I was happy that my GitHub notifications were clear and then Shingo comes in with a crazy PR!!!!!

+0 additions -0 deletions though. Love it. LGTM

orhun:
planning to do any progress on this soon? (added to 2.12.0 milestone)

ognis1205:
> planning to do any progress on this soon? (added to 2.12.0 milestone)

Yeah, I'm planning to open a PR later this week.
Does that timeline work for the 2.12.0 milestone?

orhun:
yup, 2.12.0 will happen in 2026 :)
so we have a whole year!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

ognis1205:
Sorry for the delay. I'll start working on this PR today.

orhun:
No worries at all!

ognis1205:
This is ready for review now @orhun .

I consider this a first draft, and there are likely areas open for discussion. When you have time, I'd really appreciate your thoughts and feedback.

ognis1205 (commented):
TODO:

- [x] Replace all ASCII color codes with `owo_colors`.
- [x] Consider how to handle Windows and non-TTY environments (this may also have been an issue in the previous implementation).

git-cliff/src/logger.rs
```diff
@@ -1,20 +1,24 @@
...
+use tracing_subscriber::layer::SubscriberExt;
+use tracing_subscriber::registry::LookupSpan;
+use tracing_subscriber::util::SubscriberInitExt;
+use tracing_subscriber::{EnvFilter, Registry};
 
 /// Global variable for storing the maximum width of the modules.
 static MAX_MODULE_WIDTH: AtomicUsize = AtomicUsize::new(0);
 
+/// Unicode braille spinner frames used by indicatif.
+const SPINNER_TICKS: &[&str] = &["⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "⠧", "⠇", "⠏"];
```
orhun, line 20: Would be great if we could keep the old arrow-styled progress indicator (for the first line in the logs at least)

I really like the detailed logging btw!

Let me know when this is ready for review

ognis1205: > Would be great if we could keep the old arrow-styled progress indicator (for the first line in the logs at least)

That should be doable.

We can bring back the old arrow-style progress indicator, but when implemented on top of `tracing` / `span`-based instrumentation via `tracing-indicatif`, it would look slightly different from the previous approach:

Screenshot 2026-01-18 at 16 45 03
Screenshot 2026-01-18 at 16 45 27
Screenshot 2026-01-18 at 16 46 34
Screenshot 2026-01-18 at 16 46 57

If that's acceptable, I'm happy to revert the change accordingly. In addition to the progress indicator itself, we can of course also emit explicit logs at the beginning and end of the processing step.

In an ideal world, I'd like to include a progress indicator with percentages as well. However, with the current design, where `span`s and logging live inside the core SDK, that starts to get a bit messy from a design perspective. It's technically possible, but I intentionally avoided it for now to keep the core clean.

> Let me know when this is ready for review

Regarding ANSI/ASCII color codes on Windows or non-TTY environments: since we were already using colored output before, I think it's reasonable to review this as-is for now.

If you're open to it, I'd really appreciate review feedback on the following points in the context of adding this to `git-cliff-core` as an SDK:

- Which functions should be annotated with `#[instrument]` (all vs. a subset)
- Whether `span` names should be explicitly set
- Which fields are appropriate to capture on spans

From a "keep git-cliff-core as a clean SDK" perspective, another option could be to explore whether progress reporting can be handled in a less hacky way on top of `env_logger`, if that's feasible. Just putting it out there as an alternative.

orhun (commented):
I like how this turned out!

Some points of discussion:

- The spinner gets red quickly I think. Wouldn't be nice to adjust the timing there?
- The main spinner messages seem developer oriented now. Previously, it was something like "fetching data from GitHub..." but now it says things like `add_remote_data`, `get_github_metadata` and so on. I know git-cliff is mostly directed at developers, but I still think that we could avoid programmatic-looking info-level messages. Wdyt?

git-cliff/src/logger.rs
```diff
@@ -38,120 +42,162 @@ fn max_target_width(target: &str) -> usize {
...
+}
+
+/// Emits an ANSI color escape sequence for the spinner, based on elapsed time.
+///
+/// The color gradually transitions:
+/// - green  -> yellow (0–4s)
+/// - yellow -> red    (4–8s)
+fn color_start_key(state: &ProgressState, writer: &mut dyn fmt::Write) {
+    let elapsed = state.elapsed().as_secs_f32();
+    let t = (elapsed / 8.0).min(1.0); // 8秒で変化
```
orhun: I should improve my Japanese!

ognis1205: I will remove the comment here!

ognis1205:
Thanks for the feedback!

> * The spinner gets red quickly I think. Wouldn't be nice to adjust the timing there?

No objections here. I agree it turns red a bit too quickly. I'll slow down the timing so the transition happens later.

> * The main spinner messages seem developer oriented now. Previously, it was something like "fetching data from GitHub..." but now it says things like `add_remote_data`, `get_github_metadata` and so on. I know git-cliff is mostly directed at developers, but I still think that we could avoid programmatic-looking info-level messages. Wdyt?

There is actually a reason for the current behavior. Since I implemented using `tracing-indicatif`, most of the macros ended up being defined in the `git-cliff-core` side. Because that crate also has an SDK role, I intentionally avoided defining end-user-facing messages there and instead exposed the function names directly.

That said, I don't have any objection to making the messages more human-readable. In the future, when we have a chance to refactor, I'd like to explore the idea of making `git-cliff-core` more "log-less" and keeping user-facing messaging closer to the CLI layer.

For now, I agree with your point and will update the messages to be more readable.

ognis1205:
> > * The spinner gets red quickly I think. Wouldn't be nice to adjust the timing there?
> 
> No objections here. I agree it turns red a bit too quickly. I'll slow down the timing so the transition happens later.

Updated the timing so that the transition now occurs after 32 seconds (8 seconds previously), making the progress display less abrupt.

> > * The main spinner messages seem developer oriented now. Previously, it was something like "fetching data from GitHub..." but now it says things like `add_remote_data`, `get_github_metadata` and so on. I know git-cliff is mostly directed at developers, but I still think that we could avoid programmatic-looking info-level messages. Wdyt?
> 
> There is actually a reason for the current behavior. Since I implemented using `tracing-indicatif`, most of the macros ended up being defined in the `git-cliff-core` side. Because that crate also has an SDK role, I intentionally avoided defining end-user-facing messages there and instead exposed the function names directly.
> 
> That said, I don't have any objection to making the messages more human-readable. In the future, when we have a chance to refactor, I'd like to explore the idea of making `git-cliff-core` more "log-less" and keeping user-facing messaging closer to the CLI layer.
> 
> For now, I agree with your point and will update the messages to be more readable.

Implemented a macro to output non-developer friendly progress messages in tracing spans.
Previously, messages were developer-oriented. Now, the macro allows progress updates to display user-friendly text while still working with the `tracing-indicatif` integration.

git-cliff-core/src/lib.rs
```diff
@@ -50,3 +50,19 @@ pub const DEFAULT_CONFIG: &str = "cliff.toml";
 pub const DEFAULT_OUTPUT: &str = "CHANGELOG.md";
 /// Default ignore file.
 pub const IGNORE_FILE: &str = ".cliffignore";
+
+/// Sets a human-readable message on the current progress bar span.
+/// This macro only has effect if the `tracing-indicatif` feature is enabled.
+#[doc(hidden)]
+#[macro_export]
+macro_rules! pb_msg {
```
orhun: nit: maybe rename to sth like `log_progress` or `log_pg`.

`pb` makes me think of `public` before `progressbar` to me.

orhun (commented):
Thanks for making the changes! This is in much better state now.

There are still 2 points that I want to improve:

1. The main log message while fetching:

Previously:

> ▹▹▹▹▸ Retrieving data from GitHub... (ratatui/ratatui)

Now it is:

> ⠙ Adding remote data to the changelog                                                                                                                                                       
 ↳ ⠙ Fetching GitHub metadata for the changelog                                                                                                                                          
  ↳ ⠋ Fetching all pull requests from GitHub                                                                                                                                               
  ↳ ⠋ Fetching all commits from GitHub 

The problem is we no longer show the repository name here, so it isn't clear where it is fetching the data from.

2. The colors

The log messages are always in gray color for me, which makes them less readable. Previously it was white.

Maybe we could use white again, or use colors based on the module depth? (can go from white -> gray -> dark gray etc.)

orhun:
I had another look at this and it LGTM. Just need to make the changes that I mentioned above :)

ognis1205:
Hey @orhun , thanks for the review!

I resolved all your review comments. Please review again when you have some time. Thanks!

orhun:
Cool, thanks for the changes!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

Could we maybe include the repository link in the first line? If we do it for each line then it seems a bit repetitive.

What I mean is:

```
⠴ Adding remote data to the changelog (orhun/git-cliff)                                                                                     
 ↳ ⠴ Fetching GitHub metadata for the changelog    
  ↳ ⠼ Fetching all pull requests from GitHub
  ↳ ⠼ Fetching all commits from GitHub 
```

Also, could we maybe only make the first line white and the other ones dim?

ognis1205:
> Could we maybe include the repository link in the first line? If we do it for each line then it seems a bit repetitive.
> 
> What I mean is:
> 
> ```
> ⠴ Adding remote data to the changelog (orhun/git-cliff)                                                                                     
>  ↳ ⠴ Fetching GitHub metadata for the changelog    
>   ↳ ⠼ Fetching all pull requests from GitHub
>   ↳ ⠼ Fetching all commits from GitHub 
> ```

I've updated it so the repository link is only shown at the top level to avoid repetition.

> Also, could we maybe only make the first line white and the other ones dim?

Regarding the dim color for nested lines, it doesn't seem very straightforward to achieve with the current `tracing-indicatif` API. It might be possible with a hacky workaround, but I'd prefer to keep it as-is for now.

Cyclonit (approved):
Truth be told, I would have never thought of using tracing in a "small" CLI like git-cliff, but I like it. It clearly cleans up situations where logging simply doesn't suffice for modelling control flow. The changes from "log before & log after" to actual spans in `changelog.rs` are a good example.

LGTM

Issue #1318, opened 2025-12-16
git-cliff fails to parse context it generated itself due to missing field azure_devops


taladar:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description of the bug

I am trying to add a few variables for use in templates (in particular package.metadata.deb.revision to generate a Debian changelog) so I used the --context and --from-context options with some jq in between in my release script to add those. However, rather unexpectedly, --from-context fails due to a missing field azure_devops which I neither use in my cliff.toml, in CLI options or anywhere else.

The context.json file contains a field azureDevops in a few places so I assume there is some issue where the source code produces that but expected azure_devops when parsing.

### Steps To Reproduce

1. git cliff ... --context --output context.json
2. git cliff ... --from-context context.json
3. Get error like JsonError(Error("missing field `azure_devops`", line: 58, column: 7))

### Expected behavior

I would expect it to parse the context.json it just produced itself.

### Software information

- Operating system: Gentoo Linux
- Rust version: 1.92.0
- Project version: 2.11.0

taladar:
I tried replacing all occurrences of azureDevops in the JSON file with azure_devops but that did not fix it so it might be a bit more complex after all than just a mismatch between serialization and deserialization.

taladar:
The line and column of the error point to the missing field being in the commit object, and indeed, compared to your .github/fixtures/test-from-context/context.json in the repository, the commit just has github, gitlab, gitea and bitbucket fields, but not azure_devops field. So it seems the error is about the one holding username and pull request data, not the azure_devops field that contains contributors only.

taladar:
I assume the problem is that there is no azure_devops in commit.rs around line 485 (in the manual Serialize implementation for Commit) but Deserialize is derived (on line 97 in the same file) and so Serialize does not write it but Deserialize wants it.

taladar:
I tried adding that and it fixes the error in the commit object but now I get another one, also about a missing azure_devops field so while this is part of it it is not the entire story.

orhun:
Hey, thanks for reporting!

I think the solution might be making the Azure DevOps fields an `Option<T>`

I'll look more into this later. Feel free to submit a PR btw!

taladar:
I was about to but then I noticed that it still fails so I am looking for the type that is missing that other azure_devops field.

orhun:
I'm not sure what you mean. Maybe create a PR with your changes or share logs?

taladar:
[...]

PR #1319, opened 2025-12-17
chore(azure_devops)!: rename azureDevops variable to azure_devops

taladar:
fixes https://github.com/orhun/git-cliff/issues/1318

Basically the Commit serialization was missing the new azure_devops field entirely in its manual Serialize implementation and the Release serialization applied its renaming of all fields to camel case only on serialization, the other fields that would have been affected by that had manual rename attributes on the field level. It might make sense to rethink if renaming all fields to camel case makes sense at all if all fields with multiple words manually revert that anyway.

### DIFF
--- a/.github/fixtures/test-azure-devops-integration-custom-range/cliff.toml
+++ b/.github/fixtures/test-azure-devops-integration-custom-range/cliff.toml
@@ -24,11 +24,11 @@ body = """
     {%- endif %}
 {%- endfor -%}
 
-{% if azureDevops.contributors | filter(attribute="is_first_time", value=true) | length != 0 %}
+{% if azure_devops.contributors | filter(attribute="is_first_time", value=true) | length != 0 %}
   {% raw %}\n{% endraw -%}
   ### New Contributors
 {%- endif %}\
-{% for contributor in azureDevops.contributors | filter(attribute="is_first_time", value=true) %}
+{% for contributor in azure_devops.contributors | filter(attribute="is_first_time", value=true) %}
   * @{{ contributor.username }} made their first contribution
     {%- if contributor.pr_number %} in \
       [#{{ contributor.pr_number }}]({{ self::remote_url() }}/pullrequest/{{ contributor.pr_number }}) \
--- a/.github/fixtures/test-azure-devops-integration/cliff.toml
+++ b/.github/fixtures/test-azure-devops-integration/cliff.toml
@@ -24,11 +24,11 @@ body = """
     {%- endif %}
 {%- endfor -%}
 
-{% if azureDevops.contributors | filter(attribute="is_first_time", value=true) | length != 0 %}
+{% if azure_devops.contributors | filter(attribute="is_first_time", value=true) | length != 0 %}
   {% raw %}\n{% endraw -%}
   ### New Contributors
 {%- endif %}\
-{% for contributor in azureDevops.contributors | filter(attribute="is_first_time", value=true) %}
+{% for contributor in azure_devops.contributors | filter(attribute="is_first_time", value=true) %}
   * @{{ contributor.username }} made their first contribution
     {%- if contributor.pr_number %} in \
       [#{{ contributor.pr_number }}]({{ self::remote_url() }}/pullrequest/{{ contributor.pr_number }}) \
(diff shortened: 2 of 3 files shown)


git-cliff-core/src/release.rs
```diff
@@ -61,6 +61,7 @@ pub struct Release<'a> {
     pub bitbucket: RemoteReleaseMetadata,
     /// Contributors.
     #[cfg(feature = "azure_devops")]
+    #[serde(rename = "azure_devops")]
```
orhun, line 64: Why this is needed? 🤔 Aren't we already serializing as `azure_devops` here?

orhun: Ah ok nvm just saw your other comment.

orhun:
The CI seems to be failing, can you look into it?

taladar:
I have no idea what your tests are supposed to do, all I did is add the serialization of the field to Commit and change the serde attribute on the field in Release, not sure what any of that would have to do with the three failing tests that also produce no output. Not familiar with the GitHub interface for tests.

orhun:
Seems like there is a breaking change:

> Variable `azureDevops.contributors` not found in context while rendering 'body'

This variable is now `azure_devops`...

Why are we renaming this to snake_case again?

taladar:
You used the annotation

```
#[serde(rename_all(serialize = "camelCase"))]
```

on the Release struct. This renames all the fields to camelCase but only on serialization, not deserialization.

All the other fields with multiple words (commit_id, commit_ranges, submodule_commits) have

```
#[serde(rename = "submodule_commits")]
```

annotations so I added one of those to this field too since I had to get it to use the same renaming on serialization and deserialization somehow and that was the way that affected the smallest number of fields.

taladar:
Obviously another option would be to replace the rename_all with something like

```
#[serde(rename_all = "camelCase")]
```

to affect both serialization and deserialization but I don't know if that would affect any backwards compatibility or future plans for how you want to evolve this struct, it seems you are in the middle of a transition to camelCase?

taladar:
Technically the original change to expect a new field in the context JSON was not semver compatible either. I would assume any compatibility here should primarily be between the JSON generate by --context and consumed by --from-context within any given version which is currently not compatible. I don't particularly care which bits are renamed to resolve this but the current solution of producing camelCase and consuming snake_case is not sensible.

And of course the other fix should be included either way, the one where Commit does not serialize the field at all currently.

git-cliff-core/src/commit.rs
```diff
@@ -483,6 +483,8 @@ impl Serialize for Commit<'_> {
         commit.serialize_field("gitea", &self.gitea)?;
         #[cfg(feature = "bitbucket")]
         commit.serialize_field("bitbucket", &self.bitbucket)?;
+        #[cfg(feature = "azure_devops")]
+        commit.serialize_field("azure_devops", &self.azure_devops)?;
```
ognis1205, line 491: ~~Using Camel Case in JSON is the established convention. Can we change this to `azureDevops` instead?~~

This looks good to me.

git-cliff-core/src/release.rs
ognis1205, line 64: ~~Do we really need this change?~~

This looks good to me.

taladar:
It is actually two separate fixes. The one you mention and the other problem is that the Release struct has a rename_all annotation that only applies to the derived Serialize but not to the derived Deserialize implementation, keeping the snake_case name for Deserialize while using the camelCase name in Serialize.

ognis1205:
> It is actually two separate fixes. The one you mention and the other problem is that the Release struct has a rename_all annotation that only applies to the derived Serialize but not to the derived Deserialize implementation, keeping the snake_case name for Deserialize while using the camelCase name in Serialize.

Ah, right. Thanks for the clarification, @taladar .

@orhun , why was this implemented asymmetrically? I would have expected something like:

```rust
#[serde(rename_all = "camelCase")]
```

so that both `Serialize` and `Deserialize` use the same naming.

See:

https://github.com/orhun/git-cliff/pull/1319#pullrequestreview-3601790866

orhun:
I made the necessary changes here and going to merge this. It is a breaking change but it's fine I believe (for the sake of consistency).

I think the reason why this bug occurred is I simply overlooked the serialization details when I first reviewed the Azure DevOps integration PR. Sorry about that.

PR #1321, opened 2025-12-19
feat(args): add offline flag

### LINKED ISSUE #1095: Add offline/local version of --bumped-version

barskern:
## Description

Add a `--offline` flag to enable running e.g. `--bumped-version` without contacting any external services, even if they are configured. Closes #1095.

## Motivation and Context

See #1095 .

## How Has This Been Tested?

Local unit and integration tests, including building and running the binary in my local repo.

**NOTE!** Three unrelated tests were failing when I started my implementation:

```
repo::test::get_latest_tag
repo::test::git_tags
repo::test::resolves_existing_tag_with_name_and_message
```

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/git-cliff-core/src/changelog.rs
+++ b/git-cliff-core/src/changelog.rs
@@ -254,8 +254,8 @@ impl<'a> Changelog<'a> {
                 release
                     .previous
                     .as_ref()
-                    .and_then(|release| release.version.as_ref())
-                    == Some(skipped_tag)
+                    .and_then(|release| release.version.as_ref()) ==
+                    Some(skipped_tag)
             }) {
                 if let Some(previous_release) = self.releases.get_mut(release_index + 1) {
                     previous_release.previous = None;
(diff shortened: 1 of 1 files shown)


git-cliff/src/args.rs
```diff
@@ -359,6 +359,9 @@ pub struct Opt {
     /// Load TLS certificates from the native certificate store.
     #[arg(long, help_heading = Some("FLAGS"), hide = !cfg!(feature = "remote"))]
     pub use_native_tls: bool,
+    /// Use only local git history in actoins
```
orhun:
```suggestion
    /// Use only local git history in actions
```

git-cliff-core/src/config.rs
```diff
@@ -177,6 +177,11 @@ mod serde_pattern {
 /// Remote configuration.
 #[derive(Default, Debug, Clone, Serialize, Deserialize)]
 pub struct RemoteConfig {
+    /// Run generation fully offline
+    // NOTE! Intended to be set from the CLI args, as having it in
+    // the config file doesn't really makes sense IMO.
```
orhun: I think we can also support it the configuration file :)

```suggestion
```

barskern:
> Looks good! Also we need to update the documentation/website about this new flag :)

Thank you for your patience here! I've added your suggested changes and a small section to the website. I was also looking for a way to test it, but in the way the `remote` module is implemented for the moment I wasn't able to find a clean way to do it. It seems like the remote implementations are not really tested for the moment?

orhun:
Check out the [test fixtures here](https://github.com/orhun/git-cliff/tree/main/.github/fixtures). I believe we can copy one of the GitHub integration ones, run it with `--offline` and compare the result (`expected.md`).

barskern:
It was good that you requested me to create a test! My ignorance thought this was a trivial change, but making the test made me catch a edge-case. Thank you!

W1M0R:
Thank you for this feature @barskern and @orhun!

Issue #692, opened 2024-06-11
Defining Pre-Release Versions: Beta, Alpha, RC, and Version Bumping (`--bump`)


bukowa:
### Is there an existing issue or pull request for this?

- [X] I have searched the existing issues and pull requests

### Feature description

Releasing beta, RC (release candidate), or alpha versions is a common standard practice, especially when we want to test the entire build system. Sometimes these releases help ensure that the build scripts execute correctly and provide users with a preview version (e.g., a beta version), which is essentially complete but still in testing.

Sometimes, releasing a release candidate is necessary to verify the functionality of build scripts. It would be beneficial to standardize this process, allowing for the automatic generation of changelogs for these versions and their subsequent integration into the main release. Combining this with version bumping, while maintaining logical and configurable control, would be highly advantageous.

### Desired solution

Simple idea for bumping:
```bash
git-cliff --bump-beta
git-cliff --bump-alpha
git-cliff --bump-rc
```
No simple idea for changelogs but see :
- https://github.com/orhun/git-cliff/issues/588

This doesn't seem so straightforward as everyone has their standards and this somehow should be customizable?

### Alternatives considered

Manual bumping.

### Additional context

Given the variety of standards used by different projects, it would be beneficial to provide customizable configuration keys for tags like rc, alpha, and beta. This would allow users to tailor the versioning system to their specific requirements.
- https://peps.python.org/pep-0440/

````toml
bump_beta = ...<format>
bump_alpha = ...<format>
````
OR
````toml
[rc]
header = "Release Candidate Changes"
label = "RC"
# Configuration for alpha versions
[alpha]
header = "Alpha Version Changes"
label = "Alpha"
# Configuration for beta versions
[beta]
header = "Beta Version Changes"
label = "Beta"

bukowa:
It's not so easy for ex `release-please` from google issue regarding this: https://github.com/googleapis/release-please/pull/2051

I just tried to release `tauri` app but automatic build fails because `wix` does not support anything other than numbers in version https://github.com/tauri-apps/tauri/issues/5286#issuecomment-1259235117

PR #1345, opened 2026-01-16
feat(args): add the pre-release option

anagno:
Add the possibility of  specifying pre-releases via the cli arguments. Now it is possible to specify in the cli `--pre-release [type]` and the bump version will be a prerelease

## Description

Many projects have the need of creating pre-releases for testing with users, before the final release. This PR implements and extra cli argument that the user can specify the type of the pre-release and then changelogs for pre-releases will be generated.

It makes sure that the calculation of the pre-release is correct. So if there already an `rc` version and the user specifies via the cli `--pre-release alpha`, then the calculcation will be correct and it will increment the rc version

There was a [similar issue in the past](https://github.com/orhun/git-cliff/issues/692), but it was never planned. So if it is not of interests feel free just to close the PR.

To make it work with `ingore_tags`  I had to refactor a bit the `process_releases` function. Now it happens in 2 steps, so the calculation of the next version happens first, and then we filter out the releases that we do not want.

## Motivation and Context

Give the possibility of specifying pre-releases, so that projects can create changelogs for these releases

## How Has This Been Tested?

I included a couple of new UTs and tested it manually by invonking the git-cliff cli. For example:

```
cargo run -- --bumped-version --pre-release rc
v2.12.0-rc.0
```

## Types of Changes

- [X] New feature (non-breaking change which adds functionality)
- [ ] Other 

## Checklist:

I have not yet gone through all the points ( and I have not yet update the docs, because I wanted first to get some feedback :) )

- [X] My code follows the code style of this project.
- [X] I have formatted the code with [rustfmt](https://github.com/rust-lang/rustfmt).
- [X] I checked the lints with [clippy](https://github.com/rust-lang/rust-clippy).
- [X] All new and existing tests passed.

### DIFF
--- a/git-cliff-core/src/changelog.rs
+++ b/git-cliff-core/src/changelog.rs
@@ -2,7 +2,7 @@ use std::collections::HashMap;
 use std::io::{Read, Write};
 use std::time::{SystemTime, UNIX_EPOCH};
 
-use crate::commit::Commit;
+use crate::commit::{Commit, Range};
 use crate::config::{Config, GitConfig};
 use crate::error::{Error, Result};
 use crate::release::{Release, Releases};
@@ -240,11 +240,13 @@ impl<'a> Changelog<'a> {
         Ok(())
     }
 
-    /// Processes the releases and filters them out based on the configuration.
+    /// Processes the releases and skip the releases specified on the configuration.
     fn process_releases(&mut self) {
         log::debug!("Processing {} release(s)", self.releases.len());
         let skip_regex = self.config.git.skip_tags.as_ref();
         let mut skipped_tags = Vec::new();
+
+        // Filtering out all the skipped releases and updating the previous pointers
         self.releases = self
             .releases
             .clone()
@@ -258,20 +260,8 @@ impl<'a> Changelog<'a> {
                         return false;
                     }
                 }
-                if release.commits.is_empty() {
-                    if let Some(version) = release.version.clone() {
-                        log::debug!("Release doesn't have any commits: {version}");
-                    }
-                    match &release.previous {
-                        Some(prev_release) if prev_release.commits.is_empty() => {
-                            return self.config.changelog.render_always;
-                        }
-                        _ => return false,
-                    }
-                }
                 true
             })
-            .map(Release::with_statistics)
             .collect();
         for skipped_tag in &skipped_tags {
             if let Some(release_index) = self.releases.iter().position(|release| {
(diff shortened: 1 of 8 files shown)


anagno:
@orhun Is this PR of interest, or should I close it?

orhun:
Hey, apologies for the delay. I'll get back to this soon :)

ognis1205:
Thanks for opening this PR and working on this @anagno !

I’m currently neutral about supporting pre-releases in some form.

One thing that comes to mind, though, is that Semantic Versioning does not define any specific format or naming convention for pre-release identifiers (see https://semver.org/#spec-item-9). While `alpha` / `beta` / `rc` are commonly used, they are still just conventions and not part of the specification.

Because of that, supporting only these specific identifiers (as proposed in this PR) might be somewhat limiting and not fully generic. If we decide to move forward with accepting this feature, I'd personally prefer a more flexible approach that allows arbitrary pre-release identifiers rather than restricting it to a predefined set, although I'm not entirely sure how feasible that would be within the current implementation.

Just my first-glance feedback. Curious to hear if others have thoughts or alternative ideas from @anagno on how we might approach this.

ognis1205 (commented):
Thanks for opening this PR and working on this @anagno !

I’m currently neutral about supporting pre-releases in some form.

One thing that comes to mind, though, is that Semantic Versioning does not define any specific format or naming convention for pre-release identifiers (see https://semver.org/#spec-item-9). While alpha / beta / rc are commonly used, they are still just conventions and not part of the specification.

Because of that, supporting only these specific identifiers (as proposed in this PR) might be somewhat limiting and not fully generic. If we decide to move forward with accepting this feature, I'd personally prefer a more flexible approach that allows arbitrary pre-release identifiers rather than restricting it to a predefined set, although I'm not entirely sure how feasible that would be within the current implementation.

Just my first-glance feedback. Curious to hear if others have thoughts or alternative ideas from @anagno on how we might approach this.

- https://git-cliff.org/docs/usage/bump-version#bump-to-a-specific-version-type

Would extending the existing `--bump` option to support pre-releases achieve a similar result? 🤔

anagno:
I tried to follow a similar logic to the approach used in the the [commit-and-tag-version]( https://github.com/absolute-version/commit-and-tag-version/blob/master/README.md?plain=1#L276) project 

Regarding generalization: we can extend the behavior, but we are still constrained by what the next_version crate supports, since that crate is responsible for computing the next version.  I can make some more experiments, but I think it is feasible. Should it just accept any string and then pass it as parameter?

The decision not to use the existing `--bump` flag was intentional. My thinking was that a user might want to request both a major-level bump and a pre-release at the same time. In that case, combining everything into a single flag felt limiting, so I introduced a separate flag for pre-releases.

anagno:
I pushed a commit that generalizes the implementation. If no pre-release type is provided via the CLI, `alpha` is used as the default, due to limitations in the `next_version` crate.

ognis1205 (commented):
> I tried to follow a similar logic to the approach used in the the [commit-and-tag-version](https://github.com/absolute-version/commit-and-tag-version/blob/master/README.md?plain=1#L276) project

Thanks for explaining the rationale. That helps a lot. I understand the background and why you chose to follow that approach.

> Regarding generalization: we can extend the behavior, but we are still constrained by what the next_version crate supports, since that crate is responsible for computing the next version. I can make some more experiments, but I think it is feasible. Should it just accept any string and then pass it as parameter?

`next_version` already supports prerelease increments, although the incrementing behavior is defined by its own rules. Since there isn't a strict specification governing prerelease value semantics or how they should be incremented, we would inevitably be choosing one interpretation anyway.

Given that, and to keep responsibilities clearly separated, I think it would be more consistent with `git-cliff`'s existing design to rely on `next_version` for the increment logic instead of introducing custom handling.

> The decision not to use the existing --bump flag was intentional. My thinking was that a user might want to request both a major-level bump and a pre-release at the same time. In that case, combining everything into a single flag felt limiting, so I introduced a separate flag for pre-releases.

This is probably the most debatable part of the change.

After reconsidering it, I agree that introducing a separate flag for prereleases can make sense if we want to support combinations like a major bump together with a prerelease. However, I still think the new flag should conceptually depend on `--bump`, rather than behave as a fully independent option.

For example, something along the lines of:

```rust
#[derive(Parser)]
struct Cli {
    #[arg(long, value_enum)]
    bump: Option<BumpOption>,

    #[arg(long, required_if_eq("bump", "prerelease"))]
    prerelease: Option<String>,
}
```

As you mentioned, a user might want to request both a major level bump and a prerelease at the same time. Precisely because of that, the interaction between these flags becomes semantically significant and could easily shift user expectations around the existing `--bump` behavior.

Since this directly affects how version computation is interpreted, and potentially interferes with the current `--bump` semantics, I think it would be better to step back and discuss the intended specification and user-facing behavior in an issue first before finalizing the implementation.

anagno:
@ognis1205 Should I create the issue or will you do it ?

ognis1205:
> @ognis1205 Should I create the issue or will you do it ? 

Could you create it?
I think it would be helpful since you have the full context.

anagno:
@ognis1205 @orhun I added a couple of UTs documenting what is happening with the `next_version` and the new added flag. More or less trying to implement some of the cases mentioned in the [issue](https://github.com/orhun/git-cliff/issues/1380#issuecomment-3958975332)

Issue #1327, opened 2026-01-01
Add --quiet flag


ajiho:
### Is there an existing issue or pull request for this?

- [x] I have searched the existing issues and pull requests

### Feature description

For projects that integrate git-cliff midway, the console can produce a lot of warnings. I’m not sure if there’s already an option to address this—if there is, feel free to disregard this.

<img width="1492" height="335" alt="Image" src="https://github.com/user-attachments/assets/74dd2e9e-cf57-4847-b654-e2d09ff20331" />

### Desired solution

As mentioned above

### Alternatives considered

no

orhun:
Hmm, maybe we could add a `--quiet` option in contrast to `--verbose`.

You can disable logging via unsetting the `RUST_LOG` environment variable.

lud-wj:
You could log if it is not filtered ?

I have some commits fully in parenthesis like `(typo)` or `(cliff)` or `(changelog)` for stuff that does not impact really the project or the docs and is purely dev centric. I'd really like not to have warnings for those.

I added RUST_LOG=0 in my env but I wonder if that can impact other programs installed with cargo install.

ognis1205:
Hi all,

Regarding this issue, the root cause is likely the changes I introduced in [PR #1244](https://github.com/orhun/git-cliff/pull/1244/files#diff-1c527f230d9fe61c0ffca65b8c03fe724a07753b8caeea26c05271e2688a9813R91-R100), where commits that do not follow the conventional commit specification were being logged as `WARN`. 

This is because `git-cliff` is designed to work based on conventional commits, so the `WARN` was intended to notify users about commits that do not match the expected format.

For this issue, I see a few possible approaches:

1. Implement a `--quiet` flag to suppress logs entirely.
2. Revert the log level from `WARN` to `TRACE`.
3. Revert the log level to `TRACE` but still output a single `WARN` summarizing the number of commits that did not match the conventional commit format (like statistics).

I apologize for any inconvenience this may have caused and appreciate any feedback on the preferred approach.

lud-wj:
@ognis1205 It's not as bad, it just adds some noise but it's not breaking anything.

A flag in the config to not emit a `WARN` could be possible? Otherwise option 3 seems fine indeed.

ognis1205:
> [@ognis1205](https://github.com/ognis1205) It's not as bad, it just adds some noise but it's not breaking anything.
> 
[...]

PR #1355, opened 2026-01-19
feat: implement commit processing summary

### LINKED ISSUE #1354: Tracking: emit a summary warning for skipped commits due to parse errors (profiling)

### Feature description

Currently, when commit parsing fails, detailed error logs are emitted for each failure.
In repositories with many malformed or unsupported commits, this can lead to very noisy logs at the `warn` level, making it harder for users to notice the overall impact or other important warnings.

At the same time, completely suppressing these logs would reduce debuggability when users actually want to understand why commits were skipped.

### Desired solution

Introduce a two-level logging approach for commit parse errors:

- [x] Downgrade individual commit parse error logs to `trace`
- [ ] Emit a single warn-level summary message, for example:

```code
WARN: 5 commits were skipped due to parse errors (run with --verbose for details)
```

This makes the high-level outcome immediately visible, while still allowing interested users to inspect the detailed reasons by enabling verbose / `trace` logging.

### Alternatives considered

- Keep all parse error logs at `warn` level
- Silently ignore parse errors
- Add a new CLI flag to control verbosity for parse errors

### Additional context

- https://github.com/orhun/git-cliff/issues/1352
- https://github.com/orhun/git-cliff/issues/1327

ognis1205:
## Description

Adds aggregated commit processing summaries, logging errored commits in a concise way.
Errors are categorized by `CommitProcessingErrorKind` (`Io`, `Parse`, `Json`, `Field`, `Group`, `Skipped`, `Other`).
The summary includes total commits processed and counts per error kind, and emits `warning` or `info` messages depending on severity.

Closes #1354 

## Motivation and Context

https://discord.com/channels/1093977388892819553/1093978266135707668/1462836253295771668

## How Has This Been Tested?

- Implemented unit tests.
- Manual testing with sample commits (`git-cliff` repository) to verify correct aggregation and logging output.
- Checked that both skipped commits and errors appear in the summary with the correct log level.

Screenshot 2026-01-20 at 15 12 05

## Types of Changes

- [ ] Other 

## Checklist:

  - [x] `cargo test`

### DIFF
--- a/git-cliff-core/src/changelog.rs
+++ b/git-cliff-core/src/changelog.rs
@@ -16,6 +16,7 @@ use crate::remote::gitea::GiteaClient;
 use crate::remote::github::GitHubClient;
 #[cfg(feature = "gitlab")]
 use crate::remote::gitlab::GitLabClient;
+use crate::summary::Summary;
 use crate::template::Template;
 
 /// Changelog generator.
@@ -90,10 +91,18 @@ impl<'a> Changelog<'a> {
     }
 
     /// Processes a single commit and returns/logs the result.
-    fn process_commit(commit: &Commit<'a>, git_config: &GitConfig) -> Option<Commit<'a>> {
+    fn process_commit(
+        commit: &Commit<'a>,
+        git_config: &GitConfig,
+        summary: &mut Summary,
+    ) -> Option<Commit<'a>> {
         match commit.process(git_config) {
-            Ok(commit) => Some(commit),
+            Ok(commit) => {
+                summary.record_ok();
+                Some(commit)
+            }
             Err(e) => {
+                summary.record_err(&e);
                 let short_id = commit.id.chars().take(7).collect::<String>();
                 let summary = commit.message.lines().next().unwrap_or_default().trim();
                 log::trace!("{short_id} - {e} ({summary})");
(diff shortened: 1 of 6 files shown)


ognis1205:
cc: @orhun

Screenshot 2026-01-20 at 15 12 05

ognis1205 (commented):
Instead of passing `Summary` down the call stack, would it be cleaner to inject it via a callback (no-op by default)? This keeps `process_commit` focused on commit processing and avoids coupling it to aggregation concerns.

orhun:
As pointed out in the last comment in https://github.com/orhun/git-cliff/issues/1327#issuecomment-3773306953, I don't think we should add more visible logs for the info level:

> INFO  git_cliff_core::changelog > Processed 18 commit(s) in total (`split_commits` option may cause duplicates)
> INFO  git_cliff_core::changelog > 2 commit(s) were skipped due to intentionally skipped commit(s) (run with `--verbose` for details)

I think these logs should be debug instead. WDYT?

> Instead of passing Summary down the call stack, would it be cleaner to inject it via a callback (no-op by default)? This keeps process_commit focused on commit processing and avoids coupling it to aggregation concerns.

Yeah, it might be a good idea. I'm also fine with the current approach as well. I'll leave this up to you to refactor if you want :)

ognis1205:
> As pointed out in the last comment in [#1327 (comment)](https://github.com/orhun/git-cliff/issues/1327#issuecomment-3773306953), I don't think we should add more visible logs for the info level:
> 
> > INFO  git_cliff_core::changelog > Processed 18 commit(s) in total (`split_commits` option may cause duplicates)
> > INFO  git_cliff_core::changelog > 2 commit(s) were skipped due to intentionally skipped commit(s) (run with `--verbose` for details)
> 
> I think these logs should be debug instead. WDYT?

I don't have strong objections either, so I'm fine with changing all of these to `debug`.

The main reason some of them were at `info` was to give users a clearer sense of progress, but if/once the migration to `tracing` is successful, that concern should be addressed more cleanly there.

> 
> > Instead of passing Summary down the call stack, would it be cleaner to inject it via a callback (no-op by default)? This keeps process_commit focused on commit processing and avoids coupling it to aggregation concerns.
> 
> Yeah, it might be a good idea. I'm also fine with the current approach as well. I'll leave this up to you to refactor if you want :)

Since this refactoring doesn't involve any `pub` APIs and wouldn't change the UI/UX, and we're currently trying to move quickly toward a patch release, I'd prefer to keep the current approach for now.

In any case, I do think the codebase will benefit from some refactoring down the line, so I'm happy to propose incremental refactoring PRs when the timing is better.

git-cliff-core/src/commit.rs
```diff
@@ -283,6 +283,7 @@ impl Commit<'_> {
         filter: bool,
     ) -> Result<Self> {
         let lookup_context = serde_json::to_value(&self).map_err(|e| {
+            // TODO: This should be updated as a JsonError?
```
ognis1205: What do you think about this part @orhun ?

orhun: Hmm, yes (if there isn't any side effects)

git-cliff-core/src/summary.rs
```diff
@@ -0,0 +1,261 @@
...
+#[derive(Default, Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Summary {
+    /// The total number of commits that were processed.
+    pub processed: usize,
+    /// The number of commits grouped by processing error kind.
+    ///
+    /// Each entry represents how many commits fell into a particular
+    /// [`CommitProcessingErrorKind`] during processing.
+    pub by_kind: HashMap<CommitProcessingErrorKind, usize>,
```
orhun: We could use `indexmap` here

git-cliff-core/src/commit.rs
ognis1205: Since our top priority right now is getting the patch release out, so I will keep this as-is for the moment. I'll remove the comment and avoid making any changes here for now.

We can revisit this later once the release is out, assuming there are no side effects to consider.

ognis1205:
Thanks for the review @orhun ! I've addressed your comments and pushed the updates.

git-cliff-core/src/changelog.rs
```diff
@@ -200,12 +210,34 @@ impl<'a> Changelog<'a> {
...
+            summary.processed
+        );
+        for (&kind, &count) in &summary.by_kind {
+            if count == 0 {
+                continue;
+            }
+            let message = format!(
+                "{count} commit(s) were skipped due to {kind}(s) (run with `--verbose` for \
+                 details)",
+            );
```
orhun, line 232:
```suggestion
            let message = format!(
                "{count} commit(s) were skipped due to {kind}(s) (run with `-vv` for details)",
            );
```

orhun (approved):

PR #1360, opened 2026-01-30
fix: handling of nested annotated tags

GChernikov:
Fix handling of nested annotated tags (tags pointing to other tags instead of commits directly).

## Description 
 
When an annotated tag points to another annotated tag (instead of directly to a commit), git-cliff silently skips it. This PR fixes the issue by using `obj.peel(ObjectType::Commit)` to follow the entire chain of tag objects until reaching the final commit. 

## Motivation and Context

In some workflows, tags are "promoted" by creating new tags pointing to existing ones. For example: 

v1.0.0-stable -> v1.0.0-rc (tag) -> v1.0.0-staging (tag) -> commit

The current code uses `tag.target().into_commit()` which only resolves one level of indirection. When the target is another tag object (not a commit), `into_commit()` fails and the tag is silently skipped from the changelog.

## How Has This Been Tested?

Added `git_nested_tags` test that:       
  1. Creates a temp repository with an empty commit
  2. Creates an annotated tag `v1.0.0-staging` pointing to the commit                                                                                                                                      
  3. Creates a nested annotated tag `v1.0.0-stable` pointing to `v1.0.0-staging`                                                                                                                           
  4. Verifies that `v1.0.0-stable` correctly resolves to the original commit                                                                                                                               
                                                                                                                                                                                                           
The test fails without the fix and passes with it.

Before fix - nested tag is missing:                                                                                                                                                                      
  ```bash                                                                                                                                                                                                  
  $ git tag -a v1.0.0-staging -m "staging"                                                                                                                                                                 
  $ git tag -a v1.0.0-stable -m "stable" $(git rev-parse v1.0.0-staging)                                                                                                                                   
  $ git cliff --tag-pattern ".*-stable$"                                                                                                                                                                   
  # v1.0.0-stable is missing from output
  ```                                                                                                                                                                   

After fix - nested tag is included correctly. 

## Types of Changes

<!--- What types of changes does your code introduce? Put an `x` in all the boxes that apply: -->

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Other <!--- (provide information) -->

## Checklist:

<!--- Go over all the following points, and put an `x` in all the boxes that apply.   -->
<!--- If you're unsure about any of these, don't hesitate to ask. We're here to help! -->

  - [x] `cargo test`

<!--- Thank you for contributing to git-cliff! ⛰️  -->

### DIFF
--- a/.github/fixtures/test-nested-tags/cliff.toml
+++ b/.github/fixtures/test-nested-tags/cliff.toml
@@ -1,0 +1,26 @@
+[changelog]
+header = ""
+body = """
+{% if version %}\
+{% set clean_version = version | replace(from="v", to="") | replace(from="-stable", to="") %}\
+## [{{ clean_version }}] - {{ timestamp | date(format="%Y-%m-%d") }}
+{% else %}\
+## [unreleased]
+{% endif %}\
+{% for group, commits in commits | group_by(attribute="group") %}
+### {{ group | upper_first }}
+{% for commit in commits %}
+- {{ commit.message }}
+{%- endfor %}
+{% endfor %}
+"""
+trim = true
+
+[git]
+conventional_commits = true
+filter_unconventional = true
+commit_parsers = [
+  { message = "^feat", group = "Features" },
+  { message = "^fix", group = "Bug Fixes" },
+]
+tag_pattern = "v.*-stable$"
--- a/.github/fixtures/test-nested-tags/cliff.toml
+++ b/.github/fixtures/test-nested-tags/cliff.toml
@@ -2,8 +2,7 @@
 header = ""
 body = """
 {% if version %}\
-{% set clean_version = version | replace(from="v", to="") | replace(from="-stable", to="") %}\
-## [{{ clean_version }}] - {{ timestamp | date(format="%Y-%m-%d") }}
+## [{{ version | replace(from="v", to="") }}] - {{ timestamp | date(format="%Y-%m-%d") }}
 {% else %}\
 ## [unreleased]
 {% endif %}\
@@ -23,4 +22,4 @@ commit_parsers = [
   { message = "^feat", group = "Features" },
   { message = "^fix", group = "Bug Fixes" },
 ]
-tag_pattern = "v.*-stable$"
+tag_pattern = "v.*"
--- a/git-cliff-core/src/repo.rs
+++ b/git-cliff-core/src/repo.rs
@@ -560,7 +560,7 @@ impl Repository {
             }
         }
         if !topo_order {
-            tags.sort_by(|a, b| a.0.time().seconds().cmp(&b.0.time().seconds()));
+            tags.sort_by_key(|a| a.0.time().seconds());
         }
         Ok(tags
             .into_iter()
--- a/git-cliff-core/src/repo.rs
+++ b/git-cliff-core/src/repo.rs
@@ -541,10 +541,11 @@ impl Repository {
                     message: None,
                 }));
             } else if let Some(tag) = obj.as_tag() {
-                if let Some(commit) = tag
-                    .target()
+                // Use peel to resolve nested tags to the final commit
+                if let Some(commit) = obj
+                    .peel(git2::ObjectType::Commit)
                     .ok()
-                    .and_then(|target| target.into_commit().ok())
+                    .and_then(|o| o.into_commit().ok())
                 {
                     if use_branch_tags && !self.should_include_tag(&head_commit, &commit)? {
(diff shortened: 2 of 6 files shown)


git-cliff-core/src/repo.rs
```diff
@@ -762,6 +763,58 @@ mod test {
         Ok(())
     }
 
+    #[test]
+    fn git_nested_tags() -> Result<()> {
```
orhun, line 814: Would also love to have a [test fixture](https://github.com/orhun/git-cliff/tree/main/.github/fixtures) for this so that we would see how this behaves in the changelog generation context.

orhun (commented):
thanks for the PR!

git-cliff-core/src/repo.rs
GChernikov: Added a test fixture `test-nested-tags` that covers this scenario.                                                                                                                                                                                                                                                                                                                                                  

The fixture creates a 3-level nested tag chain:                                                                                                                                                          
v0.1.0-stable → v0.1.0-rc → v0.1.0-staging → commit                                                                                                                                                      
                                                                                                                                                                                                           
With `tag_pattern = "v.*-stable$"`, only the nested `v0.1.0-stable` tag is matched. The test verifies that it correctly resolves to the underlying commit and appears in the changelog.                  
                                                                                                                                                                                                           
Without the fix, the output shows `[unreleased]` instead of `[v0.1.0-stable]` because the nested tag fails to resolve

GChernikov: Also updated the fixture to show this more clearly.

Now the template strips `v` prefix and `-stable` suffix, so the changelog shows clean versions:                                                                                                          

Tags: `v0.1.0-staging` → `v0.1.0-rc` → `v0.1.0-stable`                                                                                                                                                   
Output: `## [0.1.0] - 2022-04-06`                                                                                                                                                                        
                        
This better reflects the actual use case — promotion workflow with clean changelog output.

GChernikov: @orhun could you please review changes? Is there anything alse I can do to help you merge this PR?

GChernikov: @orhun thanks again! I’ve added the requested test fixture (`test-nested-tags`) and wired it into the fixture tests. All checks are green now. Could you please take another look when you have a moment? Happy to adjust anything.

git-cliff-core/src/repo.rs
```diff
@@ -762,6 +763,58 @@ mod test {
...
+        let (repo, temp_dir) = create_temp_repo();
+        let path = temp_dir.path();
+
+        Command::new("git")
+            .args(["commit", "--allow-empty", "-m", "init"])
```
orhun: Please add "--no-gpg-sign" argument otherwise this gets stuck for me (due to pinentry client).

.github/fixtures/test-nested-tags/cliff.toml
```diff
@@ -0,0 +1,26 @@
...
+trim = true
+
+[git]
+conventional_commits = true
+filter_unconventional = true
+commit_parsers = [
+  { message = "^feat", group = "Features" },
+  { message = "^fix", group = "Bug Fixes" },
+]
+tag_pattern = "v.*-stable$"
```
orhun: I think this test fixture is not testing the "peel" change introduced in this PR.

The pattern is checked before we could start peeling the object here:

https://github.com/orhun/git-cliff/blob/988e8638432dcfc762dfec3ea470eb82aab80380/git-cliff-core/src/repo.rs#L483

So it would be better to remove the `-stable` suffix from this configuration option and adjust the test fixture based on that.

orhun (commented):
Sorry for long delay, I just had some time to look at this again!

git-cliff-core/src/repo.rs
GChernikov: Done. Sorry for the long delay!

.github/fixtures/test-nested-tags/cliff.toml
GChernikov: Simplified the fixture: replaced the 3-level tag chain with a 2-level one (v0.1.0 → rc-1 tag object → commit) and changed tag_pattern to v.*, removing the -stable suffix entirely.

Could you take another look when you have a chance?

GChernikov:
Also fixed sort_by → sort_by_key in repo.rs and release.rs to satisfy the new clippy lint that appeared after rebasing onto main

orhun (approved):
LGTM!

PR #1368, opened 2026-02-05
fix(remote): GitLab API data models to deserialize safely

### LINKED ISSUE #1366: Panic when description is null in GitLab Merge Request JSON

### Description of the bug

When running git-cliff as part of our CI/CD pipeline, a panic occurs if the description field in a merge request’s metadata is null instead of an empty string.

``` BASH
thread 'main' (351) panicked at git-cliff-core/src/changelog.rs:590:18:
Could not get gitlab metadata: JsonError(Error("invalid type: null, expected a string", line: 1, column: 370937))
```

This issue started occurring after upgrading from git-cliff 2.10.1 to 2.11. It is still present in 2.12.

### Steps To Reproduce

1. Have a merge request in GitLab where "description": null
2. Run git-cliff

### Example JSON:
```JSON
{
  "id": 319131,
  "iid": 149,
  "project_id": 19011,
  "title": "MR Title",
  "description": null,
  ...
}
```

### Expected behavior

The description field should be treated as optional if it is null.

### Screenshots / Logs

_No response_

### Software information

- Operating system:
- Rust version:
- Project version: git-cliff 2.11 and 2.12 (issue occurs), 2.10.1 (working)

### Additional context

_No response_

ognis1205:
## Description

The current implementation of the GitLab Remote API data models diverges from the official OpenAPI specification in terms of optional/required fields:

https://gitlab.com/gitlab-org/gitlab/-/blob/master/doc/api/openapi/openapi.yaml

As a result, JSON ser/de errors occur in practice. This change fixes those mismatches to ensure reliable ser/de behavior.

Closes #1366 #687 

## Motivation and Context

The current implementation of the GitLab Remote API data models diverges from the official OpenAPI specification in terms of optional and required fields. This mismatch caused JSON ser/de errors, which could lead to panics when fetching data from the GitLab remote API.

This PR first fixes the optionality mismatches in the data models, and as a result resolves the panic that occurred during GitLab remote fetches.

Additionally, similar `JsonError`s should have also occurred in implementations prior to v2.10:

https://github.com/orhun/git-cliff/commit/b82221abd1981b6ecce8ab428fede8165ebb4246#diff-9d830e0be5fc20135aee3ba39934c06263208881475d426a20c85a3b7dbba385L242-L251

However, due to the use of `take_while`, these errors were unintentionally silenced, resulting in silent failures instead of visible errors. This fix therefore also addresses issues related to those previously hidden
JSON deserialization errors.

## How Has This Been Tested?

- `cargo test`

No additional tests were added, as this change only adjusts data model definitions and does not modify runtime logic.

N/A

## Types of Changes

- [ ] Other 

## Checklist:

  - [x] `cargo test`

### DIFF
--- a/git-cliff-core/src/remote/gitlab.rs
+++ b/git-cliff-core/src/remote/gitlab.rs
@@ -127,7 +127,7 @@ pub struct GitLabMergeRequest {
 impl RemotePullRequest for GitLabMergeRequest {
     fn number(&self) -> i64 {
         self.iid
-            .expect("Merge request iid is required for git-cliff semantics")
+            .expect("Merge request id is required for git-cliff semantics")
     }
 
     fn title(&self) -> Option<String> {


ognis1205:
One concern I have is that the way required/optional fields are modeled on the `git-cliff` side feels somewhat ad-hoc. While this PR should fix concrete ser/de issues, the overall data model design around optionality may not be entirely clean from a design perspective.

ognis1205:
`#[serde(default)]` would be a bit cleaner.

orhun:
Yeah, we could use `#[serde(default)]` I agree

README.md
```diff
@@ -1,3 +1,5 @@
+TEST
+
```
orhun:
```suggestion
```

orhun (approved):

ognis1205:
> Yeah, we could use `#[serde(default)]` I agree

Yeah, I initially thought `#[serde(default)]` would be enough, but it turns out inputs like `{ "description": null }` are possible, so `Option` is still needed.

To keep things consistent and improve UX, I made these fields `Option` across the board and moved the invariants to the usage sites via `expect`, rather than encoding them in comments on the struct fields.

ognis1205:
Note:

I think this PR likely fixes the issue in

https://github.com/orhun/git-cliff/issues/687

as well, though it'd be nice to get a quick confirmation from the community.

orhun:
Maybe leave a short comment in the related issue for testing?

ognis1205:
> Maybe leave a short comment in the related issue for testing?

I will leave a comment shortly.

git-cliff-core/src/remote/gitlab.rs
```diff
@@ -19,152 +19,149 @@ pub(crate) const TEMPLATE_VARIABLES: &[&str] = &["gitlab", "commit.gitlab", "com
...
-    pub web_url: String,
+    pub web_url: Option<String>,
     /// Labels
     pub labels: Vec<String>,
 }
 
 impl RemotePullRequest for GitLabMergeRequest {
     fn number(&self) -> i64 {
         self.iid
+            .expect("Merge request iid is required for git-cliff semantics")
```
orhun:
```suggestion
            .expect("Merge request id is required for git-cliff semantics")
```

Issue #1370, opened 2026-02-10
Docker publishes unstable builds with latest tag


Garbee:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description of the bug

The Docker image workflow currently publishes the `latest` tag even for non-release builds (schedule/branch builds). This makes `latest` point at unstable/nightly builds, which causes automatic update systems (like Dependabot, that follow `latest`) to upgrade to nightlies instead of maintaining a stable release.

### Steps To Reproduce

No real steps here, it's just from looking at the docker container history.

### Expected behavior

The `latest` tag should be applied to stable builds only. If a nightly test system is needed then it should be driven by a separate tag, such as `nightly` for example.

### Software information

Not applicable, this is a bug in the deployment pipeline.

### Additional context

The current workflow has `latest` as a fixed tag on every build.

https://github.com/orhun/git-cliff/blob/91e65f5bba474b680849adf09db7e1ad1decd153/.github/workflows/docker.yml#L35

This can be made conditional. Like so, `type=raw,value=latest,enable=${{ startsWith(github.ref, 'refs/tags/') }}`. This given the workflow's context will only enable the `latest` tag when the event reference is a tag.

If nightly testing is needed via a tag, then `nightly` can be introduced for all other builds, like this, `type=raw,value=nightly,enable=${{ github.ref == 'refs/heads/main' }}`.

Garbee:
In case it is accepted that this change is needed, I just drafted PR #1371 to apply the changes. Doing some testing to link to before opening it up to review.

orhun:
Hello, thanks for reporting this and the PR!

This sounds fair to me.
Additionally, maybe we could push another tag for the actual "latest" commit, e.g. it could be named `main`?

Garbee:
That is what the `nightly` recommendation at the end was. Yes, absolutely. I did find a bug in the workflow that I'm testing to fix while I'm here. Because currently one of the tests in the workflow doesn't *guarantee* it tests the image that just deployed. So a race condition could occur and the test is invalid. (Plus on PRs it is wasted time testing since those images are not deployed.)

Now, what you want the tag to be between `nightly` and `main`, I don't care. It's a one line change and I'm happy to do whatever. `nightly` is just more well understood as "unstable so don't trust it. But please test and make sure we don't break you."

Garbee:
Ok, #1371 is open for review. Changed the tag name to `main` as well.

PR #1371, opened 2026-02-10
feat(docker): apply latest tag to stable only

Garbee:
## Description

This patch adds an enablement check to `latest` to only apply on tags. Then introduces a new `main` tag for all other builds for any automated testing to look for for un-stable work.

A secondary fix in here, is the security scan now directly tests the tag produced by it's SHA. Previously, a race condition existed and I noticed it in doing this work. I also felt upon seeing it that perhaps the constant builds were tagged as latest in order to make the security scan work. So this ensures the scans continue to work but are improved by not ever being able to test something they aren't intended to.

## Motivation and Context

The `latest` tag has meaning to automatic dependency update systems. It tells them to issue a change against that tag as it changes. Having unstable builds tagged as `latest` introduces a maintenance burden to consumers of the container since they can't rely on the tag to be stable. Meaning the only way to update the container pin is manually.

Fixes: #1370 

## How Has This Been Tested?

For the tag, in the forked repository the [updated tag](https://github.com/Garbee/git-cliff/actions/runs/21863771266/job/63099446223#step:3:51) can be seen. (In this it is the original `nightly` name, run is from before I switched over to `main` for the tag.) There is also no `latest` tag showing in the output.

The security scan can also bee seen attempting to run [against the explicit tag](https://github.com/Garbee/git-cliff/actions/runs/21863771266/job/63099446223#step:10:3). 

> [!IMPORTANT]
> I explicitly did not guard the security scan to not run on forks. I think the workflow needs more reworking to deploy to forked repositories so the scan can happen. Perhaps delete the image after the scan so they don't clutter up the GitHub Package Registry. But, it is good that scans happen in forks so people get early alerts to regressions or problems being introduced. 

## Types of Changes

- [x] Other - Updates deployment of docker tags

## Checklist:

  - [ ] `cargo +nightly fmt --all`
  - [ ] `cargo clippy --tests --verbose -- -D warnings`
  - [x] `cargo test`

### DIFF
--- a/website/docs/docker.md
+++ b/website/docs/docker.md
@@ -31,7 +31,7 @@ Tags are generated from the CI and follow these rules:
 
 - `latest`: only on version tag builds
 - `main`: only on pushes to the `main` branch
-- `sha-<short>`: commit SHA builds
+- `sha-<short>`: commit SHA builds (e.g. `sha-954106f`)
 - `X.Y.Z`: SemVer tag derived from Git tag (e.g. `2.12.0`)
 
 ### Building


Garbee:
That doesn't seem related to my changes unless the extra commit I added to refine things has messed something in it up.

orhun:
We can also update the documentation about these available image tags, e.g.

```patch
diff --git a/website/docs/docker.md b/website/docs/docker.md
index a88b520a..4cc32067 100644
--- a/website/docs/docker.md
+++ b/website/docs/docker.md
@@ -13,7 +13,16 @@ Docker builds are [automated](https://github.com/orhun/git-cliff/tree/main/.gith
 
 ### Usage
 
-The easiest way of running **git-cliff** (in the git root directory with [configuration file](/docs/configuration) present) is to use the available tags from [Docker Hub](https://hub.docker.com/r/orhunp/git-cliff):
+The easiest way of running **git-cliff** (in the git root directory with [configuration file](/docs/configuration) present) is to use the available tags from [Docker Hub](https://hub.docker.com/r/orhunp/git-cliff).
+
+### Tags
+
+Tags are generated from the CI and follow these rules:
+
+- `latest`: only on version tag builds
+- `main`: only on pushes to the `main` branch
+- `sha-<short>`: commit SHA builds
+- `X.Y.Z`: SemVer tag derived from Git tag
 
 ```bash
 docker run -t -v "$(pwd)":/app/ "orhunp/git-cliff:${TAG:-latest}"
```

.github/workflows/docker.yml
```diff
@@ -32,9 +32,23 @@ jobs:
...
+        env:
+          META_JSON: ${{ steps.meta.outputs.json }}
+        run: |
+          tag_value="$(
+            jq -r '
+              first(.tags[] | select(contains(":sha-")))
+              | capture(":(?<tag>[^:]+)$").tag
+            ' <<< "$META_JSON"
+          )"
+          echo "tag=$tag_value" >> "$GITHUB_OUTPUT"
```
orhun: What exactly are we doing here? This jq query seems a bit problematic to maintain/debug if something goes wrong. Isn't there an alternative solution?

Garbee: This jq is to pull the `sha-{hash}` tag from the metadata. That way the later test actually targets the correct image instead of just whatever `latest` is. (This is actually a race condition on `main` in case of slow runners and a currently wasted run in PRs since those images are not deployed anywhere.)

AFAIK, there is no real better way to do this. No matter what, we're parsing data out of some text. Using the JSON output to specifically pull the tags is the simplest and most accurate way.

The only reason this would ever need to be changed going forward is if the `sha` tag is renamed (update the `contains` check). Or if Docker for some reason decides to move where the tags are in the JSON structure (in which case, update the `.tags[]` path in the `first` call.)

I'll add a comment in to help explain what the query is doing. That way should Docker change anything, it should be as simple as feeding the new data structure in with the expected output into AI and it'll generate up a query.

Garbee: So, while I was writing the comment I thought of a better way to do the setup. Now the step just looks for `ghcr.io` and the `:sha-` strings together and pulls that value whole. Then the security scan just uses the full output value instead of reconstructing it. much simpler and should be way easier to maintain if anything does change from docker going forward.

website/docs/docker.md
```diff
@@ -25,6 +25,15 @@ Or you can use the image from the [GitHub Package Registry](https://github.com/o
 docker run -t -v "$(pwd)":/app/ "ghcr.io/orhun/git-cliff/git-cliff:${TAG:-latest}"
 ```
 
+### Tags
+
+Tags are generated from the CI and follow these rules:
+
+- `latest`: only on version tag builds
+- `main`: only on pushes to the `main` branch
+- `sha-<short>`: commit SHA builds
```
orhun:
```suggestion
- `sha-<short>`: commit SHA builds (e.g. `sha-954106f`)
```

website/docs/docker.md
```diff
@@ -25,6 +25,15 @@ Or you can use the image from the [GitHub Package Registry](https://github.com/o
...
+- `X.Y.Z`: SemVer tag derived from Git tag
```
orhun:
```suggestion
- `X.Y.Z`: SemVer tag derived from Git tag (e.g. `2.12.0`)
```

orhun (approved):
thx, looks great now! 💯

PR #1375, opened 2026-02-15
chore(clippy): fix pedantic lint

ognis1205:
## Description

This PR fixes a subset of existing `clippy::pedantic` lints in the codebase.

The changes are intentionally scoped to a limited number of clippy lints to keep the review manageable and to allow reviewers to evaluate each fix incrementally, commit by commit.

- https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#unnecessary_wraps
- https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#ignored_unit_patterns
~~- https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#needless_for_each~~
- https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#explicit_iter_loop
- https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#cast_possible_truncation
~~- https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#redundant_else~~
- https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#must_use_candidate
- https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#struct_excessive_bools
- https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#needless_continue
- https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#ref_option
~~- https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#format_push_string~~
- https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#single_match_else
- https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#non_std_lazy_statics
- https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#needless_raw_string_hashes

See also:

- https://doc.rust-lang.org/stable/cargo/reference/manifest.html#the-lints-section

## Motivation and Context

Applying `clippy::pedantic` on the existing codebase currently produces a large number of warnings, which makes it difficult for contributors to identify actionable issues.

This PR reduces that noise by incrementally fixing selected pedantic lints, so that future contributors can more easily surface meaningful feedback from `clippy::pedantic` during development.

## How Has This Been Tested?

- `cargo clippy --all-targets -- -W clippy::pedantic`
- `cargo test`

N/A

## Types of Changes

- [ ] Other 

## Checklist:

  - [x] `cargo test`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,23 +71,11 @@ env UPDATE_EXPECT=1 cargo test
 
 4. Run CI checks locally - `clippy` (warnings are errors)
 
-
 ```sh
 cargo clippy --tests --verbose -- -D warnings
 ```
 
-5. Run CI checks locally - `clippy` (**optional**, but recommended for pedantic linting)
-
-```sh
-cargo clippy --all-targets --verbose -- -W clippy::pedantic
-```
-
-> [!NOTE]
->
-> - You may allow specific pedantic lints **only with a clear justification**.
-> - Running `clippy` with pedantic lints is **optional**, but it can serve as a helpful guideline for new code and implementations, helping maintain consistency and catch potential issues early.
-
-6. Run formatting checks – `rustfmt`
+5. Run formatting checks – `rustfmt`
 
 ```sh
 cargo +nightly fmt --all -- --check --verbose


git-cliff/src/args.rs
```diff
@@ -469,6 +469,7 @@ impl Opt {
     /// input string into contents of the path returned by [`home_dir`].
     ///
     /// [`home_dir`]: dirs::home_dir
+    #[allow(clippy::unnecessary_wraps)]
```
orhun, line 475: Is clippy complaining because we wrap the return statement in `Ok(())`? Is it fine if we return `Ok(())` in a new line?

git-cliff-core/src/changelog.rs
```diff
@@ -144,22 +144,24 @@ impl<'a> Changelog<'a> {
...
-                        .message
-                        .lines()
-                        .map(|line| { format!("    | {}", line.trim()) })
-                        .collect::<Vec<String>>()
-                        .join("\n")
-                );
-                unmatched_count += 1;
+        for commit in commits.iter() {
+            {
+                let is_unmatched = commit.group.is_none();
```
orhun: I find for_each more readable here (and also we should avoid using scopes like this because they increase the indentation). So should we just ignore this "needless for each" lint?

git-cliff-core/src/remote/azure_devops.rs
```diff
@@ -188,7 +188,7 @@ impl RemoteClient for AzureDevOpsClient {
 impl AzureDevOpsClient {
     /// Constructs the URL for Azure DevOps commits API.
     fn commits_url(api_url: &str, remote: &Remote, ref_name: Option<&str>, page: i32) -> String {
-        let skip = page * MAX_PAGE_SIZE as i32;
+        let skip = page * i32::try_from(MAX_PAGE_SIZE).expect("MAX_PAGE_SIZE should fit in i32");
```
orhun: We can just make `MAX_PAGE_SIZE` `i32`

git-cliff-core/src/commit.rs
```diff
@@ -352,29 +352,27 @@ impl Commit<'_> {
             if parser.sha.clone().map(|v| v.to_lowercase()).as_deref() == Some(&self.id) {
                 if self.skip_commit(parser, protect_breaking) {
                     return Err(AppError::GroupError(String::from("Skipping commit")));
-                } else {
-                    self.group = parser.group.clone().or(self.group);
-                    self.scope = parser.scope.clone().or(self.scope);
-                    self.default_scope = parser.default_scope.clone().or(self.default_scope);
-                    return Ok(self);
                 }
```
orhun, line 364: I prefer explicit else statements tbh

git-cliff-core/src/repo.rs
```diff
@@ -182,12 +182,24 @@ impl Repository {
...
+                    .into_iter()
+                    .map(Self::normalize_pattern)
+                    .collect::<Vec<_>>()
+            });
+            let exclude_patterns = exclude_path.map(|patterns| {
+                patterns
+                    .into_iter()
+                    .map(Self::normalize_pattern)
+                    .collect::<Vec<_>>()
+            });
```
orhun, line 200: Fixing this lint increases the LoC... so maybe let's re-consider it?

git-cliff-core/src/remote/bitbucket.rs
```diff
@@ -160,7 +162,8 @@ impl BitbucketClient {
         );
 
         if let Some(ref_name) = ref_name {
-            url.push_str(&format!("&include={ref_name}"));
+            write!(url, "&include={}", urlencoding::encode(ref_name))
+                .expect("Writing ref name query should never fail");
```
orhun: I'd prefer the other way around tbh, can we ignore this lint?

git-cliff-core/src/commit.rs
```diff
@@ -11,7 +13,9 @@ use crate::error::{Error as AppError, Result};
 
 /// Regular expression for matching SHA1 and a following commit message
 /// separated by a whitespace.
-static SHA1_REGEX: Lazy<Regex> = lazy_regex!(r#"^\b([a-f0-9]{40})\b (.*)$"#);
+//static SHA1_REGEX: Lazy<Regex> = lazy_regex!(r#"^\b([a-f0-9]{40})\b (.*)$"#);
+static SHA1_REGEX: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r#"^\b([a-f0-9]{40})\b (.*)$"#).expect("valid SHA1 regex"));
```
orhun: Did you look into what `lazy_regex` is doing under the hood? Performance wise, doesn't it make more sense to handle the regex compilation at build time? (if that's what `lazy_regex` is doing)

git-cliff/src/args.rs
ognis1205: `clippy` complains because the function always returns `Ok` and never produces an `Err`. However this function is used as a `clap` `value_parser`, which expects a `Fn(&str) -> Result<T, E>`, so the `Result` is required by the API. That's why I kept the allow here.

git-cliff-core/src/commit.rs
ognis1205: I took another look at the implementation of lazy_regex:

- https://github.com/Canop/lazy-regex/blob/e9245929f73083a7f659dcadc485659d78de11c6/src/proc_macros/mod.rs#L81-L84
- https://github.com/Canop/lazy-regex/blob/e9245929f73083a7f659dcadc485659d78de11c6/src/proc_macros/mod.rs#L13-L22
- https://github.com/Canop/lazy-regex/blob/e9245929f73083a7f659dcadc485659d78de11c6/src/proc_macros/regex_code.rs#L21-L23
- https://github.com/Canop/lazy-regex/blob/e9245929f73083a7f659dcadc485659d78de11c6/src/proc_macros/regex_code.rs#L24-L78

At first glance it looks like `Regex::new` is executed at compile time inside the proc macro, but it seems this is only used for syntax validation. The generated code still constructs the regex at runtime.

The expanded code is effectively equivalent to something like:

```rust
static RE: Lazy<Regex> = Lazy::new(|| {
    RegexBuilder::new("pattern")
        .build()
        .unwrap()
});
```

So performance-wise it should be roughly the same as using `std::sync::LazyLock` with `Regex::new()`.

Given that, I personally lean toward using the `std` implementation as well, both to reduce dependencies and from a supply-chain/security perspective.

ognis1205:
I don't have a strong objection to allowing some `clippy::pedantic` lints. However, if we go with that approach, we'll run into the issue we discussed before: some allow rules would need to be passed via CLI options in the GitHub Action.

Personally, if we encourage contributors to run `clippy::pedantic`, I'd prefer not to have those options scattered across different places.

Given that, I think we have a few options:

1. Keep the CI configuration as-is and pass the allow options only in CI, but not explicitly recommend contributors run `clippy::pedantic` locally (maintainers can check it when needed).
2. Provide a shortcut (e.g., via justfile or makefile) that runs something like `cargo clippy --all-targets --verbose -- -W clippy::pedantic` with the necessary allow flags.
3. Remove the workflow that runs `cargo clippy --all-targets --verbose -- -W clippy::pedantic`.

Any thoughts on this @orhun ?

git-cliff-core/src/commit.rs
orhun: Cool, thanks for the insight!

orhun:
I think running the pedantic lints only in the CI (as a part of the repo) and not recommending it to the contributors make the most sense here.

ognis1205:
Regarding the clippy lints you mentioned in your comments, I've made the following changes:

- https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#needless_for_each
- https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#redundant_else
- https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#format_push_string

These lints are now disabled in ci, and I also removed the "pedantic lints" section from `CONTRIBUTING.md`. When you have a chance, could you take a look and review again?

Thanks!

orhun:
thanks!

PR #1379, opened 2026-02-22
refactor(args): centralize include-path resolution after repo discovery

### LINKED ISSUE #1369: Setting `--workdir` results in an empty output

### Description of the bug

When setting `--workdir <path>` to the path of the repository or setting `--workdir <parent-dir>` to the parent of the repository: The `git-cliff` output is empty.

This used to work in `2.8.0`. Maybe the meaning of `--workdir` changed related to some of the changes in the last few versions regarding `--include-path`, path resolution etc.

### Steps To Reproduce

1. Setup the repository:
    ```bash
    mkdir git-cliff-workdir
   cd git-cliff-workdir/
   git init
   git commit -m "feat: Changelog entry" --allow-empty
    ```
2. These commands work as expected:
     ```bash
     # from inside git-cliff-workdir/
     # works
     git-cliff

     cd ..
     # works
     git-cliff --repository git-cliff-workdir
     ```
3. These however, output an empty changelog:
   ```bash
   # from inside git-cliff-workdir/
   git-cliff --workdir .
   cd ..
   # from the parent directory of git-cliff-workdir/
   git-cliff --workdir git-cliff-workdir/
   git-cliff --repository git-cliff-workdir --workdir ./
   ```

### Expected behavior

`git-cliff` should output the same changelog as for the other two commands, so for the example above:
```
## [unreleased]

### 🚀 Features

- Changelog entry
```

### Screenshots / Logs

```bash
> ./git-cliff-workdir-repro.sh
+ mkdir git-cliff-workdir
+ cd git-cliff-workdir/
+ git init
...
+ git commit -m 'feat: Changelog entry' --allow-empty
[master (root-commit) 927f9d3] feat: Changelog entry
+ git-cliff
 WARN  git_cliff > "cliff.toml" is not found, using the default configuration
## [unreleased]

### 🚀 Features

- Changelog entry
+ git-cliff --workdir .
 WARN  git_cliff > "./cliff.toml" is not found, using the default configuration
+ cd ..
+ git-cliff --repository git-cliff-workdir
 WARN  git_cliff > "cliff.toml" is not found, using the default configuration
## [unreleased]

### 🚀 Features

- Changelog entry
+ git-cliff --workdir git-cliff-workdir/
 WARN  git_cliff > "git-cliff-workdir/cliff.toml" is not found, using the default configuration
+ git-cliff --repository git-cliff-workdir --workdir ./
 WARN  git_cliff > "./cliff.toml" is not found, using the default configuration
+ rm -rf git-cliff-workdir/
```

### Software information

- Project version: `git-cliff 2.12.0`

### Additional context

Might be related to:

- 2.11.0
  - (args) Set the include-path if workdir is set (#1293) - (50b8312)
- 2.10.0
  - (config) Support using include and exclude paths in the config (#1173) - (7c2f922)

After further investigation:
- Since setting `--workdir` now also sets the include path, empty commits without any file changes will be ignored.
- Absolute paths passed as `--workdir` are not handled properly. The include patterns themselves include the absolute path, while the path of the files itself are compared using the relative path in `Repository::should_retain_commit`.

o1x3:
## Description

Extracts path resolution for `--workdir` and `--include-path` into a single `resolve_include_paths` function that runs after repo discovery, where the repo root is available.

Previously, the workdir block set `args.include_path` to a raw workdir path before repo discovery (absolute or cwd-relative), which never matched the repo-relative paths from git2's diff API. This broke `--workdir` in v2.11.0.

The new function handles three cases:
- Config/CLI include paths are normalized to repo-relative (strips absolute repo root prefix)
- `--workdir` set to a subdirectory adds a repo-relative glob; workdir at root means no filter
- Implicit cwd scoping when running from a subdirectory (existing behavior, preserved)

All paths (root, cwd, workdir) are canonicalized before comparison to handle symlinks.

## Motivation and Context

Fixes #1369

As discussed in the issue, the path resolution logic was scattered across three locations in `lib.rs`. @orhun asked for a proper refactor rather than a band-aid fix.

## How Has This Been Tested?

- 9 unit tests for `resolve_include_paths` covering: workdir at root, workdir subdirectory, workdir outside repo, implicit cwd scoping, no scoping at root, repository arg suppressing implicit scoping, config patterns combined with workdir
- 2 unit tests for `normalize_to_repo_relative`
- New fixture test `test-workdir-root` (`--workdir .` with empty + file-changing commits)
- Manual verification of all scenarios from the issue repro script

## Types of Changes

- [ ] Other

## Checklist:

  - [x] `cargo test`

### DIFF
--- a/git-cliff/src/lib.rs
+++ b/git-cliff/src/lib.rs
@@ -13,7 +13,7 @@ pub mod logger;
 use std::env;
 use std::fs::{self, File};
 use std::io::{self, Write};
-use std::path::{Path, PathBuf};
+use std::path::{Component, Path, PathBuf};
 use std::time::{SystemTime, UNIX_EPOCH};
 
 use args::{BumpOption, Opt, Sort, Strip};
@@ -183,6 +183,52 @@ pub fn init_config(name: Option<&str>, config_path: &Path) -> Result<()> {
     Ok(())
 }
 
+fn normalize_path(path: &Path) -> PathBuf {
+    let mut normalized = PathBuf::new();
+
+    for component in path.components() {
+        match component {
+            Component::Prefix(prefix) => normalized.push(prefix.as_os_str()),
+            Component::RootDir => normalized.push(component.as_os_str()),
+            Component::CurDir => {}
+            Component::ParentDir => {
+                if normalized.as_os_str().is_empty() || !normalized.pop() {
+                    normalized.push(component.as_os_str());
+                }
+            }
+            Component::Normal(part) => normalized.push(part),
+        }
+    }
+
+    normalized
+}
+
+fn resolve_workdir_include_path(
+    root: &Path,
+    cwd: &Path,
+    workdir: &Path,
+) -> Result<Option<Pattern>> {
+    let resolved_workdir = if workdir.is_absolute() {
+        normalize_path(workdir)
+    } else {
+        normalize_path(&cwd.join(workdir))
+    };
+
+    if !resolved_workdir.starts_with(root) || resolved_workdir == root {
+        return Ok(None);
+    }
+
+    let path = resolved_workdir.join("**").join("*");
+    if let Ok(stripped) = path.strip_prefix(root) {
+        log::info!(
+            "Including changes from the workdir: {}",
+            resolved_workdir.display()
+        );
+        return Ok(Some(Pattern::new(stripped.to_string_lossy().as_ref())?));
+    }
+
+    Ok(None)
+}
 /// Processes the tags and commits for creating release entries for the
 /// changelog.
 ///
@@ -291,6 +337,11 @@ fn process_repository<'a>(
     let cwd = env::current_dir()?;
     let mut include_path = config.git.include_paths.clone();
     if let Ok(root) = repository.root_path() {
+        if let Some(workdir) = args.workdir.as_ref() {
+            if let Some(pattern) = resolve_workdir_include_path(&root, &cwd, workdir)? {
+                include_path.push(pattern);
+            }
+        }
         if cwd.starts_with(&root) &&
             cwd != root &&
             args.repository.as_ref().is_none_or(Vec::is_empty) &&
@@ -867,4 +918,53 @@ pub fn write_changelog<W: io::Write>(
     }
 
     Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn as_str(pattern: &Option<Pattern>) -> Option<&str> {
+        pattern.as_ref().map(Pattern::as_str)
+    }
+
+    #[test]
+    fn workdir_at_root_returns_none() {
+        let root = Path::new("/repo");
+        let cwd = Path::new("/repo");
+        let result = resolve_workdir_include_path(root, cwd, Path::new(".")).unwrap();
+        assert!(result.is_none());
+    }
+
+    #[test]
+    fn workdir_subdirectory_scopes_to_subdir() {
+        let root = Path::new("/repo");
+        let cwd = Path::new("/repo");
+        let result = resolve_workdir_include_path(root, cwd, Path::new("sub")).unwrap();
+        assert_eq!(as_str(&result), Some("sub/**/*"));
+    }
+
+    #[test]
+    fn relative_workdir_from_parent_repo_root_returns_none() {
+        let root = Path::new("/parent/repo");
+        let cwd = Path::new("/parent");
+        let result = resolve_workdir_include_path(root, cwd, Path::new("./")).unwrap();
+        assert!(result.is_none());
+    }
+
+    #[test]
+    fn relative_workdir_from_parent_subdirectory_scopes_to_subdir() {
+        let root = Path::new("/parent/repo");
+        let cwd = Path::new("/parent");
+        let result = resolve_workdir_include_path(root, cwd, Path::new("repo/sub")).unwrap();
+        assert_eq!(as_str(&result), Some("sub/**/*"));
+    }
+
+    #[test]
+    fn absolute_workdir_outside_repo_returns_none() {
+        let root = Path::new("/repo");
+        let workdir = Path::new("/other/path");
+        let result = resolve_workdir_include_path(root, Path::new("/tmp"), workdir).unwrap();
+        assert!(result.is_none());
+    }
 }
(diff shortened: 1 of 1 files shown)


o1x3:
hey @orhun @ognis1205 - draft is up for the refactor we discussed in #1369. main change is pulling the include-path resolution into one function that runs after repo discovery instead of the three scattered spots. still a draft so happy to rework anything based on feedback.

git-cliff/src/lib.rs
```diff
@@ -273,40 +327,24 @@ fn process_repository<'a>(
...
-                );
-                include_path = vec![Pattern::new(stripped.to_string_lossy().as_ref())?];
-            }
-        }
-    }
-
-    let include_path = (!include_path.is_empty()).then_some(include_path);
+    let include_path = if let Ok(root) = repository.root_path() {
+        let canonical_root = fs::canonicalize(&root).unwrap_or(root);
+        let canonical_cwd = fs::canonicalize(&cwd).unwrap_or(cwd);
```
ognis1205: Regarding the use of `unwrap` on `fs::canonicalize`:

According to the documentation:

https://doc.rust-lang.org/std/fs/fn.canonicalize.html#errors

`fs::canonicalize` can fail in several common cases, such as when the path does not exist or a non-final component is not a directory. Propagating the error instead of using `unwrap` would allow us to inform the user about what went wrong.

git-cliff/src/lib.rs
```diff
@@ -273,40 +327,24 @@ fn process_repository<'a>(
...
+        let canonical_workdir = args
+            .workdir
+            .as_ref()
+            .map(|w| fs::canonicalize(w).unwrap_or(w.clone()));
```
ognis1205: Same as https://github.com/orhun/git-cliff/pull/1379/changes#r2876700091 .

git-cliff/src/lib.rs
```diff
@@ -181,6 +181,60 @@ pub fn init_config(name: Option<&str>, config_path: &Path) -> Result<()> {
...
+}
+
+/// Strips the repo root prefix from absolute patterns.
+fn normalize_to_repo_relative(canonical_root: &Path, pattern: &Pattern) -> Result<Pattern> {
+    let pattern_path = Path::new(pattern.as_str());
+    if pattern_path.is_absolute() {
+        if let Ok(stripped) = pattern_path.strip_prefix(canonical_root) {
+            return Ok(Pattern::new(stripped.to_string_lossy().as_ref())?);
+        }
+    }
```
ognis1205: Is this a new feature? I believe this absolute path handling should not be part of this PR's scope. The previous implementation didn't have special handling for absolute paths.
If we want to add this behavior, it should be done in a separate PR with proper tests and documentation.

git-cliff/src/lib.rs
```diff
@@ -181,6 +181,60 @@ pub fn init_config(name: Option<&str>, config_path: &Path) -> Result<()> {
...
+
+    if let Some(workdir) = canonical_workdir {
+        if workdir != canonical_root {
+            if let Ok(rel) = workdir.strip_prefix(canonical_root) {
+                let glob = rel.join("**").join("*");
+                log::info!("Including changes from workdir: {}", workdir.display());
+                patterns.push(Pattern::new(glob.to_string_lossy().as_ref())?);
+            }
+        }
+    } else if patterns.is_empty() &&
```
ognis1205: I'm not sure if this `patterns.is_empty()` condition is necessary here.

ognis1205 (changes requested):
Thanks for the effort!

I'm quite cautious about this PR, especially around the directory handling changes.

Historically, this area has been fragile: fixing one bug has often led to unexpected behavior elsewhere, and we've gone back and forth with multiple fixes. Because of that history, I'm very hesitant to introduce substantial changes now.

On top of that, I believe the `git-cliff` crate as a whole needs a broader refactoring. Although this PR is labeled as a refactoring, I feel it also contains unclear functional changes, and I'm not convinced this is the right time for such a refactoring.

In particular, simply consolidating directory handling into a single function doesn't address the root issues. Given the past instability, I'm concerned we'll reintroduce bugs that we haven't fully anticipated.

Personally, I'd prefer to avoid major changes to the existing implementation for now, and instead fix the `workdir` issue based on the current code. I think that would let us isolate the problem more cleanly and reduce the risk of regressions.

cc: @orhun  any thoughts on this?

ognis1205:
Thanks again for working on this.

As mentioned in my review, I'm still a bit hesitant about the scope of the changes here. Given that this area has historically been quite fragile, I'd prefer to keep the existing implementation mostly intact for now and focus only on fixing the `--workdir` issue.

Would you be open to adjusting this PR to target the `--workdir` bug while preserving the current structure as much as possible?

o1x3:
@ognis1205 @orhun sorry for the delay on my side. I had to take some time off for personal reasons, but I am back now and still available to take this forward.

I went back through the review feedback and reworked this into a smaller fix for the `--workdir` bug locally. I dropped the broader include path refactor and the extra path normalization changes from this PR.

What changed in the local rework:
- stopped setting `include_path` early from `--workdir`
- kept the current `cwd` scoping logic in place
- only add a workdir include pattern later, after repo discovery, when the workdir is inside the repo and not the repo root
- kept focused tests for the workdir cases
- kept the root `--workdir .` fixture test

I also ran:
- `cargo test --package git-cliff`
- `cargo clippy --tests --verbose --package git-cliff -- -D warnings`

I still need to push the narrowed version, but the broader refactor pieces are out now. Please take another look when I push the update.

Also happy to collaborate if that would make this easier to move forward.

git-cliff/src/lib.rs
o1x3: I removed the canonicalize based approach from this PR while narrowing the scope.

git-cliff/src/lib.rs
o1x3: I removed the canonicalize based approach from this PR while narrowing the scope.

git-cliff/src/lib.rs
o1x3: Removed from this PR. I agree that this was extra behavior and not needed for the `--workdir` fix itself.

git-cliff/src/lib.rs
o1x3: Removed with the larger refactor. The current version keeps the existing structure and only changes the `--workdir` handling.

orhun:
Hey again, I kinda lost track on this. What is the latest state and do you wish to proceed with a review?

PR #1443, opened 2026-03-12
Fix: avoid random testimonial during hydration

### LINKED ISSUE #1385: Hydration mismatch warning in Testimonials component when using SSG

### Description of the bug

When the website is built using SSG, the `Testimonials` component selects a random testimonial using `Math.random()` during render.

Because this logic runs both:

- at build time (server-side rendering)
- and again during client-side hydration

the rendered content may differ between server and client. This results in a React hydration mismatch warning in the browser console.

### Steps To Reproduce

1. `yarn build`
2. `yarn serve`
3. Open the homepage in a browser
4. Open the browser developer console
5. Observe a hydration mismatch warning

### Expected behavior

The website should hydrate without React warnings.

### Screenshots / Logs

Image

### Software information

- Operating system: macOS Tahoe 26.3
- Rust version: 1.91.0
- Project version: 2.12.0

### Additional context

This issue occurs because `Math.random(`) is executed during render. In an SSG environment, this leads to non-deterministic output between server and client.

omar-y-abdi:
## Problem

Issue #1385 reproduces on the homepage after building and serving the website. `Testimonials` can render one testimonial in the static HTML and a different one during hydration, which triggers a React hydration mismatch warning.

## Cause

`Testimonials` calls `Math.random()` during render. SSG and the first client render therefore choose independently, so the server markup and hydrated client tree can diverge.

## Fix

The component now renders a deterministic testimonial for the server render and initial client render. After mount, a `useEffect` selects a random testimonial on the client, preserving the randomized behavior without mismatching hydration.

## Testing

`./node_modules/.bin/tsc`
`./node_modules/.bin/docusaurus build`
`cargo test`
`cargo clippy --tests --verbose -- -D warnings`
Headless Chrome against the built site showed no hydration-related console warnings after the change.

Closes #1385

### DIFF
--- a/website/src/components/Testimonials/index.tsx
+++ b/website/src/components/Testimonials/index.tsx
@@ -57,7 +57,11 @@ const getRandomTestimonial = () => {
 };
 
 export default function Testimonials(): JSX.Element {
-  const review = getRandomTestimonial();
+  const [review, setReview] = React.useState(testimonials[0]);
+
+  React.useEffect(() => {
+    setReview(getRandomTestimonial());
+  }, []);
 
   return (
     <div className={styles.testimonialsList}>


orhun (approved):
haven't tested but overall LGTM

website/src/components/Testimonials/index.tsx
```diff
@@ -57,7 +57,11 @@ const getRandomTestimonial = () => {
 };
 
 export default function Testimonials(): JSX.Element {
-  const review = getRandomTestimonial();
+  const [review, setReview] = React.useState(testimonials[0]);
```
ognis1205, line 60: I'd suggest using `null` as the initial state instead.

With this implementation, the first testimonial (`testimonials[0]`) will briefly flash on the screen before the `useEffect` swaps it for a random one (`componentDidMount`).

Starting with `null` provides a cleaner transition and avoids showing placeholder data for a split second. While the logic after `componentDidMount` is lightweight and the difference might be negligible in most environments, using `null` is a bit more robust for consistent UX. We could technically add a skeleton loader for the null state, but that feels like overkill here, simply starting with a clean null should be enough.

What do you think?

ognis1205 (commented):
Thanks for the fix! Overall looks good.

website/src/components/Testimonials/index.tsx
omar-y-abdi:
```diff
- const getRandomTestimonial = () => {
-     const randomIndex = Math.floor(Math.random() * testimonials.length);
-     return testimonials[randomIndex];
- };
-
- export default function Testimonials(): JSX.Element {
-
-   const review = getRandomTestimonial();
-
-   return (
+ export default function Testimonials(): JSX.Element {
+
+   const review = testimonials[0];
+
+   return (
```
Appreciate the thought, but I'd go further, drop `useState`/`useEffect` entirely and just use `const review = testimonials[0]`. SSR and hydration match exactly, no flash, no empty state, no pop-in. Your `null` approach fixes the mismatch too but trades the flash for a briefly empty hero, which is worse UX IMO. 

Tested all three variants locally and the deterministic one was cleanest. The only one where SSR-HTML and hydrated DOM matched exactly with zero visual artifacts.

Whenever you got time test it yourself, tell me what you think

ognis1205: Thanks for the feedback.

> Your null approach fixes the mismatch too but trades the flash for a briefly empty hero, which is worse UX IMO.

It's quite common to use a skeleton and render an empty placeholder on the first pass. In that sense, rendering an empty hero initially is a fairly standard pattern to avoid hydration issues. Because of that, I don't think the UX impact here would necessarily be problematic.

> Appreciate the thought, but I'd go further, drop useState/useEffect entirely and just use const review = testimonials[0].
> Tested all three variants locally and the deterministic one was cleanest. The only one where SSR-HTML and hydrated DOM matched exactly with zero visual artifacts.

That would remove the randomness entirely, so the deterministic approach probably isn't really an option here.

omar-y-abdi: Fair point on keeping the randomness! 
I still kinda think `useState(testimonials[0])` with `useEffect` gives the best of both worlds, users get to see real content right away while still getting a random testimonial on mount. What do you think?

ognis1205: I see your point about showing content immediately for better UX. However, I have a few concerns regarding that approach:

- From a programming semantics perspective, I don't think it's ideal to display `testimonials[0]` as a "placeholder" if it's not the final intended content. Showing data that will be replaced almost instantly feels like a "false start" rather than meaningful content.
- While it might happen fast, some users on slower devices may notice the content "flickering" or swapping.

In my opinion, keeping it `null` (or using a proper loading skeleton) is more honest to the component's state. It ensures we only ever show the "correct" random testimonial once it's determined.

omar-y-abdi: I truly didn't think that far, and with your raised points and I concede, I retract previous statements gracefully hahaha. 
You're right in your thinking, I just got a little too focused on the UX looking good that I never broadened my thinking beyond it, damn! 
You saying:

> Showing data that will be replaced almost instantly feels like a "false start" rather than meaningful content.

and

> on slower devices may notice the content "flickering" or swapping

made me understand your thought process and I'm with you, these two will play a huge role in user experience. 

I hear you fr!

ognis1205 (changes requested):

omar-y-abdi:
@orhun Hey, this is ready to merge whenever you get a chance.

PR #1448, opened 2026-03-18
feat: support more configuration files

danielpza:
## Motivation and Context

Fixes https://github.com/orhun/git-cliff/issues/1446

## Types of Changes

- [ ] Other 

## Checklist:

  - [x] `cargo test`


orhun:
Thanks for the PR! Just wanted to say this goes into the right direction. Let me know when this is ready for review

danielpza:
> Thanks for the PR! Just wanted to say this goes into the right direction. Let me know when this is ready for review

Hey! I think the current code should work, but I want to explore a few changes first to make it a bit more robust.

I'll be away for a couple of weeks though. I'll let you all know when it's ready. Or you can give it a try yourself if you are inclined.

Cheers.

danielpza:
Hey @orhun, it should be good for review now. I added a couple of tests and updated the documentation to cover the changes.

I did notice there are some inconsistencies in how the configuration files are loaded, but I decided to leave that out of the scope of this PR. Eg:

- The current configuration file (`cliff.toml`) is read from `workdir` (if set), but the same doesn't apply when searching the configuration in the parent directories. For the parent directories it always searches up from [`env::current_dir()`](https://github.com/danielpza/git-cliff/blob/aac29128d051b319ae2dc4a2ab9acd0f0a5fb814/git-cliff/src/lib.rs#L572)
- The order it loads the configuration files it's not intuitive imo:
  - it searches for the `cliff.toml` file (or args.config) in current directory or `workdir` if set (**project configuration file**)
  - searches for user configuration file in home directory (**user configuration file**)
  - searches for `cliff.toml` in parent directories. (**project configuration file**)

  I would expect it to search for **project configuration files** (including current and parent directories) first and then to search for the **user configuration file**

[Related code](https://github.com/danielpza/git-cliff/blob/aac29128d051b319ae2dc4a2ab9acd0f0a5fb814/git-cliff/src/lib.rs#L554C5-L590C7)

Let me know your thoughts.

git-cliff-core/src/config.rs
```diff
@@ -610,4 +619,33 @@ mod test {
         assert!(!Remote::new("test", "").is_set());
         assert!(!Remote::new("", "").is_set());
     }
+
+    #[test]
+    #[allow(clippy::unwrap_used)]
```
orhun: Please avoid allowing this lint and using `unwrap`. You can return `Result<()>` from tests instead.

git-cliff-core/src/config.rs
```diff
@@ -610,4 +619,33 @@ mod test {
...
+        let dir = tempdir().unwrap();
+        assert_eq!(Config::retrieve_project_config_path(dir.path()), None);
+    }
+
+    #[test]
+    #[allow(clippy::unwrap_used)]
+    fn find_project_config_file_returns_first_match_in_priority_order() {
+        // check config files in order of priority. config.toml has the highest priority to preserve
+        // backward compatibility cliff.toml > .cliff.toml > ... > .config/cliff.toml
+
```
orhun:
```suggestion
    /// Check config files in order of priority.
    /// cliff.toml has the highest priority to preserve
    /// Backward compatibility cliff.toml > .cliff.toml > ... > .config/cliff.toml
    fn find_project_config_file_returns_first_match_in_priority_order() {

```

website/docs/configuration/index.md
```diff
@@ -1,10 +1,25 @@
...
 sidebar_position: 4
 ---
+
 # Configuration
 
 **git-cliff** configuration file supports [TOML](https://github.com/toml-lang/toml) (preferred) and [YAML](https://yaml.org) formats.
 
+**git-cliff** will look for a configuration file first in the project directory, then in the global user directory. If no configuration file is found, **git-cliff** will use the default configuration values.
+
+## Project Configuration
```
orhun, line 11: I'm not sure if separating this section as Project and User Configuration makes sense here. What was the reasoning behind it?

danielpza: I wanted to avoid confusion since the naming convention for project and global user configurations are different.

Since the global user configuration is already scoped to `git-cliff` namespace, it doesn't make sense for additional scoping.

- :heavy_check_mark:  `/home/<user>/.config/git-cliff/cliff.toml`
- :x:  `/home/<user>/.config/git-cliff/.cliff.toml` unnecessary scoping, other programs (git, mise, etc) don't follow this convention
- :x:  `/home/<user>/.config/git-cliff/.config/cliff.toml` unnecessary scoping, other programs (git, mise, etc) don't follow this convention

I also used https://mise.jdx.dev/configuration.html#global-config-config-mise-config-toml as reference, they have a more detailed section for all the configuration files possible locations and patterns, and how they differ from each other.

git-cliff-core/Cargo.toml
```diff
@@ -83,6 +83,7 @@ urlencoding = "2.1.3"
 cacache = { version = "13.1.0", features = ["mmap", "tokio-runtime"], default-features = false }
 time = "0.3.44"
 chrono = { version = "0.4.41", features = ["serde"] }
+tempfile = "3.27.0"
```
ognis1205: Is this dependency only used in tests, or do we also need it in the main code?

git-cliff-core/src/config.rs
```diff
@@ -610,4 +619,34 @@ mod test {
...
+    fn find_project_config_file_returns_none_when_no_config_exists() -> Result<()> {
+        let dir = tempdir()?;
+        assert_eq!(Config::retrieve_project_config_path(dir.path()), None);
+        Ok(())
+    }
+
+    #[test]
+    /// Check config files in order of priority.
+    /// cliff.toml has the highest priority to preserve
+    /// Backward compatibility cliff.toml > .cliff.toml > ... > .config/cliff.toml
```
ognis1205: I think the doc comment `///` is unnecessary here since this is a test. The function name already explains the intent well, so we might not need an additional comment here.

This comment seems to describe the behavior of `retrieve_project_config_path` rather than the test itself. It might be better to move it to the function as a doc comment.

git-cliff-core/src/config.rs
```diff
@@ -528,6 +528,14 @@ impl Config {
         }
         None
     }
+
+    /// Returns the first valid configuration file found in `dir`.
```
ognis1205, line 532: https://github.com/orhun/git-cliff/pull/1448/changes#r3043456251

git-cliff-core/src/config.rs
```diff
@@ -610,4 +619,34 @@ mod test {
...
+        );
+
+        fs::write(dir.path().join("cliff.toml"), "")?;
+        assert_eq!(
+            Config::retrieve_project_config_path(dir.path()),
+            Some(dir.path().join("cliff.toml")),
+        );
+
+        Ok(())
+    }
```
ognis1205, line 646: Since these tests are fairly small and closely related, it might be reasonable to combine them into a single test. Also, following the existing naming convention, something like `find_project_config_file` could be a better fit.

ognis1205 (changes requested):
Thanks for the PR!

This is slightly tangential to the changes in this PR, but I wanted to share a broader design thought.

From my perspective, `git-cliff` is fundamentally a project-level tool, and its configuration should ideally be tied to the repository to ensure reproducibility. In that sense:

- The presence of sections like `remote` in `cliff.toml` suggests that the config is repository-specific
- We already support generating a project-local config via `git cliff --init`
- Keeping configuration within the repository helps maintain consistency across environments

Given that, supporting xdg / os-specific config directories might introduce additional maintenance overhead while providing relatively limited benefit, and could blur the intended design direction.

Personally, I would even consider deprecating user-level `cliff.toml` at some point, if it aligns with the overall direction of the project.

Curious to hear what others think. cc: @orhun

git-cliff-core/src/config.rs
danielpza: I don't have a strong opinion either way. @orhun suggested moving it there https://github.com/orhun/git-cliff/pull/1448#discussion_r3041231065. What's the consensus/best practice here?

git-cliff-core/Cargo.toml
danielpza: only in tests, moved it to the dev dependencies section https://github.com/orhun/git-cliff/pull/1448/commits/1f8dc2b0e60f4ce25f7cfe0237cdc0f59158d806

git-cliff-core/src/config.rs
danielpza: Addressed in https://github.com/orhun/git-cliff/pull/1448/commits/b53b9dfd1727fe15f587d173cda4ee3e702ffe2f

git-cliff-core/src/config.rs
danielpza: https://github.com/orhun/git-cliff/pull/1448/commits/b53b9dfd1727fe15f587d173cda4ee3e702ffe2f

git-cliff-core/src/config.rs
danielpza: updated in https://github.com/orhun/git-cliff/pull/1448/commits/b53b9dfd1727fe15f587d173cda4ee3e702ffe2f

website/docs/configuration/index.md
Cyclonit: I think it would be cleaner to just list the actual paths it will search through instead of filenames and trying to explain relative path situations. E.g.

- `./cliff.toml`
- `./.cliff.toml`
- `./.config/.cliff.toml`
- `~/cliff.toml`
- `~/.cliff.toml`
- `~/.config/.cliff.toml`

A bit more verbose, but easier to understand than trying to deduplicate parts of the paths.

orhun (approved):
LGTM, thanks!

PR #1456, opened 2026-03-31
test(fixtures): add include-path multi-value handling tests

niklasmarderx:
Add two fixture tests that verify `--include-path` handles multiple glob patterns correctly (#1455):

- **test-include-path-repeated-flag**: `--include-path 'website/**/*' --include-path 'docs/**/*'`
- **test-include-path-space-separated**: `--include-path 'website/**/* docs/**/*'`

Both use the same commit history (files in `website/`, `src/`, and `docs/`) and expect the same filtered output — only commits touching `website/` or `docs/` paths.

Closes #1455


orhun:
Hey, thanks for the PR!

> The workflow entries for test-fixtures.yml need to be added separately since fork pushes can't modify workflow files. Here's what to add to the matrix:

Is that a new limitation that GitHub has? I think you should be able to change those files in your fork and push them to this PR, no?

niklasmarderx:
You're right, I was mistaken — just pushed the workflow entries in e03d189. Should be good now.

ognis1205 (changes requested):
A couple of points regarding the fixture tests:

- Newly added fixture tests should all pass in CI.
- The following case seems to be missing, could you add it?

> 1. Environment variable with space-separated globs:
> ```sh
> export GIT_CLIFF_INCLUDE_PATH="website/**/* .github/**/*"
> git cliff
> ```

- Ideally, each fixture test should correspond to a single expected output. However, given the current CI setup, this is not straightforward. As a workaround, please add a comment in `workflow/test-fixtures.yml` indicating that the three fixture test entries should use identical `expected.md` content.

niklasmarderx:
I'll fix it

niklasmarderx:
Added the env-var fixture in 63ee5ff. Since the run-fixtures-test action controls the git-cliff invocation, I extended it with an include-path-env input that sets GIT_CLIFF_INCLUDE_PATH before running git-cliff. The new test-include-path-env-var fixture uses the same commit history and expected output as the other three include-path tests.

niklasmarderx:
Thanks for the review, @ognis1205!

I've addressed both points in commit 04bad02:

1. **New fixture test with `website/**/* .github/**/*`**: Added `test-include-path-env-var-github` which sets `GIT_CLIFF_INCLUDE_PATH="website/**/* .github/**/*"`. Since `.github/**/*` doesn't match any commits in the fixture history, only `website/**/*` commits appear in the output — this verifies that space-separated globs work correctly even when one glob matches nothing.

2. **Workflow comment**: The `test-fixtures.yml` already contains a `# NOTE` comment clarifying that the four existing include-path tests share identical `expected.md` content. I also added a separate comment above the new `test-include-path-env-var-github` entry explaining its distinct expected output.

.github/actions/run-fixtures-test/action.yml
```diff
@@ -46,6 +50,9 @@ runs:
       shell: bash
       run: |
         set +e
+        if [ -n '${{ inputs.include-path-env }}' ]; then
+          export GIT_CLIFF_INCLUDE_PATH='${{ inputs.include-path-env }}'
+        fi
```
ognis1205, line 55: We could introduce an optional `env.sh` per fixture and load it in a dedicated initialization step like this:

```yaml
- name: Loading environmental variables
  working-directory: ${{ inputs.fixtures-dir }}
  shell: bash
  run: |
    if [ -f "./env.sh" ]; then
      set -a
      source ./env.sh
      set +a
      env | grep -E '^(GIT_CLIFF_)' >> "$GITHUB_ENV"
    fi
```

This makes it explicit that fixture-specific environment variables are defined and then propagated across steps via `GITHUB_ENV`, rather than being passed through action inputs. It would also allow us to remove inputs like `include-path-env` and keep each fixture self-contained.

We should still ensure that only a limited set of variables, e.g. `GIT_CLIFF_*`, are propagated to avoid leaking unintended environment variables. cc: @orhun

.github/workflows/test-fixtures.yml
```diff
@@ -143,6 +143,24 @@ jobs:
...
+            command: --include-path 'website/**/* docs/**/*'
+          - fixtures-name: test-include-path-config
+          - fixtures-name: test-include-path-env-var
+            include-path-env: "website/**/* docs/**/*"
+          # NOTE: This test verifies that GIT_CLIFF_INCLUDE_PATH correctly
+          # handles space-separated globs where one glob matches no commits
+          # (i.e. .github/**/* does not exist in the fixture's git history).
+          # Only website/**/* commits are expected in the output.
+          - fixtures-name: test-include-path-env-var-github
+            include-path-env: "website/**/* .github/**/*"
```
ognis1205, line 163: We could consider removing this test, as this behavior should already be covered by unit tests.

ognis1205 (changes requested):
Overall looks great!

- It seems the CI for the newly added fixture tests is currently failing, so could you take a look?
- We may also want to add similar tests for `--exclude-path`, but that could be done in a separate PR.

Thanks!

niklasmarderx:
Fixed the trailing newline in `test-include-path-env-var-github/expected.md` — it was missing the extra `\n` that all other fixtures have, which caused the diff to fail.

niklasmarderx:
CI is green now. The trailing newline fix in `test-include-path-env-var-github/expected.md` resolved the last failure — all fixture tests pass.

orhun (approved):
LGTM, thanks!

Issue #1080, opened 2025-03-06
Version tags don't get sorted topologically?


dpad:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description of the bug

Sorry if this is just a mistake, I'm not sure if I'm just misunderstanding things or if there is some weird behaviour with the "topo_order" option...

Basically, I have 3 commits tagged with version numbers like "v0.1.0", "v0.9.0", "v0.10.0" respectively. Topologically, the latest version (on the HEAD commit) is v0.10.0. However, running `git-cliff --topo-order --latest` will print out the "v0.9.0" changelog, and `git-cliff --topo-order --unreleased` will print out the "v0.10.0" changelog. Removing the `--topo-order` argument produces the expected result.

### Steps To Reproduce

1. git init
2. git commit --allow-empty -m "feat: new"
3. git commit --allow-empty -m "fix: whatever"
4. git commit --allow-empty -m "doc: stuff"
5. git tag v0.1.0 HEAD~2
6. git tag v0.9.0 HEAD~
7. git tag v0.10.0

```git
$ git log --pretty=oneline
871adbf82b0dba2a4719fff11793387934a3cc11 (HEAD -> master, tag: v0.10.0) doc: stuff
b2c33a4df36176363f8a77421c917f212bfdcf98 (tag: v0.9.0) fix: whatever
457b3c4e0fadb9c8801e44d16d860054d9331609 (tag: v0.1.0) feat: new
```

```
$ git tag -l
v0.1.0
v0.10.0
v0.9.0
```

```
$ git cliff --topo-order --latest 
 WARN  git_cliff > "cliff.toml" is not found, using the default configuration.
# Changelog

All notable changes to this project will be documented in this file.

<!-- generated by git-cliff -->
```

```
$ git cliff --topo-order --unreleased
 WARN  git_cliff > "cliff.toml" is not found, using the default configuration.
# Changelog

All notable changes to this project will be documented in this file.

## [0.10.0] - 2025-03-06

### 📚 Documentation

- Stuff

<!-- generated by git-cliff -->
```

Generating the full changelog seems to have the tags in the correct order:

```
$ git cliff --topo-order
 WARN  git_cliff > "cliff.toml" is not found, using the default configuration.
# Changelog

All notable changes to this project will be documented in this file.

## [0.10.0] - 2025-03-06

### 📚 Documentation

- Stuff

## [0.9.0] - 2025-03-06

### 🐛 Bug Fixes

- Whatever

[...]

PR #1227, opened 2025-07-28
fix: topological sorting of tags

mbabauer:
…git-cliff/issues/1080)

## Description

This is an attempt to fix the issue described in Issue #1080.  

## Motivation and Context

When multiple tags are included in the next release and the tags jump up in digit size, ex moving from 1.9.0 to 1.10.0, the "highest" version was being identified as 1.9.0 due to the tags being sorted alphabetically instead of by the actual version.  This causes the next version to be misidentified as 1.10.0 instead of 1.11.0.

The `semantic_version_compare` function attempts to sort by version. If it fails, it should fall back to alphabetical sort (see like 589).  All my changes were to `repo.rs`. The other files are the results of me running the `cargo +nightly fmt`. I am fairly new to the Rust toolchain so if that's not correct and you need me to revert those changes I can.

This is an attempt to fix [Issue #1080](https://github.com/orhun/git-cliff/issues/1080).

## How Has This Been Tested?

1) First, I create a unit test to recreate the issue. The test creates tags `v0.9.0`, `v0.10.0` and `v0.11.0` on a temporary repo. The test breaks because it sorted the `v0.9.0` tag would sort to the end of the vector. Once I could see it happening in a repeatable test I...
2) Created the fix so that the test passed.  In the test I should expect the sorted order to be: `v0.9.0`, `v0.10.0` then `v0.11.0`.
3) I also included a unit test to test the new `semantic_version_compare` function to have a complete set of tests.  It tries to test some "edge" cases, like ensuring `v1.0` is less than `v1.0.0` and how to handle things like alpha releases and the fall-back to alpha-sorting when the comparison is between 2 non- Semver strings.

## Types of Changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation (no code change)
- [ ] Refactor (refactoring production code)
- [x] Other - formatting changes due to running the formatter tool

## Checklist:

- [x] My code follows the code style of this project.
- [ ] I have updated the documentation accordingly.
- [x] I have formatted the code with [rustfmt](https://github.com/rust-lang/rustfmt).
- [x] I checked the lints with [clippy](https://github.com/rust-lang/rust-clippy).
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #1457, opened 2026-04-01
fix/issue 1080

kristof-mattei:
- **fix: fixing bug identified in [Issue #1080](https://github.com/orhun/git-cliff/issues/1080)**
- **chore: re-running fmt with nightly**
- **chore: re-re-running fmt with nightly**
- **chore: switching out unwrap() with expect() per linter**
- **chore: swapping to plain array per linter**
- **chore: adding additional scenarios in  to cover missing code lines**
- **fix: address PR comments**


orhun:
Hello, this one is the same as #1227, but with the suggestions applied?

PR #1477, opened 2026-04-07
refactor(deps): replace dirs_next with etcetera

tessus:
## Description

This changed replaces the crate `dirs` with `etcetera`. The dirs crate unfortunately still returns the `~/Library/...` directories and does not follow the XDG standard on macOS (which is also a Unix system at its core).
The directories in `~/Library/...` are supposed to be used by macOS GUI applications and not terminal apps.

However, the change still maintains backwards compat in case a config file is stored in `~/Library/Application Support/git-cliff`.

## Motivation and Context

see #1475

## How Has This Been Tested?

- cargo test
- manual testing

## Types of Changes

- [ ] Other 

## Checklist:

  - [x] `cargo test`

### DIFF
--- a/git-cliff-core/Cargo.toml
+++ b/git-cliff-core/Cargo.toml
@@ -9,7 +9,7 @@ homepage = "https://github.com/orhun/git-cliff"
 repository = "https://github.com/orhun/git-cliff"
 keywords = ["changelog", "generator", "conventional", "commit"]
 edition = "2024"
-rust-version = "1.85.1"
+rust-version = "1.87.0"
 
 [features]
 default = ["repo"]
@@ -52,7 +52,7 @@ glob.workspace = true
 regex.workspace = true
 log.workspace = true
 secrecy.workspace = true
-dirs.workspace = true
+etcetera.workspace = true
 thiserror = "2.0.17"
 serde = { version = "1.0.228", features = ["derive"] }
 serde_json = "1.0.145"
--- a/git-cliff-core/src/config.rs
+++ b/git-cliff-core/src/config.rs
@@ -3,6 +3,7 @@ use std::str::FromStr;
 use std::sync::LazyLock;
 use std::{fmt, fs};
 
+use etcetera::{BaseStrategy, choose_base_strategy};
 use glob::Pattern;
 use regex::{Regex, RegexBuilder};
 use secrecy::SecretString;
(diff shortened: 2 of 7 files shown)


git-cliff-core/Cargo.toml
```diff
@@ -9,7 +9,7 @@ homepage = "https://github.com/orhun/git-cliff"
 repository = "https://github.com/orhun/git-cliff"
 keywords = ["changelog", "generator", "conventional", "commit"]
 edition = "2024"
-rust-version = "1.85.1"
```
orhun, line 12: This should be done in a separate commit/PR (for visibility in the changelog)

git-cliff-core/src/remote/mod.rs
```diff
@@ -99,6 +100,8 @@ impl Remote {
         if !self.is_set() {
             return Err(Error::RemoteNotSetError);
         }
+        let strategy = choose_base_strategy()
+            .expect("cannot determine current OS's default strategy (layout)");
```
orhun, line 105: Can you add a comment here for why this can't panic etc. (similar to what we did in `rustypaste`)

git-cliff-core/src/config.rs
```diff
@@ -488,34 +489,25 @@ impl Config {
...
         for supported_path in [
+            strategy.config_dir().join("git-cliff").join(DEFAULT_CONFIG),
+            // paths for backwards compatibility
             #[cfg(target_os = "macos")]
-            Some(Config::retrieve_xdg_config_on_macos().join(DEFAULT_CONFIG)),
-            dirs::config_dir().map(|dir| dir.join("git-cliff").join(DEFAULT_CONFIG)),
+            strategy
+                .home_dir()
+                .to_path_buf()
+                .join("Library/Application Support/git-cliff")
```
orhun, line 507: Is hardcoding this safe?

git-cliff-core/Cargo.toml
orhun: also, is this MSRV bump related to the `etcetera` dependency or something else?

tessus: At first yes, since etcetera 0.11.0 requires 1.87.0. I looked into the changes between etcetera 0.10.0 and 0.11.0 and it removes the dependency for an unmaintained crate.
But I am not sure why the MSRV in etcetera was raised, since there is no reason for it. Or at least I couldn't find a valid reason.

Then I've noticed that 1.85.1 is over a year old. Most Rust projects keep the MSRV to current -2, -4, or at max at -6. 1.85.1 is -9 or -10.
In other projects I have experienced issues when the MSRV was too old, especially in Docker builds.

Either way, I didn't think much about raising it 2 versions, which is still pretty old.

Please let me know, if this is an issue.

As for a seperate PR: How about I create 2 commits in this PR and in case it is merged, don't squash them. On the other side I can also create a separate PR, but in the end it makes no difference. Whatever you prefer.

git-cliff-core/src/remote/mod.rs
tessus: I added the comment in the other file and thought one comment is enough as I didn't want to duplicate the comment. 

https://github.com/orhun/git-cliff/pull/1477/changes#diff-d6715128f2b8aa1fef49c9d3e4a98067df34d71aad00ba6350d7302bdbe72bf9R497

But I can add the same comment here.

git-cliff-core/src/config.rs
tessus: Yes, it is safe. We are using the same in rustypaste-cli. (I have tested it. It works as it is supposed to.) May I ask why you don't think so? Or did I miss something? Your question caught me a bit off guard. I couldn't think of anything that would make this unsafe.

This path is only checked on macOS for backwards compat and in case there isn't a config file in the `~/.config/...` location. Btw, the dirs crate also "hard-codes" this path: `$HOME + /Library/Application Support/` (as config dir).

git-cliff-core/Cargo.toml
orhun: If we don't squash then the credits in the changelog might be a bit weird. Maybe we could mark this commit as breaking and add a comment about MSRV bump in the commit body? Not sure if that or rebasing make more sense here.

git-cliff-core/src/remote/mod.rs
orhun: Please do :)

git-cliff-core/src/config.rs
orhun: Oh I'm not familiar with macOS paths so that's why I asked. If you say so then it's all good 👍🏼

git-cliff-core/Cargo.toml
tessus: > If we don't squash then the credits in the changelog might be a bit weird.

Ah, because they will point to the same PR and the PR would have a title that doesn't mention MSRV? Hmmm, I see.

Ok, I'll create a separate PR and after it is merged I'll rebase this one.

git-cliff-core/src/remote/mod.rs
tessus: Done.

git-cliff-core/Cargo.toml
tessus: here's the new PR: #1479

orhun (approved):
thx!

PR #1482, opened 2026-04-17
feat(config): search .cliff.toml as alternative config file name

niklasmarderx:
Closes #1446

## What this does

When git-cliff searches for a config file it now checks both `cliff.toml` and `.cliff.toml`, in that order. Users who keep their configs under dot-prefixed names (a pattern common with `.eslintrc`, `.rustfmt.toml`, `.prettierrc`, etc.) can do the same for git-cliff without any extra flags.

## How it works

A new `CONFIG_CANDIDATES` constant in `git-cliff-core` lists the names to try:

```rust
pub const CONFIG_CANDIDATES: &[&str] = &[DEFAULT_CONFIG, ".cliff.toml"];
```

All three lookup sites use this slice:

- **`Config::retrieve_config_path`** — user-level config directory (and legacy macOS paths)
- **parent-directory walk** in `run_with_changelog_modifier` — the ancestor search that finds a `cliff.toml` in a parent directory

`cliff.toml` is still index 0, so it always wins over `.cliff.toml` when both exist. `DEFAULT_CONFIG` and the `--config` default are unchanged, so existing setups are completely unaffected.

## Testing

Added a unit test to `config::test` that asserts `cliff.toml` is first in `CONFIG_CANDIDATES` and that `.cliff.toml` is present.

### DIFF
--- a/git-cliff-core/src/config.rs
+++ b/git-cliff-core/src/config.rs
@@ -11,7 +11,7 @@ use serde::{Deserialize, Serialize};
 
 use crate::embed::EmbeddedConfig;
 use crate::error::Result;
-use crate::{DEFAULT_CONFIG, command, error};
+use crate::{CONFIG_CANDIDATES, command, error};
 
 /// Default initial tag.
 const DEFAULT_INITIAL_TAG: &str = "0.1.0";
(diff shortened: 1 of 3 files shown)


orhun:
thanks for the PR, but this was already implemented in #1448

PR #1483, opened 2026-04-17
feat(context): expose determined bump type in release context

### LINKED ISSUE #1220: Add determined version bump type inside git-cliff context

### Feature description

Add version type key to context/`--context` (ex. 'major', 'minor', 'patch', 'unchanged'/'null', etc.). With this feature, users can parse how the current version was bumped from the previous for use in changelogs or CI pipelines and other tools. This can currently be inferred by parsing the version keys at the top level and within the 'previous' key in the context json object, however having the functionality built-in would make it much easier to do this kind of work.

Split out from conversation regarding #1091.

### Desired solution

A 'version-change-type' key next to 'version' whose value can indicate any of the possible semver types git-cliff can bump between (I believe this excludes pre-releases? ex. rc/dev/beta/alpha?):
- 'major'
- 'minor'
- 'patch'

The value can be `null` in the case that there was no change available between the current and previous version (first/default version or unable to automatically bump, for example: no conventional commit history, etc.).

### Alternatives considered

- `--bumped-type` cli argument to output the bumped version type
- Only including 'version-change-type' key when using --bump (this is the most likely use case for this feature, but it reduces information available in the context for previous versions, so it might just make more sense to include a 'version-change-type' value in the context for any version that follows a previous version)

### Additional context

_No response_

ChihebBENCHEIKH1:
Closes #1220

Adds a `bump_type` field to the `Release` struct, exposing the determined version bump type in both templates and the JSON context output (`-x`).

**Values:** `"major"`, `"minor"`, `"patch"`, or `null`

**Usage in templates:**
{% if bump_type == "major" %}Breaking release{% endif %}

**In JSON context (`--context`):**
{"version": "1.2.0", "bump_type": "minor", ...}

**How it works:**
- When `bump_type` is forced via config, uses that directly
- Otherwise, compares the previous and next semver versions to determine the type
- Returns `null` when no previous version exists (initial tag)

**Changes:**
- Added `bump_type: Option<BumpType>` to `Release` struct
- Modified `calculate_next_version_with_config` to return the determined bump type
- Updated `bump_version()` to populate the field
- Added `bump_version_type` test covering auto-detection, forced config, and initial tag cases

### DIFF
--- a/git-cliff-core/src/release.rs
+++ b/git-cliff-core/src/release.rs
@@ -47,6 +47,9 @@ pub struct Release<'a> {
     pub statistics: Option<Statistics>,
     /// Arbitrary data to be used with the `--from-context` CLI option.
     pub extra: Option<Value>,
+    /// The type of version bump that was applied.
+    #[serde(rename = "bump_type")]
+    pub bump_type: Option<BumpType>,
     /// Contributors.
     #[cfg(feature = "github")]
     pub github: RemoteReleaseMetadata,
@@ -85,7 +88,7 @@ impl Release<'_> {
     ///
     /// It uses the default bump version configuration to calculate the next
     /// version.
-    pub fn calculate_next_version(&self) -> Result<String> {
+    pub fn calculate_next_version(&self) -> Result<(String, Option<BumpType>)> {
         self.calculate_next_version_with_config(&Bump::default())
     }
 
@@ -104,7 +107,10 @@ impl Release<'_> {
     ///
     /// It uses the given bump version configuration to calculate the next
     /// version.
-    pub(super) fn calculate_next_version_with_config(&self, config: &Bump) -> Result<String> {
+    pub(super) fn calculate_next_version_with_config(
+        &self,
+        config: &Bump,
+    ) -> Result<(String, Option<BumpType>)> {
         match self
             .previous
             .as_ref()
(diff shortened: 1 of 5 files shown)


ChihebBENCHEIKH1:
Nope, all done! The force pushes were just formatting fixes. CI is green now.

git-cliff-core/src/release.rs
```diff
@@ -85,7 +88,7 @@ impl Release<'_> {
     ///
     /// It uses the default bump version configuration to calculate the next
     /// version.
-    pub fn calculate_next_version(&self) -> Result<String> {
+    pub fn calculate_next_version(&self) -> Result<(String, Option<BumpType>)> {
```
orhun: Instead of a tuple here, wouldn't be more convenient/idiomatic to create a struct, called `NextVersion`?

it clashes with the `next_version::NextVersion`, but we can rename the import :)

orhun (commented):
Looks good overall, just one minor thing :)

git-cliff-core/src/release.rs
ChihebBENCHEIKH1: Good call! Introduced a NextVersion struct and renamed the next_version::NextVersion import to NextVersionTrait to avoid the clash.

orhun (approved):
LGTM, thanks!

Issue #655, opened 2024-05-26
Support processing the commits starting from a specific tag


landerss1:
### Is there an existing issue or pull request for this?

- [X] I have searched the existing issues and pull requests

### Feature description

Currently, there's support for processing commits starting from the current or latest tag, or commits without any tag, but not starting from a specific tag. This would make the transition from unconventional commits to conventional commits  much smoother. Then, you could keep the old changelog as, e.g. `HISTORY.md`, and then start a new `CHANGELOG.md` based on commits after a specific tag.

### Desired solution

Introduce a new command-line argument that can be used to specify the tag that should be used to start fetching commits from.

### Alternatives considered

Can't think of any.

orhun:
Hello!

Have you tried the git `range` option?

e.g.

```sh
git cliff v0.1.0..HEAD
```

Or even more shorter:

```sh
git cliff v2.2.1..
```

Other examples:

```sh
git cliff 4c7b043..a440c6e
git cliff 4c7b043..HEAD
git cliff HEAD~2..
```

I think this is what you are looking for :)

landerss1:
Ah, Thanks. I saw the range CLI argument in the docs, but there was no description or examples. I have tried it now, and it solves my problem, but you have to go one tag further back than expected, so you have to specify `range start=<desired-tag>-1`. Is this what's to be expected?

orhun:
I think yes, since [git ranges](https://git-scm.com/docs/git-range-diff) work like that. It is handled by the underlying git library (`git2`).

I expanded the examples in d4512521fbcfb971c94aa7794d78bced1ddec7a1

lermontex:
@orhun, as far as I understand, git ranges will conflict with the --unreleased, --latest, etc. flags, so git ranges is not entirely suitable, since it will not allow flexible configuration

It would be great if it was possible to specify the tag from which to start the story

There is an initial_tag parameter in the configuration, maybe it’s worth using it?

https://git-cliff.org/docs/configuration/bump/#initial_tag

orhun:
> as far as I understand, git ranges will conflict with the --unreleased, --latest, etc. flags, so git ranges is not entirely suitable, since it will not allow flexible configuration

Yeah, using those flags will override the commit range.

Can you give me an example case (i.e. git history) to make this matter more clear? I'm happy to implement it as long as we have the correct behavior with [test fixtures](https://github.com/orhun/git-cliff/tree/main/.github/fixtures#fixtures) in place.

[...]

Issue #1122, opened 2025-03-27
Monorepo / path filters might exclude tags


JP-Ellis:
### Is there an existing issue for this?

- [x] I have searched the existing issues

## Background

Let's suppose we have a monorepo with two projects and the following structure:

```
.github/
project-foo/
project-bar/
README.md
```

Releases are tagged `foo-vX.Y.Z` and `bar-vX.Y.Z` for the two projects.

Let us suppose, for this scenario, that I just made some changes to the way the packages are released in `.github/` and I release new version of `bar`, tagging that commit that _only_ modifies files within `.github/`.

## Problem

It would appear that Git Cliff filters commits _first_ and then looks for release tags within these commits.

The problem here is that by filtering commits first and listing tags from the resulting subset, the commit (and therefore associated tag) in the scenario above also gets removed.

While it is correct to filter out the commit from the changelog for `bar` since it doesn't touch any files within `project-bar/`, I would say that it is incorrect for Git Cliff to be filtering the _tags_ as well.

## Solution

I'm not familiar with Git Cliff's internals, but I think Git Cliff needs to:

1. Match the release tags _irrespective_ of the associated commit's changes; and then,
2. Generate the changelog from the subset of commits that match the include/exclude paths, with the tags from (1) merely providing labels to help group the commits.

### Steps To Reproduce

Replicate the scenario above

### Expected behavior

The version tag is correctly picked up, even if the associated commit does not touch any files within the path filters.

### Software information

- Operating system: `Darwin 24.3.0` (macOS 15.3)
- Rust version: N/A
- Project version: 2.8.0

### Additional context

See [this Discord thread](https://discord.com/channels/1093977388892819553/1354325727834542151)

orhun:
@Cyclonit are you adding a fixture for this and testing it out? 👀

Cyclonit:
@orhun, yes I'll work on this in the evening. I reckon we don't really have to care about path filters etc. Instead I would change the commit processing, such that ignored commits aren't dropped outright. They need to be kept until all of the release tags have been found and processed.

JP-Ellis:
Is there any help I can offer here? Perhaps through a PR review, or to do some testing?

sam-stenvall-iceye:
This affects us as well, it makes git-cliff very hard to use correctly since you have to remember to touch a file pertaining to a particular `--include-path` whenever you tag a release, otherwise that tag will never get picked up.

ognis1205:
Following.

Issue #208, opened 2023-06-30
relevant tags are missing when using include-path


cjorge-graphops:
**Describe the bug**

When using --include-path, the first release tag goes missing.
Also, if there is a tag that points to a commit that is filtered out by --include-path, then that tag no longer shows on the change log and the commits that do show, show as unreleased instead.

**To reproduce**

I'm running the following script to reproduce the issues:

```
#!/bin/bash

rm -Rf reproduce-repo
mkdir reproduce-repo
cd reproduce-repo
git init
mkdir included-path
touch included-path/{a,b,c}
git add included-path/a
git commit -m "feat: add a"
git tag release-v0.1.0
git add included-path/b
git commit -m "feat: add b"
git tag release-v0.2.0
git add included-path/c
git commit -m "feat: add c"
git tag release-v0.3.0
git-cliff -i
sed -i 's/tag_pattern = .*/tag_pattern = "release*"/g' cliff.toml
echo "######## WITHOUT INCLUDE PATH #############"
git-cliff -c cliff.toml
echo "######## WITH INCLUDE PATH, FIST TAG MISSING #############"
git-cliff -c cliff.toml --include-path "included-path/**"
mkdir other-path
touch other-path/d
git add other-path/d
git commit -m "feat: add d"
touch included-path/e
git add included-path/e
git commit -m "feat: add e"
touch other-path/f
git add other-path/f
git commit -m "feat: add f"
git tag release-v0.4.0
echo "######## WITH INCLUDE PATH, SHOWS AS UNRELEASED #############"
git-cliff -c cliff.toml --include-path "included-path/**"
```

**Expected behavior**

I expected that when using --include-path, all the first three releases would show on the change log and instead the first one is missing.
I get this, which misses release-v0.1.0 tag:

> 
> # Changelog
> 
> All notable changes to this project will be documented in this file.
> 
> ## [release-v0.3.0] - 2023-06-30
> 
> ### Features
> 
> - Add c
> 
> ## [release-v0.2.0] - 2023-06-30
> 
> ### Features
> 
> - Add b
> 
> 

Regarding release-v0.4.0, I would expect it to show on change log but the relevant commit shows as unreleased:

> # Changelog
> 
> All notable changes to this project will be documented in this file.
> 
> ## [unreleased]
> 
> ### Features
> 
> - Add e
> 
> ## [release-v0.3.0] - 2023-06-30
> 
> ### Features
> 
> - Add c
> 
[...]

PR #1489, opened 2026-04-24
feat(cli): introduce commit ranges with interval-notation endpoint flags

cds-amal:
This PR refactors commit-range selection on the CLI by borrowing math-interval notation, which captures both **location** (which commit) and **inclusivity** (whether the endpoint is in the range). A range is bounded by a `start` and an `end`; `[` / `]` are inclusive, `(` / `)` are exclusive.

Internally, the range is now a pure value (`CommitRange` with explicit `Endpoints`) flowing through a transform -> resolve -> execute pipeline that's inspectable (`--dry-run`) and unit-testable.

### New endpoint options

| CLI flag          | Config key          | Meaning                                  |
| ----------------- | ------------------- | ---------------------------------------- |
| `--start-at X`    | `start_at = "X"`    | Include `X`; walk forward. `[X, ...`     |
| `--start-after X` | `start_after = "X"` | Exclude `X`; walk forward. `(X, ...`     |
| `--end-at Y`      | `end_at = "Y"`      | Include `Y`; walk back. `..., Y]`        |
| `--end-before Y`  | `end_before = "Y"`  | Exclude `Y`; stop before it. `..., Y)`   |

Naming convention: `*_at` is inclusive, `*_after` / `*_before` is exclusive. Within each pair, at most one may be set. The two pairs are independent, so any inclusivity combination is describable.

### How legacy flags map to endpoints

Every git-cliff invocation already selects a contiguous slice of commits; the legacy flags are named shortcuts for picking the two endpoints of that slice, with inclusivity baked in.

| Legacy         | Equivalent endpoint form                              |
| -------------- | ----------------------------------------------------- |
| (no flags)     | `--end-at HEAD` (left defaults to first commit)       |
| `--unreleased` | `--start-after <last_tag> --end-at HEAD`              |
| `--latest`     | `--start-after <prev_tag> --end-at <last_tag>`        |
| `--current`    | `--start-after <prev_tag> --end-at <current_tag>`     |
| `<A>..<B>`     | `--start-after A --end-at B`                          |

### Conflicts and precedence

- `--start-at` and `--start-after` cannot be combined; same for `--end-at` and `--end-before`. This holds within the CLI and within `cliff.toml` independently.
- Across sources, the CLI overrides config one side at a time. Setting `--start-at` or `--start-after` replaces both `start_at` and `start_after` from config (so `--start-at X` cleanly wins over a stale `start_after = "Y"` in `cliff.toml`); the right side behaves the same way.
- The new endpoint options cannot be combined with the legacy range flags (`--latest`, `--current`, `--unreleased`, `--bump`, positional `A..B`). Pick one style.

### Previewing with `--dry-run`

`--dry-run` prints the computed interval, the number of commits it covers, and the git revision range that will be walked, then exits without rendering a changelog:

```
$ git cliff --start-at v0.1.0 --end-at v0.2.0 --dry-run
range:    [v0.1.0, v0.2.0]                <-- user-facing interval
commits:  3                               <-- commit count
emitted:  02deb7a7...^..9f14f5d1...       <-- git revision range
```

### Examples

Select a range with explicit endpoint inclusivity (`*-at` is inclusive; `*-after` / `*-before` is exclusive):

```bash
# include v1.0.0 itself; walk forward to HEAD
git cliff --start-at v1.0.0

# everything since v1.0.0, but not v1.0.0 itself
git cliff --start-after v1.0.0

# everything up to and including v2.0.0
git cliff --end-at v2.0.0

# strictly between two tags, both excluded
git cliff --start-after v1.0.0 --end-before v2.0.0
```

These options can also be set in `cliff.toml` under `[git]`:

```toml
[git]
start_after = "v1.0.0"
end_at      = "v2.0.0"
```

## Motivation and Context

This started from a discussion / research described in #655 .The legacy commit-range flags (`--latest`, `--current`, `--unreleased`, positional `A..B`) are named shortcuts that bundle endpoint location and inclusivity together: `--unreleased` resolves to "after the last tag, up to and including HEAD", `--latest` to "after the previous tag, up to and including the last tag", and so on. Inclusivity is implicit, locked into each flag's semantics. Adding new endpoint kinds (Nth-most-recent-tag, dates, branch tips) would either expand that combinatorics or thread a parallel path through argument parsing, repo traversal, and release assembly.

Math-interval notation factors location from inclusivity cleanly: `[X, Y]` is bounded inclusive on both sides; `(X, Y)` is bounded exclusive; `[X, Y)` mixes them. Modeling a `CommitRange` as two optional `Endpoint`s (each with `rev` and `inclusive`) captures any combination as a pure data value flowing through a transform -> resolve -> execute pipeline. Future endpoint kinds become pure follow-ups: new transforms thread through the same data path, no fan-out across the CLI / config / walker layers. `--dry-run` falls out for free as "render the interval and the resolved revspec, then exit".

### Implications of this architectural change

See: [this issue comment](https://github.com/orhun/git-cliff/issues/655#issuecomment-4322435127)

No tracking issue.

## How Has This Been Tested?

The collapsed block below records actual `--dry-run` output across the verified scenarios on this very repo (git-cliff itself); the test plan after it lists what reviewers should rerun locally and what CI exercises.

<details>
<summary><strong>Verified end-to-end (click to expand)</strong></summary>

Spot-checks that the four config keys flow through the same pipeline as their CLI counterparts, that the CLI overrides config (same-key, cross-key, and partial-side), and that the trickier walker shapes (right-only exclusive) resolve cleanly. All runs use `--dry-run` so no changelog is written.

**TOML drives the range:**

```bash
$ cat cliff-range.toml
[git]
start_after = "v2.12.0"
end_at      = "v2.13.0"

$ git cliff --config cliff-range.toml --dry-run
range:    (v2.12.0, v2.13.0]
commits:  85
emitted:  988e8638...^..d2354923...
```

Flipping inclusivity on both sides:

```bash
$ cat cliff-range.toml
[git]
start_at   = "v2.12.0"
end_before = "v2.13.0"

$ git cliff --config cliff-range.toml --dry-run
range:    [v2.12.0, v2.13.0)
commits:  85
emitted:  988e8638...^..d2354923...^
```

Only one side set; the other falls back to its default:

```bash
$ cat cliff-range.toml
[git]
end_at = "v2.13.0"

$ git cliff --config cliff-range.toml --dry-run
range:    [<first-commit>, v2.13.0]
commits:  1546
emitted:  d2354923...
```

Conflict detection fires from TOML alone:

```bash
$ cat cliff-range.toml
[git]
start_at    = "v2.12.0"
start_after = "v2.12.0"

$ git cliff --config cliff-range.toml --dry-run
Error: ArgumentError("`start_at` and `start_after` are mutually exclusive")
```

**CLI overrides config, same key.** TOML pins the upper bound to `v2.13.0`; the CLI bumps it to `v2.13.1`:

```bash
$ cat cliff-range.toml
[git]
start_after = "v2.12.0"
end_at      = "v2.13.0"

$ git cliff --config cliff-range.toml --end-at v2.13.1 --dry-run
range:    (v2.12.0, v2.13.1]
commits:  87
emitted:  988e8638...^..5d3a6702...
```

**CLI overrides config, cross-key (flips inclusivity).** Config sets `start_after = "v2.12.0"` (exclusive); CLI passes `--start-at v2.12.0` (inclusive). The CLI replaces the entire left side, so the resolved range becomes `[v2.12.0, v2.13.0]` and the commit count picks up `v2.12.0` itself:

```bash
$ git cliff --config cliff-range.toml --start-at v2.12.0 --dry-run
range:    [v2.12.0, v2.13.0]
commits:  86
emitted:  988e8638...^..d2354923...
```

**CLI overrides one side; config still drives the other.** CLI moves the left endpoint back to `v2.11.0`; the right endpoint stays where the config put it:

```bash
$ git cliff --config cliff-range.toml --start-after v2.11.0 --dry-run
range:    (v2.11.0, v2.13.0]
commits:  102
emitted:  3ddfb1ee...^..d2354923...
```

Sanity: `git rev-list --count v2.12.0..v2.13.0` is `85`, `..v2.13.1` is `87`, `v2.12.0^..v2.13.0` is `86`, and `v2.11.0..v2.13.0` is `102`, all matching the dry-run counts above.

**Right side only, exclusive.** No left bound; `--end-before` gives an upper-exclusive range that walks back from the parent of the named tag:

```bash
$ git cliff --end-before v2.12.0 --dry-run
range:    [<first-commit>, v2.12.0)
commits:  1460
emitted:  988e8638...^
```

Compare with `--end-at v2.12.0` for the inclusive sibling: `[<first-commit>, v2.12.0]` at 1461 commits (one more, picking up `v2.12.0` itself).

</details>

### Test plan

- [ ] `cargo test -- --skip "repo::test::git_upstream_remote"` passes
- [ ] `cargo test --no-default-features -- --skip "repo::test::git_upstream_remote"` passes
- [ ] CI Nix flake check passes (skips `resolve_with` tests like `command` / `repo`)
- [ ] CI fixture matrix includes `test-range-start-at-inclusive`, `test-range-end-before-exclusive`, and the new `test-range-end-before-only` (right-only-exclusive regression)
- [ ] Manual: `git cliff --start-at <tag> --end-at <tag> --dry-run` prints expected interval/commits/revspec
- [ ] Manual: `git cliff --end-before <tag> --dry-run` (no left bound) prints `[<first-commit>, <tag>)` and resolves without error
- [ ] Manual: legacy flags (`--latest`, `--current`, `--unreleased`, `A..B`) still produce identical changelogs to `main`

See the collapsed "Verified end-to-end" block under [How Has This Been Tested?](#how-has-this-been-tested) above for the full set of `--dry-run` logs.

## Types of Changes

(Notes: the `--end-before` walker fix is for in-PR code, not released behavior, so I left "Bug fix" unchecked. Every change here is additive or strictly more permissive: the new endpoint flags are new surface, the `set_commit_range` widening accepts a superset of what it used to, and legacy flag behavior is unchanged.)

## Checklist:

  - [x] `cargo +nightly fmt --all` (verified clean via `--check`).
  - [x] `cargo clippy --tests --verbose -- -D warnings` — no new warnings from `main`. (Pre-existing `clippy::unnecessary_sort_by` hits in `git-cliff-core` are present on `main` and untouched by this branch.)
- [?] All new and existing tests passed locally, except for 2-3 persistent woes on CI


cds-amal:
Hi @orhun , I don't know how to proceed with the failing tests; any advice please?

orhun:
Seems like a flaky test due to a network error (404) - I just restarted the job.

leoweyr:
One dimension that I think deserves an explicit decision in this redesign: **how range selection composes with path filtering (`--include-path` / `--exclude-path`).**

This matters specifically for monorepos, where a changelog is usually generated per package and therefore needs to be scoped on two axes at once — a commit range (the slice belonging to the release being prepared) *and* a path (the files belonging to that package).

### Current behavior

Today, when an explicit positional range is combined with a path filter, the range is silently dropped and git-cliff emits the entire path-filtered history instead of the requested slice:

```bash
# Intent: only the commits between the two tags that also touch the package path
git cliff --tag-pattern '^pkg/v[0-9].*' --include-path 'pkg/**' pkg/v1.0.0..pkg/v1.1.0
# Actual: the range is ignored; every path-matching commit across all history is emitted
```

Removing `--include-path` makes the same positional range behave as expected, which points to path filtering being applied *before* range resolution — the resolver then operates on the already-filtered subset and loses the requested endpoints. This is the same root cause discussed in #1122 and #208, where path filtering also drops or mislabels the tags that the range/selection logic depends on.

The practical fallback is `--current` / `--latest`, which derive the endpoints from tags internally and *do* respect `--include-path`. That works, but only because they recompute the range themselves; it gives up the ability to specify an arbitrary explicit range while a path filter is active.

### Why it's relevant here

The new endpoint flags (`--start-at` / `--start-after` / `--end-at` / `--end-before`) and the legacy `A..B` form they generalize are all expressed in terms of tags and commits. If endpoint resolution still happens on top of an already path-filtered commit set, the new flags will inherit the same limitation: a tag named as an endpoint may have been filtered out, and the resolved interval will not match what the user wrote.

### Suggested direction

Resolve range endpoints against the **full** tag/commit graph first, then apply `--include-path` / `--exclude-path` to the commits *within* the resolved interval — i.e. select the range, then filter, rather than filter, then select. This is consistent with the "match tags first, then filter commits" direction proposed in #1122, and it would let the interval model introduced here remain correct and composable in monorepo setups.

A `--dry-run` that prints the resolved revision range is a great fit for verifying this: it would let users confirm that the interval is honored independently of any active path filters.

cds-amal:
Thanks for this feedback @leoweyr. I need some time to think this through; I definitely did not consider monorepos and the path-issues they present. 🤔

orhun:
Hey, just catching up with PRs :) Before I dive into this, can you point me to the issues that this one fixes/related to? I need some context beforehand otherwise it might take me some time to digest the changes.

cds-amal:
Hey @orhun, thanks! I was just about to list what I know about @leoweyr's concern. 

Architecturally, this PR factors commit-range selection into two independent axes: location (which commit) and inclusivity (is the endpoint in the range). Modeled as math-interval notation, that gives four endpoint flags (--start-at / --start-after / --end-at / --end-before), and every legacy flag (--latest, --current, --unreleased, positional A..B) becomes a named shortcut over the same two endpoints. The payoff: future endpoint kinds (dates, Nth-most-recent-tag, branch tips) thread through one data path instead of fanning out across the CLI / config / walker layers.

The concrete feature request behind it is #655 ("start the walk from a specific tag"); that's exactly --start-at here. The design write-up (why interval notation, how the legacy flags map onto it) is the best thing to read before the diff: https://github.com/orhun/git-cliff/issues/655#issuecomment-4322435127.

No other tracking issue. One review thread to flag: @leoweyr asked whether the endpoint flags compose with --include-path in monorepos; they do, and the test-range-include-path-dry-run fixture locks it (details in the PR description under "Range/path composition"). The one wrinkle there is pre-existing and orthogonal to this PR: --include-path's greedy num_args(1..) swallows a trailing positional A..B, silently collapsing the legacy range. The new endpoint flags are named options, so they sidestep it by construction; I don't think this PR touches that bug either way (my gut is it neither fixes nor worsens it), but it's still worth its own fix. Obviously .. I don't want to rely on my guts :)

----
Small correction to my note above: the "Range/path composition" section isn't in the PR description yet (I'll fold it in shortly). Until then, here's the short version, so nothing's left dangling: the `test-range-include-path-dry-run` fixture runs `--start-after v2.6.1 --end-at v2.7.0 --include-path .github/fixtures/ --dry-run` in both flag orders against a single `expected.md`. The dry-run interval and count are path-unaware (the walker is seeded with the resolved range, then path-filters the result), so a regression that dropped the range would flip the output from `(v2.6.1, v2.7.0]` / 33 commits to the unbounded default and a much larger count, failing the diff.

Edit: added correction; I referenced an local-only work in progress investigation.

cds-amal:
This looks to be an argument-parsing issue in how `--include-path` is declared, not range resolution.

`--include-path` is declared `num_args(1..)`, so clap greedily consumes a trailing positional RANGE as one more include pattern. The range is never set; `git-cliff` then walks all of history and path-filters it, which reads as "the range was ignored". The same declaration exists on main, so this reproduces on released `git-cliff` too; 

it's a different mechanism from #1122/#208, which are about release splitting.

Put differently: the parser isn't intent-idempotent. By that I mean the property that the same arguments, reordered, always parse to the same intent; greedy multi-value options break it, so these two invocations mean different things to clap while meaning the same thing to the human who typed them:

```console
$ git cliff --include-path 'pkg/**' pkg/v1.0.0..pkg/v1.1.0 --dry-run
range:    [<first-commit>, HEAD]        <-- range swallowed as a glob

$ git cliff pkg/v1.0.0..pkg/v1.1.0 --include-path 'pkg/**' --dry-run
range:    (pkg/v1.0.0, pkg/v1.1.0]      <-- correct; `--` before the range also works
```

cds-amal:
Hi @orhun , I tracked the swallowed range down to clap's greedy `num_args(1..)` parsing and opened #1592, which fixes it on our side. I'd prefer bringing in this PR after 1592 lands. I used an LLM for code history research and test writing.

An aside, clap has some sharp edge cases. I wish the [bpaf crate for cli parsing](https://crates.io/crates/bpaf) would get more love from the community.

cds-amal:
Rebased.

orhun:
Hey!

I finally had some time to look at this:

- I think we should drop `--dry-run`: there isnt' any other CLI option that depend on it and we can simply depend on `debug!` logs instead.
- I think there is so much complexity added with this PR: there are now several overlapping ways to select commits, but users now need to understand which styles can be combined and how they relate. We should simplify this functionality somehow.
- I feel like there is some conflicts with `--unreleased` and `--latest` arguments and the newly added CLI options. I don't fully understand how they behave :/
- `limit_tags` feature seems broken (see the CI)

cds-amal:
Hey @orhun, thanks for the feedback. I agree that the CLI changes ended up adding a lot of complexity to this PR. I'd love to drop some scope here. Lemme review the current state with that in mind and post a simpler proposal.

PR #1518, opened 2026-05-06
feat(template): add commit_groups filter for ordered group rendering

### LINKED ISSUE #9: Group order doesn't match that of `commit_parsers`
**Describe the bug**
As I was tweaking my `cliff.toml` file to get the changelog to look exactly how I want it, I noticed the order of the groups is not affected by the elements in `commit_parsers`, but rather seems to be alphabetical on the group name.

**Expected behavior**
I hoped the groups would appear in the markdown in the same order as the `commit_parsers` array. Maybe there's a way to work around this? Thanks!

ChrisJr404:
Closes #9.

## Background

`commits | group_by(attribute="group")` is Tera's built-in filter, and it backs the grouped result with a `BTreeMap`. That means groups always come out in alphabetical (or emoji-codepoint) order, regardless of the order in which they were declared in `commit_parsers`. The thread on #9 covers this and the workaround that's spread across the ecosystem: prefix group names with `<!-- 0 -->`, `<!-- 1 -->`, ..., then strip the comments with `striptags`. The project's own `cliff.toml` uses that workaround.

## What this PR does

It leaves `group_by` alone (so existing templates keep working byte-for-byte) and adds a new filter, `commit_groups`, that returns an array of `{ group, commits }` records instead of a map. Because the result is an array, iteration order is deterministic and template-defined.

The filter has two modes:

- No `groups` argument: groups appear in the order they first occur in the commit list, which matches commit chronology.
- With a `groups=...` argument (an array of group names): groups are sorted to match that order. Any group not in `groups` is appended afterward in first-appearance order.

To make the common case ergonomic, `Changelog::build` now injects the ordered list of group names from `commit_parsers` into the template context as `commit_parsers_groups`. So a user template can simply write:

```tera
{% for entry in commits | commit_groups(groups=commit_parsers_groups) %}
### {{ entry.group }}
{% for commit in entry.commits %}- {{ commit.message }}
{% endfor %}{% endfor %}
```

and get groups in the order they wrote them in `cliff.toml`, with no `<!-- N -->` hack and no `striptags` filter. With the same input, switching to the existing `group_by` filter still produces alphabetical order, so this is purely additive.

## Reproducer from the issue

Given the configuration from arthrarnld's comment:

```toml
commit_parsers = [
  { message = "^feat", group = ":rocket: New features" },
  { message = "^fix",  group = ":bug: Bug fixes" },
  { message = "^perf", group = ":zap: Performance" },
  { message = "^chore", group = ":gear: Miscellaneous" },
]
```

and four commits in `chore -> fix -> perf -> feat` order, the existing `group_by` template renders:

```
### :bug: Bug fixes
### :gear: Miscellaneous
### :rocket: New features
### :zap: Performance
```

After this PR, switching the template to `commit_groups(groups=commit_parsers_groups)` renders:

```
### :rocket: New features
### :bug: Bug fixes
### :zap: Performance
### :gear: Miscellaneous
```

## Tests

- `template::test::test_commit_groups_filter_preserves_first_appearance_when_no_groups`
- `template::test::test_commit_groups_filter_uses_groups_argument`
- `template::test::test_commit_groups_filter_appends_unknown_groups`
- `template::test::test_commit_groups_filter_skips_null_groups`
- `changelog::test::changelog_group_order_matches_commit_parsers` (regression test using the issue reproducer end-to-end through `Changelog::generate`)

## Notes

- `group_by` is unchanged. Default templates are unchanged. Adoption is opt-in via the new filter.
- The `commit_parsers_groups` context key is set by `Changelog::build` before any user-supplied context is added, so `add_context("commit_parsers_groups", ...)` still wins if a caller wants to override it.
- This builds on the design discussion from the closed #339 PR (orhun's `commit-groups-filter-improved` branch) but keeps the state on the `Template` instance instead of a global static, and uses tera's filter-arg mechanism to plumb the order list through.

## Test plan

- [x] `cargo test --workspace --lib` (all new tests pass; pre-existing `repo::test::*` failures reproduce on `main` and are env-dependent)
- [x] `cargo clippy --all-targets --workspace` introduces no new lints
- [x] `cargo fmt --check` introduces no new diffs against the existing baseline
- [x] End-to-end smoke test with the binary against a 4-commit repo confirms three modes: explicit-order (matches `commit_parsers`), no-args (chronology), and existing `group_by` (alphabetical, unchanged)


orhun (approved):
Amazing. thanks a lot!

This solves a long-standing problem - so I really appreciate it!

Issue #712, opened 2024-06-17
`--prepend` cannot reliably identify the previous header


Cyclonit:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description of the bug

The CLI argument `--prepend` is supposed to result in git-cliff injecting the changelog for recent releases into an existing changelog.

1. Remove the old header.
2. Prepend the new releases' bodies.
3. Prepend the new header.

Step 1 relied on the header being static between runs of git-cliff. It performed a simple string comparison between the existing CHANGELOG.md and the current value in the `changelog.header` variable. Since #697, `changelog.header` contains a template. Thus its value might change between runs and thus prevent detecting the old header.

### Steps To Reproduce

1. Set `changelog.header` to a template such that the output after rendering changes between releases.
2. Create a new release.
3. Run git-cliff with the `--prepend` flag.

### Expected behavior

git-cliff should properly detect and replace the old header.

### Software information

- After PR #698

Cyclonit:
@orhun suggested to simply render the header template prior to running the old logic. However, this is likely insufficient. In cases where the header templating feature is used, the header is likely to change between releases. Thus rendering it will not help in detecting the old header.

If we want --prepend to work reliably, I propose to inject markers using html comments into the markdown instead:

```md
# Changelog

This is the changelog header.
<!-- END OF HEADER -->

## Release 2

- Lorem Ipsum

## Release 1

- Lorem Ipsum
<!-- END OF BODY -->

## Footer

This is the changelog footer.
```

orhun:
> If we want --prepend to work reliably, I propose to inject markers using html comments into the markdown instead:

Yes, I think that would be the best option.

Also see #433 - we might also parse the entire markdown and inject the new header. But your suggestion definitely sounds less painful.

ognis1205:
Following.

PR #1525, opened 2026-05-16
fix(changelog): normalize prepend boundary spacing

### LINKED ISSUE #1524: No newline between prepended block and old content when using --prepend

### Description of the bug

When using the --prepend option to insert new content at the top of an existing changelog file, there is no blank line between the newly prepended block and the original content. This causes the two sections to run into each other without the expected spacing.

This is especially obvious when both the prepended block (e.g., new generated changelog entries) and the old content (previous changelog) have headings: they appear directly one after another, harming readability.

### Steps To Reproduce

1. Generate a changelog with git-cliff and output it to a file (e.g., CHANGELOG.md)
2. Make additional commits
3. Run git-cliff again with --prepend CHANGELOG.md to add new changes
4. Open the file and note there is no blank line between the new section at the top and the previous content.

### Expected behavior

There should be at least one blank line (newline) between the prepended (newly generated) content and the existing content in the changelog file when using --prepend. This would ensure the new section is visually separated and readable.

### Screenshots / Logs

_No response_

### Software information

- Operating system: fedora 42
- Rust version: N/A
- Project version: 2.13.1

### Additional context

Workarounds in the configuration or template body don't reliably solve this. Fixing the join in the implementation of --prepend is more robust. See discussion: https://github.com/orhun/git-cliff/issues/712

jobcespedes:
## Description

Normalize `--prepend` output so the newly generated changelog block and existing changelog content are separated by exactly one blank line.

The change is localized to `Changelog::prepend` and does not alter normal non-prepend changelog generation. A regression test was added for the case where existing changelog content starts immediately with no leading newline.

## Motivation and Context

When `--prepend` writes generated content directly before existing changelog content, the boundary can end up without a blank line, which can violate Markdown linting expectations.

Fixes #1524

## How Has This Been Tested?

- Added regression test:
  - `changelog_prepend_inserts_blank_line_before_existing_content`
- Ran:
  - `git diff --check`

Attempted to run the targeted Cargo test:

```bash
cargo test -p git-cliff-core changelog_prepend_inserts_blank_line_before_existing_content --no-default-features
```

The test could not complete in this environment because the system C compiler/linker is unavailable (`cc` not found). Retrying with Rust’s bundled `rust-lld` also failed because required system libraries such as `libc`, `libm`, and `libpthread` were unavailable.

`cargo fmt --check` reports pre-existing formatting diffs across unrelated files, so I left those untouched to keep this change localized.

N/A

## Types of Changes

- [ ] Other

## Checklist:

  - [ ] `cargo +nightly fmt --all`
  - [ ] `cargo clippy --tests --verbose -- -D warnings`
  - [ ] `cargo test`

### DIFF
--- a/git-cliff-core/src/changelog.rs
+++ b/git-cliff-core/src/changelog.rs
@@ -640,8 +640,21 @@ impl<'a> Changelog<'a> {
         if let Some(header) = &self.config.changelog.header {
             changelog = changelog.replacen(header, "", 1);
         }
-        self.generate(out)?;
-        write!(out, "{changelog}")?;
+
+        let mut generated = Vec::new();
+        self.generate(&mut generated)?;
+        let generated = std::str::from_utf8(&generated)?;
+        let changelog = changelog.trim_start_matches(|c| c == '\n' || c == '\r');
+        if changelog.is_empty() {
+            write!(out, "{generated}")?;
+        } else {
+            let generated = generated.trim_end_matches(|c| c == '\n' || c == '\r');
+            if generated.is_empty() {
+                write!(out, "{changelog}")?;
+            } else {
+                write!(out, "{generated}\n\n{changelog}")?;
+            }
+        }
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


orhun:
Hey, thanks for the PR!

> Fixes https://github.com/orhun/git-cliff/issues/712

Are you sure this PR is related to that? To me it seems like here we're only adding a trailing newline to the output whereas in the original issue the idea was to add a marker between the release entries to properly create the final changelog.

orhun:
I think the related issue is #1524

jobcespedes:
My bad. It is actually about issue #1524. I'll update the ref

jobcespedes:
I'll discard this PR. I would like a new line not only after --prepend but also between versions. It needs confirmation this is the expected outcome of the PR.

PR #1528, opened 2026-05-21
fix(changelog): preserve blank line between prepended section and existing content

### LINKED ISSUE #1524: No newline between prepended block and old content when using --prepend

SAY-5:
## Description

When `--prepend` is used, the boundary between the newly generated section and the existing changelog can end up with no blank line between them. This happens because the body template often ends without a trailing newline and the existing changelog has its leading newline stripped along with the header by `replacen(header, "", 1)`.

`prepend` now renders the generated content into a temporary buffer, trims trailing newlines from the new section and leading newlines from the existing content, then joins them with exactly one blank line. Empty inputs on either side are handled without introducing spurious newlines.

## Motivation and Context

Fixes #1524. The behaviour and approach follow the discussion on the previously self-withdrawn #1525, but the scope here stays strictly on the `--prepend` boundary as the original report describes; nothing else in changelog generation is touched.

## How Has This Been Tested?

Added two unit tests against `Changelog::prepend`:

- `changelog_prepend_inserts_blank_line_before_existing_content` covers the bug from #1524.
- `changelog_prepend_handles_empty_existing_content` guards against introducing trailing newlines when there is no existing changelog.

Ran:

```
cargo test -p git-cliff-core --no-default-features
cargo +nightly fmt --check
```

All 42 `git-cliff-core` unit tests pass. Pre-existing `clippy` warnings in unrelated code (`changelog.rs:451` / `changelog.rs:474`) are unchanged.

## Types of Changes

## Checklist:

  - [x] `cargo test`

### DIFF
--- a/git-cliff-core/src/changelog.rs
+++ b/git-cliff-core/src/changelog.rs
@@ -640,7 +640,23 @@ impl<'a> Changelog<'a> {
         if let Some(header) = &self.config.changelog.header {
             changelog = changelog.replacen(header, "", 1);
         }
-        self.generate(out)?;
+        let mut generated = Vec::new();
+        self.generate(&mut generated)?;
+        let generated = std::str::from_utf8(&generated)?;
+        write!(out, "{generated}")?;
+        // Both the rendered output and the existing content are written
+        // through unchanged so templates keep full control over their own
+        // whitespace. Only when there is no newline at all at the boundary
+        // (the rendered output does not end with one and the existing
+        // changelog does not start with one) a blank line is inserted to
+        // keep the two sections from running into each other.
+        if !generated.is_empty() &&
+            !changelog.is_empty() &&
+            !generated.ends_with('\n') &&
+            !changelog.starts_with(['\n', '\r'])
+        {
+            write!(out, "\n\n")?;
+        }
         write!(out, "{changelog}")?;
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


npm/git-cliff/yarn.lock
orhun, line 1: Can you revert these changes? Those doesn't seem related to this PR
(going to handle them in `main` later)

git-cliff-core/src/repo.rs
orhun, line 1: Clippy changes also shouldn't live in this PR I think.

git-cliff-core/src/changelog.rs
```diff
@@ -640,8 +641,26 @@ impl<'a> Changelog<'a> {
...
+        if existing.is_empty() {
+            write!(out, "{generated}")?;
+        } else {
+            let new_section = generated.trim_end_matches(['\n', '\r']);
+            if new_section.is_empty() {
+                write!(out, "{existing}")?;
+            } else {
+                write!(out, "{new_section}\n\n{existing}")?;
+            }
+        }
```
orhun, line 659: This feels a bit forced tbh. What if the user wants no newlines as a part of their template? Do we still support such cases?

That is why `git-cliff` previously was just spitting out what it got (i.e. just calling `write!` instead of `writeln!`)

git-cliff-core/src/changelog.rs
```diff
@@ -640,8 +641,26 @@ impl<'a> Changelog<'a> {
...
             changelog = changelog.replacen(header, "", 1);
         }
-        self.generate(out)?;
-        write!(out, "{changelog}")?;
+        // Render the new section first so we can normalise the boundary
+        // between it and the existing changelog. Without this step the new
+        // section and the existing content can end up with no blank line
+        // between them when the body template ends without a trailing
+        // newline and the existing changelog has its leading newline
+        // stripped along with the header (#1524).
```
orhun: We should add an actual link or remove the issue number imo

```suggestion
        // newline and the existing changelog has its leading newline
        // stripped along with the header.
```

orhun (commented):
thanks for the PR!

orhun (approved):
good UX change, thanks a lot!

PR #1545, opened 2026-05-27
feat(changelog): warn when template uses remote variables without matching feature

### LINKED ISSUE #659: Do not fail on unexisting GitHub/GitLab template variables

arieleli01212:
Fixes #659.

## Problem

When a template references variables like `github.contributors` or `commit.gitlab` but the binary was compiled without the corresponding feature flag, the output is silently empty — no error, no warning, just missing data. This is confusing because the template looks correct but produces nothing for those fields.

## Solution

Added a `warn_if_remote_template_variables_without_feature` function that is called from `Changelog::build`. It checks each template (header, body, footer) for remote-specific variable names and emits a `tracing::warn` when the relevant Cargo feature is absent.

The check uses the existing `Template::contains_variable` method and is gated with `#[cfg(not(feature = "..."))]` so there is zero overhead when the feature is enabled.

```
WARN git_cliff_core::changelog: Template uses a `github` variable (e.g. `github.contributors`) 
but the `github` feature is not enabled. Rebuild git-cliff with `--features github` or use a 
binary that includes all remote features.
```

All five remotes are covered: `github`, `gitlab`, `gitea`, `bitbucket`, `azure_devops`.

### DIFF
--- a/git-cliff-core/src/changelog.rs
+++ b/git-cliff-core/src/changelog.rs
@@ -660,29 +660,24 @@ impl<'a> Changelog<'a> {
     }
 }
 
-/// Emits a warning when a template references remote-specific variables
-/// (e.g. `github.contributors`) for a provider whose Cargo feature is not
-/// compiled in.  The render itself will fail later with a confusing
-/// "variable not found in context" message; this warning gives users a
-/// clear, actionable hint before that happens.
+/// Emits a warning when a template references remote-specific variables for a
+/// provider whose Cargo feature is not compiled in.
 fn warn_if_remote_template_variables_without_feature(changelog: &Changelog<'_>) {
-    let templates: Vec<&Template> = [
-        Some(&changelog.body_template),
-        changelog.header_template.as_ref(),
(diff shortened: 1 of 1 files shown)


arieleli01212:
Cleaning up my open PRs — this has been quiet for a while so I'll close it. Let me know if it's worth another look and I'll reopen.

PR #1548, opened 2026-06-01
feat(git): skip commits listed in .git-blame-ignore-revs

### LINKED ISSUE #1084: Ignore .git-blame-ignore-revs hashes

arieleli01212:
Closes #1084.

When a `.git-blame-ignore-revs` file exists at the repository root, git-cliff now automatically skips any commit whose SHA is listed in it.

This follows the same approach as `.cliffignore`: the file is parsed for non-comment, non-empty lines (each line being a commit SHA), and each SHA is added to the skip list as a `CommitParser` with `skip = true`. No configuration is needed — it's always-on when the file is present, matching the "sane defaults" direction discussed in the issue.

The `.git-blame-ignore-revs` format is identical to what `git blame --ignore-revs-file` expects:
- Lines starting with `#` are comments and are ignored
- Blank lines are ignored
- All other lines are treated as commit SHAs

**Changes:**
- `git-cliff-core/src/lib.rs` — add `BLAME_IGNORE_REVS_FILE` constant
- `git-cliff/src/lib.rs` — read the file and extend the skip list, right after the existing `.cliffignore` handling


git-cliff/src/lib.rs
```diff
@@ -786,6 +786,16 @@ pub fn run_with_changelog_modifier<'a>(
...
+            let blame_ignore_file = repository.root_path()?.join(BLAME_IGNORE_REVS_FILE);
+            if blame_ignore_file.exists() {
+                let contents = fs::read_to_string(&blame_ignore_file)?;
+                let commits = contents
+                    .lines()
+                    .filter(|v| !(v.starts_with('#') || v.trim().is_empty()))
+                    .map(|v| String::from(v.trim()))
+                    .collect::<Vec<String>>();
+                skip_list.extend(commits);
+            }
```
orhun, line 798: I think this the essence of this PR and it looks good :)

We can remove the rest of the changes (and possibly update the documentation about this change)

orhun (commented):
Thanks for the PR! It looks good, however it seems like there are some unrelated changes included (in changelog, commit, etc. modules) Can you check?

Issue #1182, opened 2025-06-20
Support searching in .config folder


joshka:
### Is there an existing issue or pull request for this?

- [x] I have searched the existing issues and pull requests

### Feature description

https://dot-config.github.io/ extends the XDG standard to have a .config folder inside your project folder. It would be nice to be able to move the cliff.toml file into a .config folder to clean up the project root folder.

The main benefit of a shorter root folder is that Github displays the README higher on the screen by default.

### Desired solution

Add .config/cliff.toml to the search path when looking for a config file.

### Alternatives considered

None

### Additional context

- bacon has added this already for its config file.
- I don't see this as useful for things like cargo.toml, which is the primary config file for the project, mainly secondary tools

Happy to PR this (it's probably not more than a few lines)

orhun:
Sounds fairly easy to implement:

https://github.com/orhun/git-cliff/blob/871c3c949d9f0cb23600a0f4d16e360d20d32ad7/git-cliff/src/lib.rs#L570-L578

PR welcome!

joshka:
>Sounds fairly easy to implement:

Famous last words... In reality, this uncovers a few different problems (or more things that require a bit of a cleanup if YOLOed).

- The DEFAULT_CONFiG const is used as an index into the embedded templates, for which these are stored as two rust-embed types that effectively do the same thing. (default config and examples). I'd suggest merging these together
- The default in the clap config shouldn't be a default, but should instead be an Option
- The use of clap is pretty unidiomatic for a rust app (e.g. using --init instead of a command), which makes the code that uses the clap stuff fairly verbose and complex.

I'll definitely get to this though.

orhun:
Ah, thanks for looking into it and cleaning up stuff already. I definitely underestimated this :D

Would like to discuss a bit more about the usage of `clap`, but yeah... I agree that we can improve the config/arg handling.

lazizbekravshanov:
Hi @orhun and @joshka — I'd like to work on this if it is still available.

I checked the current issue state and found no assignee or linked/open PR, but I saw @joshka's earlier note that he intended to return to it. @joshka, are you still planning to implement this?

If it is free, my proposed scope is deliberately narrow: distinguish an explicitly supplied `--config` path from automatic discovery, add `.config/cliff.toml` to the discovery path while preserving existing precedence, and cover both discovery and explicit-path behavior with tests. I would avoid broad changes to the `clap`/`--init` interface unless you want that cleanup included.

[...]

PR #1583, opened 2026-07-13
feat(cli): support user-defined templates for --init

lazizbekravshanov:
## Description

Adds support for user-defined `--init` templates, alongside the existing built-in templates:

- `--templates-dir <PATH>` (env: `GIT_CLIFF_TEMPLATES_DIR`) — a directory of user `.toml` templates that `--init <name>` can use. A user template overrides a built-in template of the same name.
- `--list-templates` — prints all available template names (built-in + user-defined), one per line, sorted and deduplicated. This covers the completion use case from the issue.

Behavior details:

- Existing `--init` behavior is unchanged when no `--templates-dir` is given.
- A missing or non-directory `--templates-dir` is a clear error naming the offending path; non-`.toml` files in the directory are ignored.
- `--list-templates` prints and exits (like `--init`), and works outside a git repository.
- Tera body-templates (`--body` / `--template`) are unaffected — this only concerns the `--init` config presets.

## Motivation and Context

Closes #412. Users who keep their own `--init` template presets currently can't use them through git-cliff's built-in mechanism, and there was no way to enumerate the available templates (needed for shell/editor completion).

## How Has This Been Tested?

- **Unit tests** (test-first / TDD):
  - `git-cliff-core` (`embed.rs`): built-in listing (sorted, no extension); merging a user directory (ignoring non-`.toml`, deduped); user template overriding a built-in; `.toml` extension normalization; and a clear error when the directory is missing.
  - `git-cliff` (`args.rs`, `lib.rs`): parsing `--templates-dir` / `--list-templates`; `init_config` writing the user template (and falling back to the built-in).
- **`cargo test`** (workspace): all green.
- **`cargo +nightly fmt --all -- --check`**: clean.
- **`cargo clippy --tests -- -D warnings`**: clean on the changed code. (Note: on my local toolchain — rustc 1.97.0 — clippy also reports one pre-existing `useless_borrows_in_formatting` lint in `git-cliff-core/src/repo.rs`, which is unrelated to this change and outside the diff.)
- **Manual end-to-end** on the built binary: listing built-ins; merging + overriding via `--templates-dir`; the `GIT_CLIFF_TEMPLATES_DIR` env var; `--init <name> --templates-dir` writing user content; the missing-directory error; and confirming both commands work outside a git repository.

## Types of Changes

- [x] New feature (non-breaking change which adds functionality)
- [x] Documentation (no code change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] I have updated the documentation accordingly (if applicable).
[...]

Issue #412, opened 2023-12-27
Add flags and options to support user-defined templates


liuyinz:
**Is your feature request related to a problem? Please describe.**

I define templates by myself and want to init with same mechanism which git-cliff already support by only for built-in templates

**Describe the solution you'd like**

```bash
# add directory which stores user-defined templates, 
git cliff --templates-dir <PATH> (ENV: GIT_CLIFF_TEMPLATES_DIR=)

# print all available templates, both built-in and user-defined templates, which is needed for providing completion in emacs.
 git cliff --list-templates
```
**Describe alternatives you've considered**

no

**Additional context**

- allow user to define his own templates
- give a higher priority to overwrite built-in templates if they have same names

lazizbekravshanov:
Hi @orhun — I'd like to take this on as a follow-up contribution after #1182, if the feature is still desired.

I verified that the issue is open, unassigned, and has no active linked PR. Since the issue predates the current CLI, could you confirm whether the proposed interface is still what you want?

My tentative scope would be:
- add `--templates-dir <PATH>` with `GIT_CLIFF_TEMPLATES_DIR` support;
- add `--list-templates` for built-in and user-provided templates;
- let a user template with the same name override the built-in template;
- add CLI and template-resolution tests plus documentation.

I will wait for scope confirmation before implementing it.

orhun:
Yup, that sounds good 👍🏼

PR #1584, opened 2026-07-13
feat(args)!: discover config when --config is omitted

lazizbekravshanov:
## Description

Replaces the hardcoded `cliff.toml` clap default for `--config` with `config: Option<PathBuf>`, so an explicitly-supplied `--config` is distinguishable from automatic discovery.

- `--config` no longer advertises `[default: cliff.toml]` in `--help`.
- `init_config` is simplified to take an `Option<&Path>` (dropping the `if config_path == DEFAULT_CONFIG { … }` special-case).

**Behavior is intentionally preserved** — this is a type/clarity cleanup, not a behavior change:
- When `--config` is omitted, resolution falls back to the default file name and keeps the existing discovery order (project `cliff.toml` → user config directory → manifest → project discovery → built-in default).
- `.config/cliff.toml` discovery (#1448) still works.
- `--config <path>`, `--config <builtin-name>` (e.g. `detailed`), and `--init [--config <path>]` all behave exactly as before.

## Motivation and Context

Follow-up to the discussion in #1182 — @joshka suggested (and @orhun agreed) that the clap default should be an `Option` rather than a hardcoded value. This is the minimal cleanup for that.

**One open question for you, @orhun:** I kept the resolution order identical for safety, which means with `--config` omitted the implicit `cliff.toml` is still checked *before* the user config directory (unchanged behavior). If you'd prefer the `None` case to route through *pure* discovery instead (which would slightly change the project-`cliff.toml`-vs-user-config precedence), I'm happy to do that in a follow-up — just wanted to avoid changing precedence without your sign-off, since you'd mentioned wanting to discuss the `clap` usage.

## How Has This Been Tested?

- `cargo test` (workspace): all green.
- `cargo +nightly fmt --all -- --check` and `cargo clippy --tests -- -D warnings`: clean on the changed code.
- Manual end-to-end on the built binary: implicit `cliff.toml`, explicit `--config custom.toml`, `.config/cliff.toml` discovery, `--config detailed` (built-in), `--init`, and `--init --config custom.toml` — all behave as before; `--help` no longer shows the default.

## Types of Changes

## Checklist:

- [x] I have formatted the code with rustfmt.
- [x] I checked the lints with clippy.
  - [x] `cargo test`


orhun:
Hey! Yes, I think we should enable discovery when the `--config` option is omitted :)

Also, can you update the docs based on this changes?

lazizbekravshanov:
Done in 329e753 👍

When `--config` is omitted, resolution now goes purely through discovery — a project `cliff.toml` / `.cliff.toml` / `.config/cliff.toml` (searched up the directory tree), then the user config directory — instead of front-loading a hardcoded `cliff.toml`. An explicit `--config <path>` (or a built-in template name) still overrides discovery.

Also updated the docs (Configuration → File Path). Verified behaviorally: implicit `cliff.toml`, `.config/cliff.toml` discovery, explicit `--config custom.toml`, `--config detailed` (built-in), `--init`, and the no-config default. `cargo test`/clippy/fmt all clean. Thanks for the review!

lazizbekravshanov:
Hi @orhun — following up on this one 🙂 I implemented the discovery-when-`--config`-is-omitted behaviour you asked for in `329e753` (project `cliff.toml` / `.cliff.toml` / `.config/cliff.toml` searched up the tree, then the user config dir, then manifest, with an explicit `--config` still overriding), plus the docs update. It's been sitting green since, so just checking whether there's anything else you'd like changed before it can go in — more than happy to adjust.

(The two red checks are infra-only: "Test suite" is the Codecov-upload GPG flake and "Links" is GitLab returning 403s to the CI runner — both fail intermittently on `main` too. The 106 real checks pass.) Thanks!

lazizbekravshanov:
Hi @orhun — no rush at all, just a gentle bump 🙂 The discovery-when-`--config`-is-omitted behaviour you asked for is in (`329e753`), so this PR and #1583 (the `--templates-dir` / `--list-templates` feature for #412 you approved) are both ready whenever you have the bandwidth to review. Completely understand you're busy — happy to rebase, split anything out, or make the review easier however helps. Thanks so much for your time and for maintaining git-cliff!

lazizbekravshanov:
Hi @orhun, gentle follow-up here too. This is the `--config` to `Option` cleanup you asked for, plus the discovery-when-omitted behaviour (in `329e753`): passing no `--config` now runs the normal project and user config discovery, and an explicit `--config` still overrides. Docs are updated and the workspace tests pass. Let me know if there is anything else you want changed.

lazizbekravshanov:
Closing this along with #1583 to clear my queue. The Option cleanup and the discovery behavior you asked for are both done on the branch, so I can reopen and rebase it any time if you want to pick it up later.

orhun:
Hey, sorry for the delay! 

I just had a look at this and realized some breaking things:

- `--workdir` regression: config discovery starts from the current shell directory instead of the specified working directory. Could you look into that?
- Documentation mismatch: an explicit nonexistent --config path still silently falls back to the user config, despite docs saying it overrides discovery.

lazizbekravshanov:
Both good catches, thanks. Fixed in 3d3df5a.

The `--workdir` one was a real regression from making `--config` an Option. Before, `workdir.join(args.config)` always applied because config defaulted to `cliff.toml`; once it became None the discovery path never looked at the workdir. Discovery now starts from `--workdir` when it is set and the current directory otherwise.

On the docs mismatch: the fallback itself is not new, main falls back to the user config the same way when the path does not exist. What was wrong was the sentence I added claiming an unconditional override. The bad case is a nonexistent `--config` when a user config exists, which silently used a different file. It now logs a warning before falling back, and the docs describe the fallback instead of promising an override.

If you would rather a nonexistent explicit `--config` be a hard error instead of a warning, I am happy to change it, that is a behavior change so I did not assume it.

The resolution moved into `resolve_config_path` so it can be tested directly, with five tests covering the explicit, missing, workdir, current directory and user config cases. Workspace tests, clippy and nightly fmt are clean.

orhun:
Found another issue.
Could we resolve the filesystem config only after checking --config-url and built-in configs? Currently --config detailed emits a misleading “file not found, falling back” warning even though the built-in config is used successfully.

orhun (approved):
thx!

lazizbekravshanov:
Good catch, and that one is on my warning from yesterday. Fixed in ed10bfc.

`--config` takes either a path or a built-in name, and the filesystem was being resolved before `--config-url` and the built-ins were ruled out, so naming a built-in printed a missing-file warning right before it loaded successfully. The lookup now happens in the branch that actually uses it, so the warning is only reachable when the filesystem is the real source.

Checked the four paths against a rebuilt binary: `--config detailed` is quiet now, a genuinely missing path still warns, plain discovery works, and the `--workdir` fix is unaffected. Tests, clippy and nightly fmt are clean.

One thing I should be straight about: this one has no unit test. The ordering sits inside the branch chain in `run_with_changelog_modifier`, which I could not exercise without a bigger refactor than felt right to do mid-review, so it is verified by running the CLI. Happy to pull that resolution out into something testable if you would like it covered.

PR #1588, opened 2026-07-19
feat(parser): skip changelog or bump without dropping commits

### LINKED ISSUE #1579: Distinguished behaviour between version bump and changelog

### Feature description

It would be nice if commit types can be considered for the version bump but not for the changelog, and vice versa.
Currently, you can only exclude commit types entirely.

---

**From #1564:**

I only want to include certain commit types in my CHANGELOG.md, like `feat` and `fix`, but exclude less important types like `docs`. So I went with the following config:

```toml
commit_parsers = [
    { message = "^feat", group = "<!-- 0 -->🚀 Features" },
    { message = "^fix", group = "<!-- 1 -->🐛 Bug Fixes" },
]
filter_commits = true
```

Now let's say I need to make a release which only contains one `docs` commit. Then `git-cliff -u --bumped-version` says _there is nothing to bump_. With `filter_commits = false`,  it correctly calculates the next patch version, but the `docs` commit is included in the changelog.

Is there any way to make this possible? It's basically the inverse of #1520. Not sure if the proposed `bump` flag for commit parsers would also work here? If so, I'm happy to open a feature request issue.

### Desired solution

No desired solution in mind. In #1520 the following is suggested:

```toml
{ message = "^doc*", group = "<!-- 3 -->📚 Documentation", bump = true/false },
```

Which looks good to me. My gut feeling is that omitting in the changelog is more common than ignoring it for version bumps, so I would propose `include_in_changelog = true/false` (or both?).

### Alternatives considered

Don't know about any alternatives.

### Additional context

_No response_

arimu1:
## Description

Fold the #1579 changelog-vs-bump split into the existing `commit_parsers` `skip` field instead of adding extra flags.

- `skip = true` — drop the commit from processing (unchanged)
- `skip = "changelog"` — keep the commit for version bump / `filter_commits`, omit it from changelog output
- `skip = "bump"` — keep the commit in the changelog, exclude it from `--bump` / `--bumped-version`

Defaults are unchanged: omitting `skip` includes the commit in both outputs. Existing `skip = true` configs keep working.

## Motivation and Context

Fixes #1579 (follow-up to #1564 / #1520).

Three overlapping booleans (`skip` / `bump` / `include_in_changelog`) made the parser surface confusing. One `skip` control covers the same four outcomes:

| Configuration | Changelog | Version bump |
| --- | ---: | ---: |
| Default | Yes | Yes |
| `skip = true` | No | No |
| `skip = "bump"` | Yes | No |
| `skip = "changelog"` | No | Yes |

Example for a docs-only release with `filter_commits = true`:

```toml
commit_parsers = [
  { message = "^feat", group = "Features" },
  { message = "^fix", group = "Bug Fixes" },
  { message = "^docs", skip = "changelog" },
]
filter_commits = true
```

## How Has This Been Tested?

- `cargo test --workspace` (git-cliff 16; git-cliff-core 116; integration 1; doctests 2)
- Platform: macOS aarch64, Rust stable + nightly rustfmt

New coverage: `parse_commit_skip_targets`, `parse_commit_parser_skip_targets`, `skip_bump_excludes_commit_from_version_increment`.

## Types of Changes

- [ ] Other

## Checklist:

- [x] I checked the lints with [clippy](https://github.com/rust-lang/clippy).
  - [x] `cargo test`


orhun:
Hey, have you tested this locally? I'm not sure what's the different between `skip` and `include_in_changelog` options btw

arimu1:
Yes, tested locally, on a throwaway repo. Config:

```toml
[git]
conventional_commits = true
filter_unconventional = true
tag_pattern = "v[0-9]*"

commit_parsers = [
  { message = "^feat", group = "Features" },
  { message = "^fix", group = "Bug Fixes" },
  { message = "^chore", group = "Chores" },
  { message = "^docs", group = "Documentation", include_in_changelog = false },
]
```

Repo: tagged `v0.1.0`, then two commits — `feat: another feature after tag` and `docs: update usage section`.

`git-cliff --unreleased`:

```
### Features
- another feature after tag

- statistics.commit_count = 2
```

The docs commit is gone from the rendered output, but `statistics.commit_count` is still 2 — it's not removed from the release, just hidden from the changelog body. `--bumped-version` is unaffected (`v0.2.0`, driven by the `feat` commit either way).

On `skip = true` for the same docs commit, `statistics.commit_count` drops to 1 — the commit is dropped from the release entirely, not just the changelog rendering. That's the actual difference: `skip` removes a commit from the pipeline (changelog, statistics, and bump, all at once, unconditionally). `include_in_changelog` and `bump` let you keep the commit in the release but exclude it from one of those two specific outputs, independently.

Concretely, something `skip` can't do today: keep a `feat` commit visible in the changelog but stop it from triggering a minor bump. Same config, `bump = false` on the `feat` parser instead:

```
### Features
- another feature after tag

- statistics.commit_count = 1
```

`--bumped-version` → `v0.1.0` with a "next version is the same as the current version" warning, instead of `v0.2.0`. The commit is still fully present in the changelog and statistics, just excluded from the bump calculation. There's no way to get that with `skip` — `skip` would also erase it from the changelog.

So `include_in_changelog = false` and `bump = false` are inverses of each other (hide from changelog / keep in bump, vs. hide from bump / keep in changelog), and both are strictly weaker than `skip`, which removes the commit everywhere at once. I don't think they're redundant with `skip`, but happy to rename `include_in_changelog` to something less similar-sounding if that's the concern (e.g. `changelog` to mirror `bump`).

Also ran the full test suite (`cargo test --workspace`), including the new `parse_commit_inclusion_flags` test — all pass. The two red CI checks (Lints / Test suite) are pre-existing: Lints fails on the same `clippy::useless_borrows_in_formatting` in `repo.rs:621` on current `main` (untouched by this PR), and Test suite's only failure is the Codecov CLI step (`gpg: Can't check signature: No public key`) — all actual Rust tests in that job pass. Confirmed both are red on `main` too.

arimu1:
@orhun thanks — yes, the parser/config paths were exercised via the `git-cliff-core` integration tests in this PR (bump + `include_in_changelog` fixtures).

### `skip` vs `include_in_changelog`
They operate at different layers:

- **`skip`** (commit parser / `CommitParserConfig`) — drops the commit from processing entirely (never becomes a changelog entry). Existing behavior.
- **`include_in_changelog`** (new, on the *parsed commit / bump path*) — keeps the commit available for **version bump calculation** while optionally excluding it from the rendered changelog body. That is the gap #1579 hits: “count this commit for semver bump, but do not print it.”

So `skip=true` is stronger (invisible to bump + changelog). `include_in_changelog=false` is the narrower “bump-only / hide from body” knob.

CI “Test suite” / “Links” failures on the last run look infrastructure-side (Codecov GPG verify; lychee 403s on gitlab.com sign-in redirects) rather than assertion failures in the new tests — happy to re-trigger or adjust if you want anything else covered locally.

arimu1:
@orhun Re-triggered CI. The two red checks look environmental rather than from this change:

- **Test suite**: codecov action fails on GPG signature verify for the uploader binary (`Can't check signature: No public key`), not cargo tests.
- **Links**: lychee reports **403** on `gitlab.com/.../commit/...` URLs (redirect to sign-in). Pre-existing website examples, not docs introduced here.

Happy to wait on infra recovery or adjust if you prefer another approach.

orhun:
Hey, I just had another look at this and I think it's very confusing to have 3 options/controls for commit parsers.

Consider this:

| Configuration                  | Changelog | Version bump |
| ------------------------------ | --------: | -----------: |
| Default                        |       Yes |          Yes |
| `skip = true`                  |        No |           No |
| `bump = false`                 |       Yes |           No |
| `include_in_changelog = false` |        No |          Yes |

There are some overlapping controls....
Can we try to find a way to cleanly solve #1579 without compromising UX?

arimu1:
Folded the extra flags into `skip` so there is one control instead of three overlapping booleans.

| Configuration | Changelog | Version bump | In release |
| --- | ---: | ---: | ---: |
| (unset) / `false` | Yes | Yes | Yes |
| `skip = true` | No | No | No |
| `skip = "bump"` | Yes | No | Yes |
| `skip = "changelog"` | No | Yes | Yes |

`skip = true` is unchanged (hard drop). `"changelog"` and `"bump"` keep the commit in the pipeline for the other output and for `filter_commits`. Existing `skip = true` configs keep working.

Rebased onto current `main`.

orhun:
Oh thanks, this made everything much clear. But I'm still not convinced that this feature is usable, in which case did you need it specifically?

Issue #123, opened 2022-11-03
GIT_CLIFF_INCLUDE_PATH not working for multiple values


chamini2:
**Describe the bug**

**To Reproduce**
1. clone https://github.com/fal-ai/fal/commit/cb785c00746fcbb4e9d401ebc1e3da6e17759436
2. GIT_CLIFF_INCLUDE_PATH="adapter/*"
3. git cliff shows output
4. GIT_CLIFF_INCLUDE_PATH=".github/*"
5. git cliff shows output
6. GIT_CLIFF_INCLUDE_PATH="adapter/* .github/*"
7. git cliff shows NO output

**Expected behavior**
I expected git cliff to generate changelogs for commits touching ANY of the 2 directories

chamini2:
I cannot use the --include-path flag becuase I cannot get it working: look at https://github.com/orhun/git-cliff-action/pull/3

orhun:
Hmm, have you tried using a comma-separated list for `GIT_CLIFF_INCLUDE_PATH`?

orhun:
I guess this is solved in `git-cliff-action`. Let me know if you're still having problems!

sam-stenvall-iceye:
This doesn't seem to have been resolved. I've tried both space-separated and comma-separated glob patterns, but any combination of two patterns results in an empty change log being generated.

ognis1205:
Thanks for the report @sam-stenvall-iceye !

I confirmed this is still an issue. I believe this is caused by the fact that we don't properly configure the clap env variable delimiter, so multiple values from GIT_CLIFF_INCLUDE_PATH are not parsed as expected.

cc: @orhun

sam-stenvall-iceye:
What worked in the end for me was passing `--include-path` multiple times, e.g. `--include-path 'foo/**/*' --include-path bar.txt`

sam-stenvall-iceye:
@ognis1205 so with the latest fixes, both `--include-path 'foo/*' --include-path 'bar/*'` and `--include-path 'foo/* bar/*'` should work?

ognis1205:
hey @sam-stenvall-iceye ,

yes, I've tested all three approaches:

```sh
export GIT_CLIFF_INCLUDE_PATH="website/**/* .github/**/*"
cargo run
```

```sh
unset GIT_CLIFF_INCLUDE_PATH
cargo run -- --include-path 'website/**/* .github/**/*'
```

```sh
unset GIT_CLIFF_INCLUDE_PATH
[...]

PR #1592, opened 2026-07-24
fix(args)!: stop greedy multi-value options from swallowing the positional range

### LINKED ISSUE #1591: include_path silently swallows a trailing positional RANGE

### Description of the bug

`--include-path` and `--exclude-path` are declared `num_args(1..)` so they can take space-separated patterns. clap reads that greedily, so when a positional `RANGE` follows the flag, clap consumes the range as one more glob pattern. `opt.range` comes back `None`, git-cliff walks the whole history, path-filters it, and emits a changelog.

Nothing warns. The output is a well-formed changelog containing every commit that ever touched the path, which reads as "the range was ignored" if you notice at all. In a monorepo, where this flag combination is the whole point, the parse misses the stated intent and returns a wrong result that is plausible enough to ship: it reaches published release notes before anyone catches it.

This reproduces on released git-cliff (v2.13.1 and every version back to when the flags gained multi-value parsing).

### Steps To Reproduce

Any repository works; the pattern is what matters.

```bash
# 1. The range comes after --include-path: swallowed.
git cliff --include-path 'pkg/**' pkg/v1.0.0..pkg/v1.1.0

# 2. Same intent, range first: correct.
git cliff pkg/v1.0.0..pkg/v1.1.0 --include-path 'pkg/**'
```

Invocation 1 emits the full path-filtered history. Invocation 2 emits the range.

A parser-level test pins it with no repository involved: `opt.range` is `None` and `opt.include_path` holds two patterns, the second being the literal string `pkg/v1.0.0..pkg/v1.1.0`.

`--` before the range (`--include-path 'pkg/**' -- pkg/v1.0.0..pkg/v1.1.0`) also parses correctly, which is a workaround but not a fix: nothing tells you that you needed it.

### Expected behavior

Both invocations mean the same thing to the person who typed them, so they should parse to the same intent: a trailing positional `RANGE` belongs in the `RANGE` slot regardless of which options precede it.

### Screenshots / Logs

_No response_

### Software information

- Operating system: Linux (reproduces regardless)
- Rust version: 1.85+
- Project version: 2.13.1, and every release since `--include-path` gained multi-value parsing

### Additional context

The same greedy `num_args(1..)` declaration is on `--with-commit` and `--skip-commit`, which drop a trailing range the same way, so the defect is a small class rather than a single flag. `--repository` shares the declaration but rejects a range as a path and errors loudly, so it is not affected.

Surfaced by @leoweyr in review of #1489 ([comment](https://github.com/orhun/git-cliff/pull/1489#issuecomment-4708196005)), asking whether range selection composes with `--include-path` in monorepo setups.

cds-amal:
Declares the greedy multi-value options as one value per occurrence (`num_args(1)`), so a trailing positional `RANGE` lands in the `RANGE` slot instead of being absorbed as one more value.

Two commits, no new surface:

1. **`--include-path` / `--exclude-path`** (keeping `value_delimiter(' ')`): the flags the bug was reported against, where combining a range with a path filter is a documented monorepo workflow. Six parser-level regression tests and one extra CI fixture entry.
2. **`--with-commit` / `--skip-commit`**: the same defect, found by inspection rather than by report. Four more tests.

The split is deliberate. If the wider scope is unwanted, reverting the second commit leaves the reported fix intact and green; nothing in the first commit depends on it.

## Motivation and Context

Fixes #1591.

`num_args(1..)` makes clap greedy, so `git cliff --include-path 'pkg/**' pkg/v1.0.0..pkg/v1.1.0` parses the range as a glob. `opt.range` is `None`, git-cliff walks all of history, path-filters it, and emits a changelog. Nothing warns. The parse misses the stated intent and returns a wrong result that is plausible enough to ship: in a monorepo, where this flag combination is the whole point, it reaches published release notes before anyone catches it.

The reordered invocation (`git cliff pkg/v1.0.0..pkg/v1.1.0 --include-path 'pkg/**'`) parses correctly, which is what makes the defect hard to see: the same arguments in a different order mean different things to clap while meaning the same thing to the person who typed them. Call that property intent-idempotence, the guarantee that reordering arguments never changes the intent they express; greedy multi-value options break it, and they break it silently, because a glob pattern and a revision range are both just strings to clap.

Restoring that guarantee is worth a breaking parse change. Every other class of range mistake in git-cliff fails loudly at revparse; this one doesn't fail at all.

Surfaced by @leoweyr in review of #1489 ([comment](https://github.com/orhun/git-cliff/pull/1489#issuecomment-4708196005)). Split out of that PR deliberately: the compatibility trade-off below deserves its own decision, on its own diff, rather than riding along inside a feature.

### The same declaration appears on three more options

`--with-commit` and `--skip-commit` are also `num_args(1..)` and swallow a trailing range identically. Verified on this repository, before the second commit:

```
$ git cliff --with-commit "feat: synthetic" v2.12.0..v2.13.0
...whole history

$ git cliff v2.12.0..v2.13.0 --with-commit "feat: synthetic"
...the range
```

Both fail silently, for the same reason as the path flags: any string is a plausible commit message or SHA, so nothing rejects the range once it lands in the wrong slot. After the second commit, both orderings produce byte-identical output.

`--repository` shares the declaration but not the defect, so it is untouched: its `value_parser = Opt::parse_dir` rejects a range as a path and errors loudly rather than silently.

### The compatibility trade-off

For the path flags, every documented multi-pattern form still works, and each already has a CI fixture pinning it, from the work on #123, the `value_delimiter` fix in #1450, and the fixture set #1455:

- repeated flags: `test-include-path-repeated-flag`
- a single quoted space-separated value: `test-include-path-space-separated`
- the config keys: `test-include-path-config`
- the env vars (which route through `value_delimiter`, not `num_args`): `test-include-path-env-var`

All four stay green here, unmodified, so the multi-value behavior #123 asked for and #1450 delivered is provably intact rather than merely believed intact.

The second commit's surface is sharper, and worth deciding on separately. Neither `--with-commit` nor `--skip-commit` carries a `value_delimiter`, and neither should: `--with-commit "8f55e69… feat: add X"` is a documented form (`website/docs/usage/adding-commits.md`) whose single value contains spaces, so splitting on whitespace would corrupt it. `cli_with_commit_keeps_whitespace_inside_one_value` pins that. The consequence is that repeated flags become the *only* multi-value form for those two, where the path flags keep both forms. (A delimiter would be safe on `--skip-commit`, since SHA1s contain no spaces, but that is new behavior rather than a fix, so it stays out of scope.)

The one form that changes for the path flags is unquoted multiple tokens:

```bash
# BREAKS: 'docs/**' now lands in the positional RANGE slot and fails at
# revparse (or, worse luck, silently becomes the range if it happens to
# name a real revision).
git cliff --include-path 'website/**' 'docs/**'

# Migration 1, valid in every release since the flags debuted:
git cliff --include-path 'website/**' --include-path 'docs/**'

# Migration 2, v2.13.0 or newer: one quoted space-separated value.
# Relies on value_delimiter (#1450); on older releases this parses as a
# single literal pattern containing a space.
git cliff --include-path 'website/** docs/**'
```

That form appears in no docs, fixtures, or tests, though the rendered help (`--include-path <PATTERN>...`, now `<PATTERN>`) did advertise it. Hence `fix!` with a `BREAKING CHANGE:` footer.

## How Has This Been Tested?

Ten parser-level tests in `args.rs`, no repository required.

First commit, the path flags:

- `cli_include_path_does_not_swallow_trailing_positional_range` and its `--exclude-path` twin: the reported invocation now parses to the right intent.
- `cli_include_path_multiple_patterns_still_parse`: repeated flags and the quoted space-delimited value both yield two patterns.
- `cli_include_path_unquoted_second_token_falls_to_range_slot`: characterizes the trade-off above, so the compatibility change is visible in the test suite rather than only in this description.
- `cli_positional_range_before_include_path_is_not_swallowed` and `cli_double_dash_protects_trailing_positional_range`: the two orderings that already worked keep working.

Second commit, the commit flags:

- `cli_with_commit_does_not_swallow_trailing_positional_range` and `cli_skip_commit_does_not_swallow_trailing_positional_range`: the same defect, the same fix.
- `cli_with_commit_keeps_whitespace_inside_one_value`: the documented `"<sha> <message>"` form stays a single value, which is why neither flag gets a `value_delimiter`.
- `cli_repeated_flags_remain_the_multi_value_form`: repeated occurrences still accumulate for both.

The pre-existing `path_env_vars_are_split_into_multiple_patterns` test covers the env-var path and is unchanged, which is the evidence that `value_delimiter` still carries the multi-pattern case.

End to end, the CI fixture matrix now runs `test-monorepo-include-path` twice against one `expected.md`, with the range before and after `--include-path`. The path-first ordering is exactly the invocation that used to lose its range, so the two runs must now agree.

- [ ] `cargo test --workspace` passes at each commit, not just at the tip (`repo::test::git_upstream_remote` needs a branch with an upstream configured)
- [ ] `cargo test --workspace` passes with the second commit reverted, confirming the split is real
- [ ] `cargo +nightly fmt --all --check` clean
- [ ] `cargo clippy --tests -- -D warnings`: no new warnings from `main`
- [ ] CI fixture matrix: `test-monorepo-include-path` green in both argument orders
- [ ] CI fixture matrix: the four `test-include-path-*` fixtures from #1455 unchanged and green

Before and after, using `--dry-run` from #1489 to make the parse visible (the fix itself does not depend on that branch):

```
# before
$ git cliff --include-path 'pkg/**' pkg/v1.0.0..pkg/v1.1.0 --dry-run
range:    [<first-commit>, HEAD]        <-- range swallowed as a glob

# after
$ git cliff --include-path 'pkg/**' pkg/v1.0.0..pkg/v1.1.0 --dry-run
range:    (pkg/v1.0.0, pkg/v1.1.0]
```

## Types of Changes

## Checklist:

---

**Note for reviewers:** the second commit is separable by design. `git revert <sha>` on this branch applies cleanly and leaves a green tree, so narrowing the scope to the reported flags costs one command rather than a rewrite.

#1489 (range-endpoint flags) is stacked on this branch. Until this merges, #1489's diff includes this commit as an ancestor; once it lands, that diff shrinks to the feature alone.


orhun (approved):
Hey, thanks!

I'll merge this in #1614 since maintainer edits are disabled in this one.

PR #1593, opened 2026-07-27
fix(git): match directory include/exclude patterns on Windows

### LINKED ISSUE #1589: Windows: directory include/exclude patterns silently don't match (+ 2 flaky test assertions)
### Description of the bug

Ran `cargo test` on Windows and got 3 failures in `git-cliff-core`. Two are test-assertion bugs, one is a real bug affecting Windows users.

**The real one**: `normalize_pattern` in `repo.rs` checks `pattern.as_str().ends_with(path::MAIN_SEPARATOR)` to decide whether to append `**` to a directory pattern like `"docs/"`. `MAIN_SEPARATOR` is `\` on Windows, but patterns (and git's internal paths) always use `/`. So on Windows, `include = ["docs/"]` in `cliff.toml` never gets the `**` appended, and never matches anything inside `docs/`. Confirmed directly: `Pattern::new("dir/").matches_path("dir/initial.txt")` is `false`, `Pattern::new("dir/**").matches_path(...)` is `true`.

**The two flaky ones**: `propagate_error_if_no_repo_exist`/`_found` compare `format!("{error:?}")` against a plain `path.display()` string. On Windows the Debug-formatted error escapes backslashes (`\\`) while `path.display()` doesn't, so the substring check always fails even though the error message is correct. `error.to_string()` (which goes through the `thiserror`-generated `Display`) doesn't have this problem.

### Steps To Reproduce

1. `git fetch --tags`, `cargo test -p git-cliff-core` on Windows
2. `repo::test::test_should_retain_commit`, `repo::test::propagate_error_if_no_repo_exist`, `repo::test::propagate_error_if_no_repo_found` fail

### Expected behavior

Tests pass, and directory-style include/exclude patterns work the same on Windows as elsewhere.

### Software information

- Operating system: Windows 11
- Rust version: rustc 1.99.0-nightly / stable toolchain per `rust-toolchain.toml`

---

I have fixes for both (one-line separator fix + swap `{error:?}` for `.to_string()` in the two tests), checked against #1379 - no overlap, different file. Happy to open a PR.

Cyrus580529:
## Description

`normalize_pattern` decided whether a glob pattern refers to a directory by testing `pattern.as_str().ends_with(path::MAIN_SEPARATOR)`. Glob patterns and the paths git reports in a diff always use `/`, on every platform, but `MAIN_SEPARATOR` is `\` on Windows, so the test never matched there and the trailing `**` was never appended.

The effect is that `include`/`exclude` entries written as directories, like `docs/`, silently match nothing on Windows. No error, just an empty result. On Unix nothing changes, since `MAIN_SEPARATOR` is already `/`.

The second commit is separate and changes no behaviour. `propagate_error_if_no_repo_exist` and `propagate_error_if_no_repo_found` compared `format!("{error:?}")` against `path.display()`. Debug escapes backslashes and Display doesn't, so on Windows the assertion compared `\\` against `\`. They now go through `error.to_string()`.

## Motivation and Context

Found by running `cargo test -p git-cliff-core` on Windows, where three tests fail on a clean checkout. `test_should_retain_commit` already covers the `dir/` case, but it only fails on Windows, and the `test` job runs on `ubuntu-22.04` (Windows appears only in `cd.yml`, which builds release binaries and runs no tests), so this has never shown up in CI.

The new `test_normalize_pattern` asserts the output directly. It fails on Windows before this change and passes after; on Linux both versions behave identically, since `MAIN_SEPARATOR` is `/` there.

Fixes #1589.

## How Has This Been Tested?

Windows 11, stable Rust 1.96.1 for tests, nightly for rustfmt.

- `cargo test` — `git-cliff-core` goes from 66 passed / 3 failed to 69 passed / 0 failed; the rest of the workspace is unchanged.
- `cargo clippy --tests --verbose -- -D warnings` — nothing in `repo.rs`. The findings in `command.rs` and `changelog.rs` are identical on a clean `main`; `command.rs:57` only appears on Windows, because the one test in that module is `#[cfg(target_family = "unix")]` and `use super::*` is then unused.
- `cargo +nightly fmt --all -- --check` — clean.

`git_upstream_remote` reads the tracking config of the branch the working copy is on, so it fails on a local branch with no upstream yet and passes once pushed.

N/A

## Types of Changes

- [ ] Other

## Checklist:

  - [x] `cargo test`

### DIFF
--- a/git-cliff-core/src/repo.rs
+++ b/git-cliff-core/src/repo.rs
@@ -1,5 +1,5 @@
 use std::io;
-use std::path::{self, Path, PathBuf};
+use std::path::{Path, PathBuf};
 use std::result::Result as StdResult;
 use std::sync::LazyLock;
 
@@ -330,7 +330,9 @@ impl Repository {
     /// It removes the leading `./` and adds `**` to the end if the pattern is a
     /// directory.
     fn normalize_pattern(pattern: Pattern) -> Pattern {
-        let star_added = if pattern.as_str().ends_with(path::MAIN_SEPARATOR) {
+        // glob patterns and git's diff paths always use '/', whatever the host
+        // OS, so this must not be `path::MAIN_SEPARATOR`
+        let star_added = if pattern.as_str().ends_with('/') {
             Pattern::new(&format!("{pattern}**")).expect("failed to add '**' to the end of glob")
         } else {
             pattern
@@ -1191,6 +1193,20 @@ mod test {
             before,
             "no .git-blame-ignore-revs file present"
         );
+    }
+
+    #[test]
+    fn test_normalize_pattern() {
+        let normalize = |input: &str| {
+            Repository::normalize_pattern(Pattern::new(input).expect("valid pattern"))
+                .as_str()
+                .to_string()
+        };
+
+        assert_eq!(normalize("dir/"), "dir/**");
+        assert_eq!(normalize("./dir/"), "dir/**");
+        assert_eq!(normalize("./file.txt"), "file.txt");
+        assert_eq!(normalize("dir/file.txt"), "dir/file.txt");
     }
 
     #[test]
--- a/git-cliff-core/src/repo.rs
+++ b/git-cliff-core/src/repo.rs
@@ -1031,7 +1031,7 @@ mod test {
         assert!(result.is_err());
         if let Err(error) = result {
             assert!(
-                format!("{error:?}").contains(
+                error.to_string().contains(
                     format!("could not find repository at '{}'", path.display()).as_str()
                 )
             );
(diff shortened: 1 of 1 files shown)


Cyrus580529:
The three red checks here are already red on `main` — run 30181888210 from 2026-07-26 fails the same three, and the code they point at is untouched by this branch.

`Lints` is clippy 1.97's `useless_borrows_in_formatting` in `upstream_remote`, reported at `repo.rs:621` on `main` and 623 here only because this branch adds two comment lines above it. I sent that one separately as #1594. In `Test suite` the `Run tests` steps pass and only the codecov upload fails on its GPG check, and `Links` is lychee on the website.

orhun (approved):
good stuff!

PR #1600, opened 2026-08-07
fix(changelog): fall through to the next parser when a commit field is missing

### LINKED ISSUE #1598: local commit gets skipped by commit parsers if looking for label field

### Description of the bug

If I have commit parsers that first try to group by label, with then a fallback to catch any unmatched commit, local commits that aren't on the remote (and thus not from a PR) error early and are skipped due to this error, meaning they don't fall to the catch all.

### Steps To Reproduce

Example config:

```toml
conventional_commits = false
filter_unconventional = false
require_conventional = false

commit_parsers = [
    { field = "remote.pr_labels", pattern = "bug", group = "Bug fixes" },
    { message = ".*", group = "Miscellaneous" },
]
```

Then ensure have a local commit on branch that isn't on remote, and run

```sh
git-cliff --tag tag_name --unreleased --output EXAMPLE_CHANGELOG.md
```

Will get error log like

```sh
WARN  git_cliff_core::changelog > process_commits: 2 commit(s) were skipped due to field error(s) (run with `-vv` for details)
```

In verbose mode:

```sh
TRACE git_cliff_core::process                   > process_commits: 3df894a - Field error: `field 'remote.pr_labels' is missing or has unsupported type (expected a String, Number, Bool, or Null — or an Array of these scalar values)` (test456)
TRACE git_cliff_core::process                   > process_commits: 54d785b - Field error: `field 'remote.pr_labels' is missing or has unsupported type (expected a String, Number, Bool, or Null — or an Array of these scalar values)` (test123)
WARN  git_cliff_core::changelog                 > process_commits: 2 commit(s) were skipped due to field error(s) (run with `-vv` for details)
```

### Expected behavior

It would be nice if they fell through to the other commit parsers instead of being skipped entirely, since later parsers might not rely on this field being present.

### Screenshots / Logs

_No response_

### Software information

- Operating system: macOS Tahoe 26.5.2
- Rust version: 1.97
- Project version: installed from git, 122642a0e25c72562bba9b5d59c443c8a5547d46

### Additional context

I don't strictly have a use case for this, its just something I noticed whilst testing this tool out and it seemed like a non-obvious behaviour

YuriNachos:
Closes #1598.

## Problem

When a commit parser references a field that is genuinely **absent** on a commit (e.g. `remote.pr_labels` on a local / non-PR commit where `remote` is `None`), the parser returned `FieldError` and aborted the whole parse — so a single field-based parser could prevent any later (catch-all) parser from grouping the commit. Reported in #1598: a local commit gets skipped by the commit parsers if one of them looks for a field that does not exist on it.

## Fix

Distinguish a **missing** field from a **present-but-unsupported-type** field:

- the field resolves to `None` because the path is absent → skip only this field criterion and let the rest of the parser (and the remaining parsers) decide;
- the field is present but has an unsupported type (e.g. a bare `Value::Object`) → still error, exactly as before.

So a local commit no longer aborts the chain when a parser looks for `remote.pr_labels`; the catch-all parser can still group it.

## Test

`parse_commit_missing_field_falls_through_to_next_parser` — a commit with `remote = None` and two parsers (field `remote.pr_labels`, then a catch-all) now groups via the catch-all instead of erroring. The existing `parse_commit_field` (present-but-`Object` → error) still passes, preserving that contract.

`cargo test -p git-cliff-core commit::test` → 9 passed; `cargo clippy --tests -D warnings` clean.

### DIFF
--- a/git-cliff-core/src/commit.rs
+++ b/git-cliff-core/src/commit.rs
@@ -351,10 +351,14 @@ impl Commit<'_> {
             if let (Some(field_name), Some(pattern_regex)) =
                 (parser.field.as_ref(), parser.pattern.as_ref())
             {
+                let field_present;
                 let values = if field_name == "body" {
+                    field_present = true;
                     vec![body.clone()].into_iter().collect()
                 } else {
-                    tera::dotted_pointer(&lookup_context, field_name).and_then(|v| match v {
+                    let field_value = tera::dotted_pointer(&lookup_context, field_name);
+                    field_present = field_value.is_some();
+                    field_value.and_then(|v| match v {
                         Value::String(s) => Some(vec![s.clone()]),
                         Value::Number(_) | Value::Bool(_) | Value::Null => {
                             Some(vec![v.to_string()])
@@ -386,10 +390,19 @@ impl Commit<'_> {
                         }
                     }
                     None => {
-                        return Err(AppError::FieldError(format!(
-                            "field '{field_name}' is missing or has unsupported type (expected a \
-                             String, Number, Bool, or Null — or an Array of these scalar values)",
-                        )));
+                        if field_present {
+                            return Err(AppError::FieldError(format!(
+                                "field '{field_name}' is missing or has unsupported type \
+                                 (expected a String, Number, Bool, or Null — or an Array of these \
+                                 scalar values)",
+                            )));
+                        }
+                        // The field is genuinely absent on this commit (e.g.
+                        // `remote.pr_labels` when `remote` is `None`), so this parser
+                        // cannot decide it: skip only this field criterion and let the
+                        // rest of the parser (and the remaining parsers) decide. A
+                        // present-but-unsupported `Value::Object` field still errors
+                        // above. See orhun/git-cliff#1598.
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


YuriNachos:
Re-triggered CI. For the record, the red `Test suite` job is not a test failure:

```
==> Verifying GPG signature integrity
gpg: Signature made Thu Jul  9 01:37:57 2026 UTC
gpg:                using RSA key 27034E7FDB850E0BBC2C62FF806BB28AED779869
gpg: Can't check signature: No public key
==> Could not verify signature. Please contact Codecov if problem continues
```

Every `cargo test` step in that job reported `test result: ok`; the job only fails at the `codecov/codecov-action` upload step, which cannot fetch the Codecov public key. The same job is red on `main` for the same reason, so it is not specific to this branch. All the fixture jobs, `Check`, `Lints`, `Formatting`, `Doctests` and `CodeQL` are green.

YuriNachos:
The "Test suite" check is red only on the "Upload reports to codecov" step (codecov-action in TOKENLESS fork-upload mode, exited 1 with CC_FAIL_ON_ERROR). The actual Rust test run is green — 0 failures across the suite; Check / Lints / Doctests / Test fixtures all pass. This is an intermittent codecov tokenless-upload infra flake unrelated to the parser-fallthrough change. Rerunning the job (maintainer) or a push re-trigger should clear it.

YuriNachos:
A quick check-in on this one: the fall-through behaviour matches what #1598 described — a local commit with `remote = None` now skips just the `remote.pr_labels` criterion instead of aborting the parser chain, and the catch-all grouping test covers it. Is anything else needed from my side — extra fixture coverage, a rebase onto main, or changes to how the missing-field case is documented? For clarity, the red `Test suite` job is the codecov tokenless-upload step I detailed in my earlier comment; the Rust tests themselves pass. Happy to adjust whatever makes this easiest to land.

orhun (approved):
thanks!

PR #1602, opened 2026-08-09
Preserve releases sandwiched between non-consecutive skip_tags matches

YuriNachos:
## Summary
A release sandwiched between two non-consecutive `git.skip_tags` matches is no longer dropped from the changelog. More generally, the chain of any release used as a repoint target is preserved.

## Root cause
When `skip_tags` matched a release, the code used `get_mut` to repoint the skipped release at the next-older kept release, then set `previous = None` on that shared older release **in place**. For a release sitting between two non-consecutive skip matches, this corrupted the shared release's own `previous` chain, dropping it. The same flaw nullled the chain of *any* release used as a repoint target, which also affected downstream readers of `release.previous` (statistics `days_passed_since_last_release`, `calculate_next_version`).

## Changes
- `git-cliff-core/src/changelog.rs` — clone the previous release (`get(...).cloned()`) so setting `previous = None` on the detached snapshot leaves the shared older release untouched. Same single clone as before, no perf change.
- Inline table test (`changelog_generator_skip_tags_keeps_sandwiched_release`, 5 cases) — proves red→green for the sandwich case and the single-skip-with-kept-target case; consecutive/no-match cases are regression guards.

## Test plan
- [x] Inline `cargo test` cases — two fail on the old `get_mut` code (chain corrupted to `None`), pass after the clone fix.

This change was developed with AI assistance (Claude Code); every changed line was reviewed and understood.

### DIFF
--- a/git-cliff-core/src/changelog.rs
+++ b/git-cliff-core/src/changelog.rs
@@ -196,10 +196,18 @@ impl<'a> Changelog<'a> {
                     .and_then(|release| release.version.as_ref()) ==
                     Some(skipped_tag)
             }) {
-                if let Some(previous_release) = self.releases.get_mut(release_index + 1) {
+                if let Some(mut previous_release) = self.releases.get(release_index + 1).cloned() {
+                    // Repoint this release at the next-older kept release.
+                    //
+                    // The clone is detached (`previous = None`) so the snapshot
+                    // stored on this release does not carry a stale chain, but
+                    // the shared older release is left untouched. Mutating the
+                    // shared release here used to null its own `previous`, which
+                    // broke the chain of a release sandwiched between two
+                    // non-consecutive skip matches (and, more generally, dropped
+                    // the chain of any release used as a repoint target).
                     previous_release.previous = None;
-                    self.releases[release_index].previous =
-                        Some(Box::new(previous_release.clone()));
+                    self.releases[release_index].previous = Some(Box::new(previous_release));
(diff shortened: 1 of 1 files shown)


YuriNachos:
Note on the red checks, so they don't read as a problem with this branch:

**`Links`** — the failure is a broken link in `website/docs/installation/macports.md`:

```
🚫 Errors      | 1
### Errors in website/docs/installation/macports.md
```

This PR touches exactly one file, `git-cliff-core/src/changelog.rs`. It adds no links and no docs. The same workflow is red on `main` itself — the three most recent `main` runs (2026-08-09, 2026-08-04, 2026-08-02) all conclude `failure`.

**`Test suite`** — same story as #1600: the Rust tests themselves pass; the job exits non-zero on the codecov upload step, which cannot authenticate from a fork.

Neither is fixable from this branch. Happy to open a separate PR for the macports link if that's useful.

YuriNachos:
Wanted to check back on this one — the red `Links` check is a pre-existing website issue (details in my comment above), and the rest of the CI is green. Is there anything you'd like changed in the implementation or the tests before it can be reviewed?

orhun:
Could you add a concrete before/after example showing which release is lost or incorrectly linked? Adding a [test fixture](https://github.com/orhun/git-cliff/tree/main/.github/fixtures) usually helps with that.

Also:
- statistics are calculated before the previous chain is repaired. Should we instead calculate them after finalizing the release chain?
- can we fix this in the statistics calculation function instead?

PR #1605, opened 2026-08-12
fix(bump): evaluate increment regexes against the commit subject only

### LINKED ISSUE #1476: custom_major/minor_increment_regex ignored for multiline commits when conventional_commits = false

### Description of the bug

Summary                                                                                                                
                  
  When conventional_commits = false and a multiline commit is passed via --with-commit, git-cliff silently falls back to 
  a patch bump even when the commit subject matches custom_major_increment_regex or custom_minor_increment_regex. The
  bump rules are completely ignored.
                                                                                                                         
  ---
  Configuration (cliff.toml)                                                                                             
                  
  [git]
  conventional_commits = false                
                                          
  commit_parsers = [
    { message = "^breaking", group = "🚀 🚀 Breaking Changes" },                                                         
    { message = "^feat",     group = "✨ Features" },
    { message = "^fix",      group = "🐛 Bug Fixes" },                                                                   
    { message = "^other",    group = "📦 other", skip = true }
  ]                                                                                                                      
                                                                                                                         
  [bump]                                                                                                                 
  custom_major_increment_regex = "^breaking"                                                                             
                                              
  [changelog]                                                                                                            
  trim = false    
  body = """                                                                                                             
  ...                                         
  """                                                                                                                    
                  
  I have intentionally disabled conventional_commits because my commit messages do not follow the type(scope):           
  description structure. They are free-form strings that start with a known prefix (breaking, feat, fix). The
  custom_*_increment_regex fields exist precisely to support this use case.                                              
                                                                                                                         
  ---
  Reproduction                                                                                                           
                  
  # Single-line commit — works correctly, bumps MAJOR
  git cliff --config cliff.toml \             
    --tag-pattern '^nati-[0-9]+\.[0-9]+\.[0-9]+$' \
    --bump --bumped-version \                 
    --with-commit 'breaking: remove v1 endpoints' \                                                                      
    -- HEAD..HEAD                         
  # → nati-2.0.0  ✓                                                                                                      
                                                                                                                         
  # Multiline commit (subject + body, no blank-line separator) — WRONG, bumps PATCH                                      
  git cliff --config cliff.toml \                                                                                        
    --tag-pattern '^nati-[0-9]+\.[0-9]+\.[0-9]+$' \
    --bump --bumped-version \                                                                                            
    --with-commit $'breaking: remove v1 endpoints\nThis removes the entire /v1 API surface.' \
    -- HEAD..HEAD                                                                                                        
  # → nati-1.0.1  ✗  (expected nati-2.0.0)
                                                                                                                         
  The only difference between the two calls is a \n followed by body text. The subject line is identical.                
                                          
  ---                                                                                                                    
  Expected behaviour                                                                                                     
                                                                                                                         
  When conventional_commits = false, git-cliff should evaluate custom_major_increment_regex and                          
  custom_minor_increment_regex against the commit subject only (the first line), ignoring everything after the first
  newline. This is consistent with how conventional commit parsing works — the subject is always the first line.

  A commit starting with breaking should bump major.                                                                     
  A commit starting with feat should bump minor.
  This should hold whether the commit is single-line or multiline.                                                       
                                                                                                                         
  ---                                     
  Actual behaviour                                                                                                       
                                                                                                                         
  When a multiline string is passed, git-cliff appears to fail to isolate the subject and applies no bump rule — falling
  back to patch regardless of the prefix.                                                                                
                                          
  ---
  Workaround                                                                                                             
                                              
  Passing only the first line to --with-commit for bump calculation, then passing the full multiline string for changelog
   generation:                                
                                                                                                                         
  # bump call: subject only
  bump_commit_args = " ".join(                                                                                           
      f"--with-commit {shlex.quote(c.splitlines()[0])}" for c in commits
  )                                                                                                                      
                                          
  # changelog call: full multiline
  with_commit_args = " ".join(                                                                                           
      f"--with-commit {shlex.quote(c)}" for c in commits
  )                                                                                                                      
                                          
  This is a functional workaround but requires splitting one logical operation into two separate git-cliff invocations.
                                                                                                                         
  ---
  git-cliff version: 2.10.1                                                                                              
  OS: Linux                                   
                               

### Steps To Reproduce

1. Create a test git repo                                                                                              
                                                                                                                         
  mkdir /tmp/cliff-test && cd /tmp/cliff-test                                                                            
  git init                                    
  git commit --allow-empty -m "initial"                                                                                  
  git tag nati-1.0.0
                                                                                                                         
  2. Create cliff.toml                        
                                                                                                                         
  [git]           
  conventional_commits = false                                                                                           
  
  commit_parsers = [                                                                                                     
    { message = "^breaking", group = "🚀 Breaking Changes" },
    { message = "^feat",     group = "✨ Features" },
    { message = "^fix",      group = "🐛 Bug Fixes" },
  ]                                       

  [bump]                                                                                                                 
  custom_major_increment_regex = "^breaking"
  custom_minor_increment_regex = "^feat"                                                                                 
                                          
  3. Single-line commit — works correctly
                                                                                                                         
  git cliff --config cliff.toml \
    --tag-pattern '^nati-[0-9]+\.[0-9]+\.[0-9]+$' \                                                                      
    --bump --bumped-version \                 
    --with-commit 'breaking: remove v1 endpoints' \
    -- HEAD..HEAD
                                                                                                                         
  Expected: nati-2.0.0                        
  Actual: nati-2.0.0 ✓                                                                                                   
                  
  4. Same subject + body with no blank-line separator — breaks                                                           
                                          
  git cliff --config cliff.toml \                                                                                        
    --tag-pattern '^nati-[0-9]+\.[0-9]+\.[0-9]+$' \                                                                      
    --bump --bumped-version \                 
    --with-commit $'breaking: remove v1 endpoints\nThis removes the entire /v1 API surface.' \                           
    -- HEAD..HEAD 
                                                                                                                         
  Expected: nati-2.0.0 (subject still starts with breaking)
  Actual: nati-1.0.1 ✗ — patch bump, bump rules silently ignored                                                         
                                              
  5. Confirm the subject alone still works                                                                               
                                          
  git cliff --config cliff.toml \                                                                                        
    --tag-pattern '^nati-[0-9]+\.[0-9]+\.[0-9]+$' \                                                                      
    --bump --bumped-version \                                                                                            
    --with-commit 'breaking: remove v1 endpoints' \                                                                      
    -- HEAD..HEAD                         

  Expected: nati-2.0.0                                                                                                   
  Actual: nati-2.0.0 ✓ — stripping the body to subject-only restores correct behaviour
                                                                                                                         
  ---                                     
  The only variable between steps 3 and 4 is the \n followed by body text. The subject breaking: remove v1 endpoints is
  identical in both.                                                                                                     
                                          

### Expected behavior

  When conventional_commits = false, git-cliff should evaluate custom_major_increment_regex and                          
  custom_minor_increment_regex against the commit subject only (the first line), ignoring everything after the first
  newline. This is consistent with how conventional commit parsing works — the subject is always the first line.

  A commit starting with breaking should bump major.                                                                     
  A commit starting with feat should bump minor.
  This should hold whether the commit is single-line or multiline.   

### Screenshots / Logs

_No response_

### Software information

  git-cliff version: 2.12                                                                                     
  OS: Linux         

### Additional context

_No response_

YuriNachos:
## Description

When `conventional_commits` is disabled, `calculate_next_version_with_config` forwarded each commit's **full** raw message (subject + body) to `next_version::VersionUpdater::increment`. The `next_version` crate applies `custom_major_increment_regex`, `custom_minor_increment_regex` and `no_increment_regex` against the whole string for any commit that does not parse as conventional, so a commit whose **body** happens to contain the configured token drove a wrong version bump.

Concrete case from #1476: with `conventional_commits = false` and `custom_major_increment_regex = "breaking"`, a commit `fix: real bug\n\nnotes: breaking change discussed` bumps to a **major** release (the body word "breaking" matches) instead of a **patch** release.

This PR threads the `conventional_commits` flag into the bump calculation. When conventional commits are disabled, only the commit **subject** (first line) is handed to the version calculator, so body text can no longer trigger an increment. The conventional path is untouched — the full message is still passed when `conventional_commits = true`, so `BREAKING CHANGE:` footer detection and `!`-flag handling behave exactly as before.

The change reuses the existing `calculate_next_version_with_config` entry point (now a thin delegate that passes `conventional_commits = true`, preserving every existing caller/test) and moves the real work into `calculate_next_version_from_commits(config, conventional_commits)`, which `Changelog::bump_version` calls with `self.config.git.conventional_commits`.

## Motivation and Context

A body-driven major/minor bump is a silent, severe defect in a release tool: the resulting version is plausible enough to ship before anyone notices. This fixes the root cause described in #1476.

Note: the `^`-anchored reproducer from the original report no longer reproduces on `main` (it resolves through conventional-commit type matching), but the unanchored — and far more common — form of the bug remains, and this addresses the same root cause for both.

Closes #1476

## How Has This Been Tested?

- New unit test `release::test::custom_increment_regex_matches_subject_only_when_non_conventional`: a commit whose subject matches no regex but whose body contains both `breaking` and `feat`. With the fix it yields a **patch** bump (`1.0.1`); without the fix it yields a **major** bump (`2.0.0`). The test also asserts that a matching subject still drives major/minor bumps.
- Verified red → green: the new test fails on `main` behavior (`assert 1.0.1 == 2.0.0`) and passes on this branch.
- `cargo test -p git-cliff-core --lib` (all `release::` tests pass)
- Manual CLI reproduction with `--with-commit` of the exact multiline messages from the issue: body-only `breaking`/`feat` no longer bump the version; `BREAKING CHANGE:` footer detection in the default (conventional) path still produces a major bump.

## Types of Changes

### DIFF
--- a/git-cliff-core/src/changelog.rs
+++ b/git-cliff-core/src/changelog.rs
@@ -566,7 +566,10 @@ impl<'a> Changelog<'a> {
         crate::set_progress_message!("Bumping the version for unreleased changes");
         if let Some(ref mut last_release) = self.releases.iter_mut().next() {
             if last_release.version.is_none() {
-                let next = last_release.calculate_next_version_with_config(&self.config.bump)?;
+                let next = last_release.calculate_next_version_from_commits(
+                    &self.config.bump,
+                    self.config.git.conventional_commits,
+                )?;
                 tracing::debug!("Bumping the version to {}", next.version);
                 last_release.bump_type = next.bump_type;
                 last_release.version = Some(next.version.clone());
--- a/git-cliff-core/src/release.rs
+++ b/git-cliff-core/src/release.rs
@@ -118,6 +118,17 @@ impl Release<'_> {
     ///
     /// It uses the given bump version configuration to calculate the next
     /// version.
+    pub(super) fn calculate_next_version_with_config(&self, config: &Bump) -> Result<NextVersion> {
+        self.calculate_next_version_from_commits(config, true)
+    }
+
+    /// Calculates the next version based on the commits.
+    ///
+    /// When `conventional_commits` is disabled, bump evaluation uses only the
+    /// commit subject (the first line), so commit body text cannot trigger an
+    /// unintended version bump through the custom increment regular
+    /// expressions (`custom_major_increment_regex`,
+    /// `custom_minor_increment_regex` and `no_increment_regex`).
     #[cfg_attr(
         feature = "tracing",
         tracing::instrument(
@@ -128,7 +139,11 @@ impl Release<'_> {
             )
         )
     )]
-    pub(super) fn calculate_next_version_with_config(&self, config: &Bump) -> Result<NextVersion> {
+    pub(super) fn calculate_next_version_from_commits(
+        &self,
+        config: &Bump,
+        conventional_commits: bool,
+    ) -> Result<NextVersion> {
         crate::set_progress_message!(
             "Calculating the next version from commits with custom bump rules"
         );
@@ -190,7 +205,19 @@ impl Release<'_> {
                             &old_semver,
                             self.commits
                                 .iter()
-                                .map(|commit| commit.raw_message().trim_end().to_string())
+                                .map(|commit| {
+                                    let message = commit.raw_message().trim_end();
+                                    // When conventional commits are disabled,
+                                    // evaluate the bump regular expressions
+                                    // against the commit subject only, so body
+                                    // text cannot trigger an unintended version
+                                    // bump.
+                                    if conventional_commits {
+                                        message.to_string()
+                                    } else {
+                                        message.lines().next().unwrap_or_default().to_string()
+                                    }
+                                })
                                 .collect::<Vec<String>>(),
                         );
                         let bump_type = determine_bump_type(&old_semver, &new_semver);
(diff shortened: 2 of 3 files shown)


orhun (approved):
thanks

PR #1610, opened 2026-08-18
feat(changelog): add opt-in Markdown formatting for the output

### LINKED ISSUE #678: Format markdown

### Feature description

Writing a Tera template with perfect formatting is hard. Tweaking things with `{%-` instead of `{%` and `trim = true` is sometimes annoying.

### Desired solution

`git-cliff` should support outputting the formatted markdown file. e.g.

```toml
[changelog]
# formats the changelog (only supports markdown)
format = true
```

We can use a crate like [`prettify-markdown`](https://crates.io/crates/prettify-markdown) or something similar (and more up-to-date maybe).

We can detect markdown files by simply looking at the extension (`.md`).

### Alternatives considered

Users can always configure [`postprocessors`](https://git-cliff.org/docs/configuration/changelog#postprocessors) to do markdown formatting:

```toml
[changelog]
postprocessors = [
  { pattern = '.*', replace_command = "mdformat -" },
]
```

But I think it is better to support this feature out-of-the-box.

### Additional context

_No response_

ChrisJr404:
Closes #678. Adds an opt-in `[changelog] format = true` that runs the rendered changelog through a Markdown formatter, so you can get tidy output without fighting the template with `{%-` and `trim` everywhere.

It's off by default and the output is byte-for-byte what it is today unless you turn it on. When enabled it only kicks in for Markdown output (stdout or a `.md` file, detected by extension) — point `--output` at something else and it's skipped with a warning.

On the crate choice: the issue floated `prettify-markdown`, but as the last commenter noted that one's still alpha and doesn't cover much of Markdown. I went with a `pulldown-cmark` + `pulldown-cmark-to-cmark` round-trip instead — both are actively maintained and GFM-complete. The formatting is deliberately conservative: it keeps `-` as the bullet marker (matching the default templates, so existing changelogs don't churn `-` -> `*`), enables tables/strikethrough/task lists/footnotes so those survive, and doesn't reflow text or rewrite links.

One tradeoff worth flagging: the round-trip escapes bare bracket text, so a heading like `## [unreleased]` or a first release's `## [1.0.0]` with no compare URL comes out as `## \[unreleased\]`. That's standard CommonMark formatter behavior (prettier and mdformat do the same — bare `[text]` is an ambiguous undefined shortcut link) and headings that are real links like `## [1.0.0](…/compare/…)` are untouched. Didn't want to hack around it silently; happy to add an option to disable bracket escaping if you'd rather.

Kept it in `git-cliff-core` as a small `markdown` helper so it's unit-testable. Tests: the helper normalizes a messy-but-valid changelog and is idempotent, a clean one round-trips unchanged, and at the `generate()` level `format = true` produces the formatted output while leaving it unset is identical to before. `cargo test -p git-cliff-core` is green (75 passed), `cargo +nightly fmt --check` and `cargo clippy` are clean. Also smoke-tested the binary with `GIT_CLIFF__CHANGELOG__FORMAT=true` against this repo.


ChrisJr404:
Heads up on the two red checks, both look unrelated to this change. The Test suite job runs every test green and only fails on the codecov upload step (gpg: Can't check signature: No public key), which is the same failure that's currently red on main. The Links check is a lychee timeout on the contributor-covenant.org URL in CODE_OF_CONDUCT.md, a file this PR doesn't touch. A re-run should clear both whenever you get a chance.

website/docs/configuration/changelog.md
```diff
@@ -56,6 +56,19 @@ It is useful for adding indentation to the template for readability, as shown [i
...
+If set to `true`, the rendered changelog is passed through a Markdown formatter before it is written. This normalizes heading styles, list markers, and blank lines so you don't have to fight the template with `{%-` and `trim` to get tidy output.
+
+Formatting only runs when the output is Markdown, i.e. writing to stdout or to a file with a `.md` extension. It is off by default, and with it off the output is exactly what the templates render.
+
+```toml
+[changelog]
+format = true
+```
+
+This is an out-of-the-box alternative to configuring [`postprocessors`](#postprocessors) with an external tool like `mdformat`.
```
orhun: Can we wrap this in `:::note` and actually give the example to use `mdformat`? (c/p from the original issue :))

orhun:
This looks mostly good - can you add a [test fixture](https://github.com/orhun/git-cliff/tree/main/.github/fixtures) for it to validate its behavior?

Ping @alerque for additional thoughts on markdown formatting

ChrisJr404:
Added a fixture (`test-changelog-format-markdown`) that checks the Markdown formatting output. It renders a deliberately messy but valid changelog (`*` bullets and extra blank lines) with `format = true` and asserts the tidied result, so it fails if the formatting regresses. Wired it into the test-fixtures workflow matrix too. Thanks.

git-cliff-core/src/markdown.rs
```diff
@@ -0,0 +1,105 @@
...
+///
+/// The formatter options are intentionally conservative: it doesn't reflow
+/// text or rewrite links, so template output that is already valid Markdown
+/// keeps its structure.
+pub fn format_markdown(input: &str) -> Result<String> {
+    let mut options = Options::empty();
+    options.insert(Options::ENABLE_TABLES);
+    options.insert(Options::ENABLE_STRIKETHROUGH);
+    options.insert(Options::ENABLE_TASKLISTS);
+    options.insert(Options::ENABLE_FOOTNOTES);
```
orhun, line 26: Can we add [ENABLE_GFM](https://docs.rs/pulldown-cmark/latest/pulldown_cmark/struct.Options.html#associatedconstant.ENABLE_GFM) as well?

git-cliff/src/lib.rs
```diff
@@ -842,6 +842,21 @@ pub fn write_changelog<W: io::Write>(
...
         .clone()
         .or(changelog.config.changelog.output.clone());
+    // Markdown formatting only makes sense for Markdown output. Detect it from
+    // the file extension (stdout and extension-less paths are treated as
+    // Markdown, matching git-cliff's default output).
+    if changelog.config.changelog.format {
+        let is_markdown = output.as_ref().is_none_or(|path| {
+            path.extension()
+                .is_none_or(|ext| ext.eq_ignore_ascii_case("md"))
+        });
```
orhun: We should also check `args.prepend` for the file extension too.

e.g. `git cliff --prepend CHANGELOG.md`

orhun: Another issue: Using `--prepend` can duplicate the changelog header.

For example:

```toml
[changelog]
format = true
header = """
Changelog
=========
"""
```

In this case,  **git-cliff** tries to remove the old header before inserting the newly generated changelog. But it searches for the original configured text..

So **git-cliff** should remove the formatted version of the header when formatting is enabled.

alerque (approved):
I'm traveling and haven't compiled this to test, but this feature is something I'll likely use. Looking at the PR the test case at least makes sense, and `pulldown-cmark-to-cmark` is a great choice for fast and robust Markdown parsing and normalization. It might bother some people that do *not* want to use CommonMark compatible markup since it won't play nice with any random flavor you were expecting, but it's about the closest thing to a standard the Markdown world has to offer.

ChrisJr404:
Pushed the follow-ups:

- Added `ENABLE_GFM` to the formatter options.
- Fixed the header duplication with `--prepend`: when `format` is on, the existing file holds the formatted header, so the raw configured text no longer matches. It now strips the formatted header (falling back to the raw one when that is what is present) before prepending, so it is not duplicated. Covered by a new `changelog_prepend_strips_formatted_header` test.
- Markdown detection now also looks at the `--prepend` path extension, so `git cliff --prepend CHANGELOG.md` is treated as Markdown.
- Wrapped the `mdformat` mention in a `:::note` with a postprocessor example.

`cargo clippy` is clean and the core/lib tests pass locally.

orhun (approved):
amazing, thanks!

PR #1613, opened 2026-08-22
feat(remote)!: expose the pull request author as `commit.remote.pr_author`

### LINKED ISSUE #1608: Expose the pull request author in the remote context (`commit.remote.pr_author`)

### Feature description

`commit.remote.username` is the only account-level attribution a template has, and it is resolved from the **commit author's email**. When a forge writes an unexpected email into a squash commit, the changelog credits the wrong person, and there is currently no way to tell a template to prefer the pull request's author instead.

Concretely: GitHub's squash-merge can produce a commit whose author is the PR author's *name* but the merging user's *email* — `Contributor Name <maintainer@example.com>`. GitHub resolves that commit to the maintainer's account, so git-cliff renders `by @maintainer` for a PR that was opened, written, and signed off by the contributor.

In one repository this hit **10 of 31 squash-merged PRs (32%)** across four releases, quietly crediting the maintainer for a contributor's work. Every affected release note had to be corrected by hand after publication.

To be clear, this is not a bug in git-cliff: it faithfully renders what the GitHub API returns for that commit. The gap is that the better data — the pull request's author — is already fetched and then dropped.

### Desired solution

Expose the PR author as a new field, leaving `username` untouched so no existing configuration changes behavior:

```jinja
by @{{ commit.remote.pr_author | default(value=commit.remote.username) }}
```

The pull request is already matched to the commit; only its number, title, and labels are read from it (`git-cliff-core/src/remote/mod.rs`, in `update_release_metadata!`, as of 5963160):

```rust
let pull_request = pull_requests.iter().find(|pr| {
    pr.merge_commit() == Some(v.id().clone()) ||
        pr.merge_commit() == sha_short
});
commit.$remote.username  = v.username();               // <- always the commit author
commit.$remote.pr_number = pull_request.map(|v| v.number());
commit.$remote.pr_title  = pull_request.and_then(|v| v.title().clone());
commit.$remote.pr_labels = pull_request.map(|v| v.labels().clone()).unwrap_or_default();
```

`GitHubPullRequest` deserializes only `number`, `title`, `merge_commit_sha`, and `labels`, so the REST `/pulls` payload's `user.login` is discarded, and `RemotePullRequest` has no author method.

A sketch:

1. `RemotePullRequest` gains `fn author(&self) -> Option<String> { None }` — a default implementation, so no backend breaks.
2. `GitHubPullRequest` gains a `user` field and implements `author()` from `user.login`. The other forges expose an equivalent: GitLab `author.username`, Gitea `user.login`, Bitbucket `author.nickname`, Azure DevOps `createdBy`.
3. `commit.$remote.pr_author = pull_request.and_then(|v| v.author());` in the macro above.

Optionally `RemoteContributor` could carry the same value, so the contributor list agrees with the per-commit lines.

### Alternatives considered

- **`commit.author.name`** — correct in these cases, but it is a display string rather than an account, so `@`-mentions and profile links are lost.
- **`github.contributors[].username`** — same email-resolved value, same wrong account.
- **`postprocessors` mapping names to handles** — works, but it is a hand-maintained roster of contributors that silently goes stale.
- **Fixing it at merge time** — relies on whoever presses the merge button noticing the author GitHub proposes. At a 32% occurrence rate, that had already failed ten times.

### Additional context

Related, though neither overlaps this:

- **#949** (committer/author data in the remote context) asks for real names alongside usernames, for the contributor list. Worth noting that the template suggested there, `{{ commit.author.name }} (@{{ commit.remote.username }})`, renders `Contributor Name (@maintainer)` in this failure mode — the name of one person beside the handle of another, because the two halves resolve from different sources.
- **#1177** (PR linkage for rebased commits) concerns *which* commits receive PR metadata; this concerns *what* PR metadata is available once a commit has been matched. The two compose rather than collide.

Happy to open a PR if the approach looks right. The one design question I would want your call on is whether this should be a new `pr_author` field or an opt-in override of `username`.

jimisola:
## Description

Adds `pr_author` to the remote context, alongside the existing `username`:

```jinja2
by @{{ commit.remote.pr_author | default(value=commit.remote.username) }}
```

- `RemotePullRequest` gains `fn author(&self) -> Option<String>` with a default
  implementation returning `None`, so out-of-tree backends keep compiling.
- All five backends implement it: GitHub `user.login`, Gitea `user.login`,
  GitLab `author.username`, Bitbucket `author.nickname`, Azure DevOps
  `createdBy.displayName`.
- `update_release_metadata!` sets `commit.$remote.pr_author` from the pull
  request that was already matched to the commit.
- `username` is unchanged, so no existing configuration changes behavior.

Two notes for review:

- `pr_author` is set **per commit only**. `RemoteContributor` backs both
  `commit.remote` and the per-release `contributors` list, but contributors are
  deduplicated by `username` — so one entry can span pull requests opened by
  different people, and populating it there would credit the first PR's author
  for all of them. Contributor entries leave it empty.
- Bitbucket returns a raw `Name <email>` string for a commit author but an
  account object for a pull request author, so `BitbucketCommitAuthor` gained a
  `nickname` field rather than a new type, to avoid changing a public field's
  type. Note this also means Bitbucket's `username` is not a handle, so the
  docs there recommend `pr_author` on its own rather than a fallback chain.

Azure DevOps has no `@`-mentionable handle — `createdBy.displayName` is a
display name, and `uniqueName` is a UPN/email that would leak addresses into
changelogs — so that page documents `pr_author` without the `@` prefix.

`RemoteContributor` is not `#[non_exhaustive]`, so adding a field is a
source-breaking change for downstream crates constructing it by struct literal.

## Motivation and Context

Closes #1608.

`commit.remote.username` is resolved from the commit author's email. GitHub's
squash-merge can produce a commit whose author is the contributor's *name* with
the merging user's *email*, which GitHub resolves to the merging user's account
— so the changelog reads `by @maintainer` for a PR written by someone else. In
one repository this affected 10 of 31 squash-merged PRs (32%) across four
releases.

git-cliff renders what the API returns; the gap was that the better data, the
pull request's author, was already fetched and then dropped.

Per your comment on the issue, this adds `pr_author` and leaves `username`
alone.

## How Has This Been Tested?

- `cargo test --workspace --all-features` — all green.
- New `pull_request_author` and `pull_request_author_missing` unit tests per
  backend, deserializing the forge's actual JSON payload shape so the serde
  field names are covered, including the absent-author case (GitHub returns
  `"user": null` for ghost accounts).
- The existing `update_*_metadata` tests in `release.rs` now give each fixture
  pull request its own author, distinct from the commit author, so every
  assertion pins the PR-to-author mapping rather than a single shared value.
- `cargo +nightly fmt --all -- --check` clean.
- `cargo clippy --tests -- -D warnings` reports one pre-existing
  `iter_next_slice` error in `changelog.rs:567`, untouched by this change; no
  new lints from these files.

## Types of Changes

## Checklist:

  - [x] `cargo test`

### DIFF
--- a/git-cliff-core/src/commit.rs
+++ b/git-cliff-core/src/commit.rs
@@ -787,6 +787,7 @@ Refs: #123
         };
         commit.remote = Some(crate::contributor::RemoteContributor {
             username: None,
+            pr_author: None,
             pr_title: Some("feat: do something".to_string()),
             pr_number: None,
             pr_numbers: vec![],
@@ -849,6 +850,7 @@ Refs: #123
         };
         commit.remote = Some(crate::contributor::RemoteContributor {
             username: None,
+            pr_author: None,
             pr_title: Some("feat: do something".to_string()),
             pr_number: None,
             pr_numbers: vec![],
@@ -1070,6 +1072,7 @@ Refs: #123
         };
         commit.remote = Some(crate::contributor::RemoteContributor {
             username: None,
+            pr_author: None,
             pr_title: Some("feat: do something".to_string()),
             pr_number: None,
             pr_numbers: vec![],
--- a/git-cliff-core/src/contributor.rs
+++ b/git-cliff-core/src/contributor.rs
@@ -7,6 +7,12 @@ use serde::{Deserialize, Serialize};
 pub struct RemoteContributor {
     /// Username.
     pub username: Option<String>,
+    /// Account that opened the pull request.
+    ///
+    /// Unlike `username`, this is not resolved from the commit author. Only set
+    /// per commit; contributor entries leave it empty because they are
+    /// deduplicated by `username`.
+    pub pr_author: Option<String>,
     /// Title of the pull request.
     pub pr_title: Option<String>,
     /// The pull request that the user created.
--- a/git-cliff-core/src/release.rs
+++ b/git-cliff-core/src/release.rs
@@ -873,6 +873,9 @@ mod test {
             .collect(),
             vec![
                 GitHubPullRequest {
+                    user: Some(GitHubCommitAuthor {
+                        login: Some(String::from("author-1")),
+                    }),
                     title: Some(String::from("1")),
                     number: 42,
                     merge_commit_sha: Some(String::from(
@@ -883,6 +886,9 @@ mod test {
                     }],
                 },
                 GitHubPullRequest {
+                    user: Some(GitHubCommitAuthor {
+                        login: Some(String::from("author-2")),
+                    }),
                     title: Some(String::from("2")),
                     number: 66,
                     merge_commit_sha: Some(String::from(
@@ -893,6 +899,9 @@ mod test {
                     }],
                 },
                 GitHubPullRequest {
+                    user: Some(GitHubCommitAuthor {
+                        login: Some(String::from("author-3")),
+                    }),
                     title: Some(String::from("3")),
                     number: 53,
                     merge_commit_sha: Some(String::from(
@@ -903,6 +912,9 @@ mod test {
                     }],
                 },
                 GitHubPullRequest {
+                    user: Some(GitHubCommitAuthor {
+                        login: Some(String::from("author-4")),
+                    }),
                     title: Some(String::from("4")),
                     number: 1_000,
                     merge_commit_sha: Some(String::from(
@@ -913,6 +925,9 @@ mod test {
                     }],
                 },
                 GitHubPullRequest {
+                    user: Some(GitHubCommitAuthor {
+                        login: Some(String::from("author-5")),
+                    }),
                     title: Some(String::from("5")),
                     number: 999_999,
                     merge_commit_sha: Some(String::from(
@@ -934,6 +949,7 @@ mod test {
                 message: String::from("add github integration"),
                 github: RemoteContributor {
                     username: Some(String::from("orhun")),
+                    pr_author: Some(String::from("author-1")),
                     pr_title: Some(String::from("1")),
                     pr_number: Some(42),
                     pr_numbers: vec![],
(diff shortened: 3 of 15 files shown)


jimisola:
All three red checks look unrelated to this PR — details in case they're useful:

**`Test suite`** — the tests themselves passed; the job failed at *Upload reports to codecov* with `gpg: Can't check signature: No public key` → `Could not verify signature`. The same step failed on `main` in run `32535820181`, before this PR existed.

**`test-release-statistics`** — the only diff is `1599` vs `1598 day(s) passed between releases`. `run-fixtures-test/action.yml` computes the expected value as `(now - previous_release_timestamp) / 86400` (floor of elapsed seconds), while git-cliff reports the calendar-day difference. Measured from `2022-04-06 01:25:12`, the two agree all day except between 00:00 and 01:25:12 UTC:

| time (UTC) | action (floor elapsed) | git-cliff (calendar days) |
| --- | --- | --- |
| Aug 21 23:09 | 1598 | 1598 |
| Aug 22 00:16 | 1598 | 1599 |

This job ran at 00:16 UTC. Happy to open a separate issue for it if you'd like.

**`test-gitlab-integration-custom-range`** — git-cliff panicked on a network timeout, so `output.md` was empty:

```
Could not get gitlab metadata: ... error sending request for url
(https://gitlab.com/api/v4/projects/7283906/repository/commits?per_page=100&page=2&ref_name=89de5e8e50)
Caused by: operation timed out
```

That endpoint responds normally off the runner. This PR adds no fields to any GitLab struct, so deserialization there is unchanged.

I can't re-run the jobs from a fork, but I'll push a rebase once the statistics fixture's window has passed.

orhun (approved):
thanks!

PR #1617, opened 2026-08-24
test(args): cover configuration file resolution

lazizbekravshanov:
Follow-up to #1584.

`resolve_config_path` went in without direct tests, so the `--workdir` discovery behaviour you spotted there is currently not guarded by anything. This adds six tests for it and changes no production code.

## What is covered

- an explicit `--config` path that exists is used as given
- an explicit path that does not exist skips discovery and uses the user configuration
- discovery follows `--workdir` rather than the directory git-cliff was invoked from, which is the regression from #1584
- discovery uses the current directory when `--workdir` is absent
- discovery walks up to a parent directory
- the user configuration is used when no project configuration is found

## Verification

Each test was checked to fail when the behaviour it names is broken, rather than only to pass as written. Removing the explicit-config arm, making a missing explicit path fall through to discovery, ignoring `--workdir`, and limiting the ancestor walk to a single directory each fail exactly one test, and it is the test named for that behaviour.

Each test also isolates its own branch: the explicit-path test keeps its configuration outside the directory discovery would search, the missing-path test leaves a discoverable configuration in place to prove discovery is skipped, and the `--workdir` test puts a competing configuration in the current directory so it asserts precedence rather than merely finding something.

`cargo test --workspace` passes (22 in git-cliff, 113 in git-cliff-core), `cargo clippy --tests -- -D warnings` is clean, and `cargo +nightly fmt --all --check` is clean.

## Scope

These are unit tests of the resolver itself. The call site in `run_with_changelog_modifier` that passes `args.workdir` into it is not covered here, so reverting that single argument would reproduce the original bug with these tests still green. Covering that would need a CLI fixture rather than a unit test, which felt like a separate change; happy to add one if you would like the plumbing guarded too.

`temp-dir` is added as a dev-dependency of the `git-cliff` crate, matching the version and usage already present in `git-cliff-core`.

### DIFF
--- a/git-cliff/src/lib.rs
+++ b/git-cliff/src/lib.rs
@@ -1011,9 +1011,15 @@ pub fn write_changelog<W: io::Write>(
 mod tests {
     use std::fs;
 
+    use pretty_assertions::assert_eq;
     use temp_dir::TempDir;
 
     use super::*;
+
+    /// Creates a temporary directory.
+    fn temp_dir() -> Result<TempDir> {
+        Ok(TempDir::with_prefix("git-cliff-")?)
+    }
 
     /// Creates a project configuration file in the given directory.
     fn write_config(dir: &Path) -> Result<PathBuf> {
@@ -1024,18 +1030,30 @@ mod tests {
 
     #[test]
     fn explicit_config_is_used_when_it_exists() -> Result<()> {
-        let dir = TempDir::new()?;
-        let config = write_config(dir.path())?;
+        let config_dir = temp_dir()?;
+        let current_dir = temp_dir()?;
+        // The configuration lives outside the directory discovery would
+        // search, so only the explicit branch can return this path.
+        let config = write_config(config_dir.path())?;
         assert_eq!(
-            Some(config.clone()),
-            resolve_config_path(Some(&config), None, dir.path(), || None)
+            Some(config),
+            resolve_config_path(
+                Some(&config_dir.path().join(DEFAULT_CONFIG)),
+                None,
+                current_dir.path(),
+                || None
+            )
         );
         Ok(())
     }
 
     #[test]
-    fn missing_explicit_config_falls_back_to_the_user_config() -> Result<()> {
-        let dir = TempDir::new()?;
+    fn missing_explicit_config_skips_discovery_and_uses_the_user_config() -> Result<()> {
+        let dir = temp_dir()?;
+        // A discoverable project configuration is present, so this also
+        // proves that giving `--config` skips discovery entirely rather
+        // than falling through to it.
+        write_config(dir.path())?;
         let user_config = dir.path().join("user-cliff.toml");
         assert_eq!(
             Some(user_config.clone()),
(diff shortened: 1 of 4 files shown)


lazizbekravshanov:
I found a gap in this PR while re-reading it against the behaviour it is supposed to guard, and it is my own miss.

All six tests here pass an absolute path for `--workdir`. None of them passes a relative one, and the relative case is the one that is actually broken on `main` right now:

```sh
base=$(mktemp -d)
mkdir -p "$base/deep/a/b/project"

cat > "$base/deep/cliff.toml" <<'TOML'
[changelog]
header = "FROM_ANCESTOR_CONFIG\n"
body = "{% for group, commits in commits | group_by(attribute=\"group\") %}{% for c in commits %}- {{ c.message }}\n{% endfor %}{% endfor %}"
trim = true
TOML

cd "$base/deep/a/b/project"
git init -q -b main . && git config user.email t@t.t && git config user.name t
echo x > f.txt && git add f.txt && git commit -q -m "feat: add the thing"

cd "$base/deep/a"
git-cliff --workdir "$base/deep/a/b/project" --unreleased   # uses $base/deep/cliff.toml
git-cliff --workdir "b/project" --unreleased                # "cliff.toml" is not found
```

Same directory, two spellings, two results. `resolve_config_path` uses the `--workdir` value as given, and `Path::ancestors` on a relative value yields only that value's own components and then `""`, so the walk stops at the current directory instead of continuing to the root. Relative is the common spelling, and this repository's own fixture uses `--workdir .github/fixtures/`.

Resolving the search root against the current directory first covers both, since `Path::join` replaces rather than appends when the joined path is absolute:

```rust
let search_root = workdir.map_or_else(
    || current_dir.to_path_buf(),
    |workdir| current_dir.join(workdir),
);
```

I have this done on a branch, with a test for the relative case that fails before the change and passes after, and the absolute case kept alongside it so the equivalence is pinned. Workspace tests, clippy with `-D warnings` and nightly fmt are clean.

How would you like it? I can push the failing test plus the one line fix into this PR, which would mean it stops being test only, or I can leave #1617 as pure coverage and send the fix as its own PR on top. Happy either way, it is your call on how you would rather review it.

Sorry for the back and forth on this one. The original `--workdir` regression was mine, and so was writing the follow-up tests without pointing them at the case that broke.

git-cliff/src/lib.rs
orhun, line 1: Can we move these to another module? Having tests in `lib.rs` seems a bit off to me :)

orhun (commented):
Thanks for the PR!

lazizbekravshanov:
Moved them in 7fbc87f. `resolve_config_path` and its tests now live in `git-cliff/src/config_path.rs`, which also matches how `args.rs` keeps its tests next to the code they cover.

I also rebased onto main to clear the `Cargo.toml` conflict from the v2.14.1 bump, keeping both `serial_test` 3.5.0 and the `temp-dir` dev-dependency. Since that was a force push after your review, `git range-diff` confirms the two test commits replayed unchanged apart from the version bump showing up as context.

Still open whenever you have a view: the relative `--workdir` bug from my earlier comment. The fix and a test for it are ready on a branch, and I can either push them here or send them as a separate PR on top, whichever you prefer to review.

orhun (approved):
thanks!

PR #1627, opened 2026-09-01
fix(args): resolve --workdir to a repo-relative include path

### LINKED ISSUE #1369: Setting `--workdir` results in an empty output

Jorge-Polanco-Roque:
Fixes #1369.

## Problem
Pointing `--workdir` at the repository root produces an empty changelog.

## Root cause
When `--workdir` is set, the changelog is scoped by turning the workdir path into an include pattern. That pattern was absolute (built from `workdir.join("")`), but diff paths are relative to the repo root — so an absolute pattern matches nothing and the changelog comes out empty.

## Approach
Resolve `--workdir` to a **repo-relative** include pattern inside `process_repository`. If `workdir` resolves to the repo root itself, no filter is added so everything is kept. Removed the old absolute-pattern assignment in `run_with_changelog_modifier`.

## Testing
- Added the `test-workdir-repo-root` fixture (`commit.sh` + `expected.md`) and wired it into `test-fixtures.yml` with `--workdir "$PWD"`, which exercises the previously-broken absolute-path case. It produces the full changelog with the fix and empty output without it.

Note on formatting: my added lines are within your `max_width`/`comment_width` and add no imports, so they match your nightly rustfmt config; I don't have nightly locally, so a quick `cargo +nightly fmt` on your side is worth a glance.

## A note on direction
I noticed you're planning a more holistic rework of the workdir/include-path handling and that @o1x3 offered to help. This is a minimal, test-backed fix for the empty-output regression specifically — happy for it to serve as a stopgap or a reference for the larger refactor, and equally happy to close it if you'd rather fold it into that work. Per CONTRIBUTING I'm flagging it here rather than assuming a direction; let me know how you'd like to proceed.

### DIFF
--- a/.github/fixtures/test-workdir-empty-commit/cliff.toml
+++ b/.github/fixtures/test-workdir-empty-commit/cliff.toml
@@ -1,0 +1,20 @@
+[changelog]
+body = """
+{% if version %}\
+    ## [{{ version | trim_start_matches(pat="v") }}] - {{ timestamp | date(format="%Y-%m-%d") }}
+{% else %}\
+    ## [unreleased]
+{% endif %}\
+{% for group, commits in commits | group_by(attribute="group") %}
+    ### {{ group | upper_first }}
+    {% for commit in commits %}
+        - {{ commit.message | upper_first }}\
+    {% endfor %}
+{% endfor %}\n
+"""
+
+[git]
+commit_parsers = [
+    { message = "^feat", group = "Features" },
+    { message = "^fix", group = "Bug Fixes" },
+]
--- a/.github/fixtures/test-workdir-empty-commit/commit.sh
+++ b/.github/fixtures/test-workdir-empty-commit/commit.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+set -e
+
+# Regression test for https://github.com/orhun/git-cliff/issues/1369
+# A relative `--workdir .` on a repo whose commits have no file changes must
+# still list them. Setting --workdir used to add an include-path, which
+# filtered out file-less commits and produced an empty changelog.
+
+GIT_COMMITTER_DATE="2022-04-06 01:25:08" git commit --allow-empty -m "feat: add readme"
+GIT_COMMITTER_DATE="2022-04-06 01:25:09" git commit --allow-empty -m "fix: add main module"
--- a/.github/fixtures/test-workdir-empty-commit/expected.md
+++ b/.github/fixtures/test-workdir-empty-commit/expected.md
@@ -1,0 +1,10 @@
+## [unreleased]
+
+### Bug Fixes
+
+- Add main module
+
+### Features
+
+- Add readme
+
--- a/.github/fixtures/test-workdir-repo-root/cliff.toml
+++ b/.github/fixtures/test-workdir-repo-root/cliff.toml
@@ -1,0 +1,20 @@
+[changelog]
+body = """
+{% if version %}\
+    ## [{{ version | trim_start_matches(pat="v") }}] - {{ timestamp | date(format="%Y-%m-%d") }}
+{% else %}\
+    ## [unreleased]
+{% endif %}\
+{% for group, commits in commits | group_by(attribute="group") %}
+    ### {{ group | upper_first }}
+    {% for commit in commits %}
+        - {{ commit.message | upper_first }}\
+    {% endfor %}
+{% endfor %}\n
+"""
+
+[git]
+commit_parsers = [
+    { message = "^feat", group = "Features" },
+    { message = "^fix", group = "Bug Fixes" },
+]
--- a/.github/fixtures/test-workdir-repo-root/commit.sh
+++ b/.github/fixtures/test-workdir-repo-root/commit.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+set -e
+
+# Regression test for https://github.com/orhun/git-cliff/issues/1369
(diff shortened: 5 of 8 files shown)


Jorge-Polanco-Roque:
Thanks for the CI run — I pushed a fix for the failing `test-workdir-repo-root` fixture.

**Root cause:** the `commit.sh` I added was committed without its executable bit (mode `100644` instead of `100755`), so the `./commit.sh` step in the fixtures action hit a permission error, created **no commits**, and git-cliff then errored on an empty repo. Fixed by adding the exec bit — the fixture now passes.

I reproduced the exact CI invocation locally (`git init && ./commit.sh`, then `git cliff --config cliff.toml --workdir "$PWD"`) and the output matches `expected.md`.

The remaining **Lints** and **Build Nix flake** failures look unrelated to this change: the clippy errors (`collapsible_if`, `missing_docs`) are all in `git-cliff-core` files this PR doesn't touch and appear to come from a newer nightly toolchain flagging pre-existing code — my `lib.rs` change isn't among them. Happy to open a separate PR for those if it'd help.

And as noted above: I'm glad to have this closed or folded into the larger workdir/include-path rework you mentioned — it's meant only as a minimal stopgap for the empty-output regression.

Jorge-Polanco-Roque:
Pushed a fix for the failing **Lints** check — clippy is now green (`cargo clippy --tests -- -D warnings` passes). The warnings were `collapsible_if` (collapsed the nested `if let`/`if` into let-chains) and `missing_docs` on the `RustEmbed`-generated functions. **Formatting** is green too.

The remaining red checks look unrelated to this change:
- **Build Nix flake**: a transient `curl 403` while fetching a crate tarball (no `Cargo.lock` change here).
- **Check NodeJS tarball**: looks for an unpublished `git-cliff-*@2.14.1` npm package (release-time check).
- **Test suite / test-release-statistics**: `repo::test::get_latest_tag` and a date-based fixture — history/date-dependent tests that don't pass on a fork checkout, independent of this diff.

Happy to rebase or adjust if you'd like.

orhun:
Looks good overall, but please fix merge conflicts, drop the unrelated lint changes and add a fixture test to fully cover #1369

Jorge-Polanco-Roque:
Thanks for the review! Addressed all three:

- Rebased onto main and resolved the conflicts (kept the let-chain `cwd` logic from main and wrote the new `--workdir` block the same way).
- Dropped the unrelated lint commits, so the PR is now a single commit with just the `--workdir` fix.
- Added a second fixture to fully cover #1369: alongside `test-workdir-repo-root` (absolute path), `test-workdir-empty-commit` reproduces the issue's other case, a relative `--workdir .` on file-less commits, which used to get filtered out to an empty changelog. Both fixtures pass locally.

orhun (approved):
LGTM, thanks!

PR #1629, opened 2026-09-06
docs: add existing changelog migration guide

### LINKED ISSUE #1607: Document how git-cliff interacts with a existing Changelog
Currently, i dont see much documentation on how git-cliff interacts with a existing (pre git-cliff) changelog, or a guide / tutorial on how to migrate it without issues.

I had just made a migration from pre-git-cliff to git-cliff (https://github.com/tramhao/termusic/pull/754).
The first try would always completely overwrite the changelog (without any extra options), then i found `--prepend`, which was also not working too great until i found #712 suggesting that the header needs to be the exact same as defined in the config (i guess to be expected), after that it would correctly add the new release between the old and the header (unless it is the same release, where it would just add a duplicate).

To reduce those issues, i would recommend that a guide is included that:
- includes what prerequisites are necessary for the old changelog to be compatible with git-cliff
  - for example what headers may need to be modified (when they uses a different style than git-cliff)
  - and migratory markers to be placed (ex. `<!-- git-cliff dont modify -->` or `<!-- git-cliff old changelog below -->`)
- what cli & config options to use
  - like `--prepend`
- what limitations there might be
  - like the headers needing to be the same as with the config to work properly

It would be great if a similar guide could be done for release-plz (though i might open a issue there instead).

shutaruku121-bit:
## Summary

- add a migration guide for projects that already have a `CHANGELOG.md`
- document how static headers and `header_marker` interact with `--prepend`
- call out duplicate-release and output/prepend limitations
- link the guide from the existing `--prepend` usage example

Closes #1607.

## Verification

- `cd website && yarn build`
- `git diff --check`

### DIFF
--- a/website/docs/usage/examples.md
+++ b/website/docs/usage/examples.md
@@ -81,6 +81,9 @@ git cliff -o
 
 Prepend new changes to an existing changelog file:
 
+See [Migrating an existing changelog](/docs/usage/migrating-existing-changelog)
+if the file was created before you started using **git-cliff**.
+
 ```bash
 # 1- changelog header is removed from CHANGELOG.md
 # 2- new entries are prepended to CHANGELOG.md without footer part
--- a/website/docs/usage/migrating-existing-changelog.md
+++ b/website/docs/usage/migrating-existing-changelog.md
@@ -1,0 +1,119 @@
+---
+sidebar_position: 14
+---
+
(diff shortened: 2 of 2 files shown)


shutaruku121-bit:
Closing for now so I can review and validate the contribution more carefully before resubmitting.

PR #1637, opened 2026-09-10
fix(git): keep commits without file changes when only `exclude_paths` is set

kristof-mattei:
## Description

When using `exclude_paths`, still include empty commits.

## Motivation and Context

I created an empty marker commit to evoke a change in version number, however, since I had `exclude_paths` set, this PR did not contribute to the new version number. 

I marked this as a bug fix because this change still satisfies the existing documentation:

https://git-cliff.org/docs/configuration/git/#:~:text=exclude%5Fpaths%20is%20an%20optional%20array%20of%20glob%20patterns%2E%20Commits%20that%20only%20modify%20files%20matching%20these%20patterns%20will%20be%20excluded%20from%20the%20changelog

> `exclude_paths` is an _optional_ array of glob patterns. Commits that **only** modify files matching these patterns will be excluded from the changelog.

Since an empty commit does NOT modify ANY files, it does not modify ANY files in the `exclude_paths`, and thus should be included. 

## How Has This Been Tested?

* Added tests
* Tested build locally to verify

N/A

## Types of Changes

- [ ] Other 

## Checklist:

  - [x] `cargo test`

### DIFF
--- a/git-cliff-core/src/repo.rs
+++ b/git-cliff-core/src/repo.rs
@@ -380,6 +380,10 @@ impl Repository {
                 })
             }
             (None, Some(exclude_pattern)) => {
+                // don't exclude empty commits
+                if changed_files.is_empty() {
+                    return true;
+                }
                 // check if the commit has at least one changed file that does not
                 // match all exclude patterns.
                 changed_files.iter().any(|path| {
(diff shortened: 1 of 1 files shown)


orhun (approved):
thanks!

